PR #26, opened 2021-06-16
Added support for downloaded chapter titles.

Mbucari:
Added support for custom chapters to AaxToM4bConverter.

Chapters is now a base class, and the old Chapters functionality is in AAXChapters which is derived from Chapters. Refactored Chapters so it contains a Chapter list. Chapter has chapter title, start time, and end time.

Added new Chapters derived class, DownloadedChapters, to be used with MetadataDtoV10.ContentMetadata.ChapterInfo (added in my earlier AudibleApi pull request).

Added call to Api in DecryptBook to download chapter metadata and pass DownloadedChapters to AaxToM4bConverter. If call fails, null is passed and AaxToM4bConverter will retrieve chapters from aax file (existing functionality).

Refactored Cue and added support for chapter titles.

### DIFF
--- a/AaxDecrypter/UNTESTED/AAXChapters.cs
+++ b/AaxDecrypter/UNTESTED/AAXChapters.cs
@@ -1,0 +1,40 @@
+﻿using System.Diagnostics;
+using System.Globalization;
+using System.Linq;
+using Dinah.Core.Diagnostics;
+
+namespace AaxDecrypter
+{
+    public class AAXChapters : Chapters
+    {
+        public AAXChapters(string file)
+        {
+            var info = new ProcessStartInfo
+            {
+                FileName = DecryptSupportLibraries.ffprobePath,
+                Arguments = "-loglevel panic -show_chapters -print_format xml \"" + file + "\""
+            };
+            var xml = info.RunHidden().Output;
+
+            var xmlDocument = new System.Xml.XmlDocument();
+            xmlDocument.LoadXml(xml);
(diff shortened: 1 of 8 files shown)


rmcrackan:
Holy crap, this is really good work. (Your having mirrored my weird unofficial conventions is praise-worthy on its own.) You're obviously a skilled C# developer and also have a good grasp on the audible api. I'm surprised I haven't heard of you before and I'll bet this isn't the last time. Thanks for chipping in!

Mbucari:
Thanks for the kind words! A couple years ago I made an audible library backup for myself. It didn't interact with the API, it just used the Windows app's cache files. I really wanted to add downloaded chapters title support, so I started googling and found your repo. I think I'm going to abandon my project and work on yours to make it more to my liking, if that's all right with you!

rmcrackan:
Sure, I'm happy to look at any PRs. There's a lot in here that I'm proud of, a lot which needs improvement, and a fair amount of tech debt. You're welcome to do as much or little as you want. I'm emphatic about keeping this non-obligation. I work on it when I feel like, I don't when I don't, and I welcome anyone else to do the same. If you want ideas, I've got those too.

Other values: other than what's technically mandatory to use the api, I've tried hard to not send or store any identifiable info -- including in log files so that people can send me those for bugs. I'm obviously anti DRM -- I'm for people having control over their own files and data. I'm anti piracy but what people do with their liberated stuff is their responsibility. I want to keep this free and open source but the moment it even looks like I could be legally culpable for anything I'll drop this like it's made of lava. I'm a programmer trying to give back to the audiobook community, not a political activist looking for trouble.

PR #27, opened 2021-06-19
Add support for AXXC

Mbucari:
Added AudibleKey  and AudibleIV to Book. and created DbContext migration.

Modified AaxToM4bConverter.CreateAsync to accept AudibleKey and AudibleIV instead of DecryptKey.

Modified AaxToM4bConverter to decrypt AAXC. Upgraded FFMPEG to 4.4-19, which supports AAXC scheme. It works, but ffmpeg command could use some tweaking for speed.


rmcrackan:
This looks really cool. With your last PRs it was a small change set so I just did a lot of copy/pasting to view the changes locally. I know just enough about git to not be entirely useless. Do you know if it's possible to bring your changes onto my laptop such that Visual Studio will show me which files are changed from my main?

Mbucari:
I'm brand new to git. TBH, I'm just a hobbyist and this is my first collaboration. But if I were you trying to test my code prior to pulling in changes, I'd create a new local repository on your pc with a new solution and pull in my AudibleApi and Libation repos.

AaxDecrypter/DecryptSupportLibraries.cs
```diff
@@ -16,6 +16,5 @@ public static class DecryptSupportLibraries
         public static string ffmpegPath { get; } = Path.Combine(decryptLib_, "ffmpeg.exe");
         public static string ffprobePath { get; } = Path.Combine(decryptLib_, "ffprobe.exe");
         public static string atomicParsleyPath { get; } = Path.Combine(decryptLib_, "AtomicParsley.exe");
-        public static string mp4trackdumpPath { get; } = Path.Combine(decryptLib_, "mp4trackdump.exe");
```
rmcrackan, line 19: Should also delete the actual assembly and its entry in the csproj file

FileLiberator/DownloadBook.cs
```diff
@@ -40,7 +41,53 @@ private static string getDownloadPath(LibraryBook libraryBook)
...
+				libraryBook.Book.Locale,
+				Account = libraryBook.Account?.ToMask() ?? "[empty]",
+				tempAaxFilename,
+				actualFilePath,
+				length,
+				contents
+			});
+			throw ex;
+		}
 		private async Task<string> downloadBookAsync(LibraryBook libraryBook, string tempAaxFilename)
```
rmcrackan, line 91: If unused, this can be deleted

FileLiberator/DownloadBook.cs
```diff
@@ -40,7 +41,53 @@ private static string getDownloadPath(LibraryBook libraryBook)
 				libraryBook.Book.Title,
 				"aax",
 				libraryBook.Book.AudibleProductId);
+		private async Task<string> downloadAacxBookAsync(LibraryBook libraryBook, string tempAaxFilename)
```
rmcrackan, line 44: 1) typo: aacx should be aaxc -- the last 2 letters are flipped

2) These new AudibleKey and AudibleIV are new to me. Are all audible books now encoded per book rather than per account activation key? Is this true for both aax and aaxc? Also, if true for both types, we can just reuse the downloadBookAsync name so it doesn't appear that it works only for non-aax

mkb79:
The AAXC  decryption key/IV pair is unique for each file. AAX encrypted files can be decrypted with the activation bytes, which are unique per account.

rmcrackan:
> The AAXC decryption key/IV pair is unique for each file. AAX encrypted files can be decrypted with the activation bytes, which are unique per account. 

Oh, so both methods are needed. Thanks

mkb79:
> > The AAXC decryption key/IV pair is unique for each file. AAX encrypted files can be decrypted with the activation bytes, which are unique per account.
> 
> Oh, so both methods are needed. Thanks

If you want free audiobooks or Audible Plus books, then you need AAXC. Otherwise AAX will do the job too.

FileLiberator/DownloadBook.cs
rmcrackan: Follow up to previous comments -- it looks like this should still be used for aax. The poorly named property "IsNonLibraryAudiblePlus" can help here

Mbucari:
I think adding Audible Plus support to Libation is important because the desktop app doesn't support Audible Plus content, so the only way to listen is through a mobile app.

Mbucari:
> 
> 
> > The AAXC decryption key/IV pair is unique for each file. AAX encrypted files can be decrypted with the activation bytes, which are unique per account.
> 
> Oh, so both methods are needed. Thanks

I don't think both are needed. The /1.0/content/asin/licenserequest seems to always return the AAXC format. If you get rid of the workaround and only use that API, you should only ever be served aaxc files.

rmcrackan:
> I think adding Audible Plus support to Libation is important because the desktop app doesn't support Audible Plus content, so the only way to listen is through a mobile app.

Agreed. I'm excited to get this working.

I'm running into a lot of errors during my testing. It might be quicker to communicate in some way other than github. My email is in my profile if you'd like to exchange a more direct way to chat

rmcrackan:
> I'm running into a lot of errors during my testing

I've narrowed many errors down to Api.Download decryptLicenseResponse(). this is what my "plainText" looks like, so JObject.Parse(plainText) throws:

d??z1?w?di	F????]EO?^???7?Q
?5/??tx/???>?????(iX?E?n?????/???Z)?:R? L?????????74q??D^?#U\?Nt??????7e?0???4^Rn[??&<X#n??|w?&?'g?
????\??.??e?I?????vE?s?\O??%????u?P?E$9G???*i??O$?5??????
?????.?mw?e???;????_;?u/?[
???{?????o~??4?Im"?????????!?PUC6X+/?a?|G??VMz)?0nb??Z*M????Z?X?j??s???p??BRy?0A,?&?????]q?O????I??Gu??B??N???hF?8??F?p

Mbucari:
> 
> 
> > I'm running into a lot of errors during my testing
> 
> I've narrowed many errors down to Api.Download decryptLicenseResponse(). this is what my "plainText" looks like, so JObject.Parse(plainText) throws:
> 
> d??z��1?�w?��di F????]EO?^?�??7?�Q�
> ?5/?�?tx/??�?>???�??(iX?E?n?????��/???Z)?:R? L??�?�????�??74q??D^�?#�U?�Nt??????7e?0???4^Rn[??&<X#n�??���|w?&?'g?
> ??????.??e?I??�???vE?s?\O??%????u?�P?E$9G???�_i??O�$?�5????�?? ?????.�?mw?e???;????_�;?u/?[ ???�{?????o~??4?Im"????�?????!�?PUC6X+/?a?|G??V�Mz)?�0nb??Z�_M�????Z?X?j??s???p??B�Ry?0A,?&?????]�q?O????I??G�u??��B??N?�??hF?8??F?p

To decrypt that, you need the DeviceType, DeviceSerialNumber and AmazonAccountId in Identity (added in latest AudibleApi commits). The Api only retrieves those when libation logs in and registers the device, then saves them as part of your identity. If you're testing on your old AccountSettings.json, those are probably empty strings. Probably need to add code to re-authorize Libation after upgrading.

rmcrackan:
I'll mess with this manually for testing. For release though, how much extra work would it be to include a migrate step into LibationLauncher.Program? Or alternately, getLicenseResponseDecryptionKey could throw an instructive error if those are null.

Mbucari:
> 
> 
> I'll mess with this manually for testing. For release though, how much extra work would it be to include a migrate step into LibationLauncher.Program? Or alternately, getLicenseResponseDecryptionKey could throw an instructive error if those are null.

I'll look into it.

rmcrackan:
This is AWESOME. For the first time -- aaxc decryption. I successfully tested on an audible plus and non-audible plus title

See all prev notes about minutia and the one important one about those nulls during getLicenseResponseDecryptionKey . After these are in and the merge is complete, I've got some ideas based on your changes before we release it into the wild.

rmcrackan:
> If you want free audiobooks or Audible Plus books, then you need AAXC. Otherwise AAX will do the job too.

@mkb79  : First of all -- good to see you again. I think the guy who wrote the api in crystal has stopped, so it's now just the people on this thread.

Second: Thanks for the clarification. I tested Mbucari's changes and he's also right: you can use the aaxc method on all books. Although it adds complexity for that path, it means we can just use one path for both rather than separate methods for aax vs aaxc

mkb79:
@rmcrackan : Yeah, the method used to download AAXC files are able to download every sort of Audible books. It's the same method which the iOS and Android Audible App uses. The drawback is that this method needs a unique key/IV pair for every file to decrypt them. You have to find a good method to store them for later use. If you loose them, you have to download the file again (if your account is valid or paid for Audible Plus titles)! In my eyes best option is to remove encryption  immediately after downloading the file.

rmcrackan:
> In my eyes best option is to remove encryption immediately after downloading the file.

Agreed. The user-facing part of my program only gives the option to "Liberate" which is my catch-all term for download + decrypt.

Mbucari:
> 
> 
> This is AWESOME. For the first time -- aaxc decryption. I successfully tested on an audible plus and non-audible plus title
> 
> See all prev notes about minutia and the one important one about those nulls during getLicenseResponseDecryptionKey . After these are in and the merge is complete, I've got some ideas based on your changes before we release it into the wild.

Fixed minutia in Libation. nulls fixed in AudibleApi.

Mbucari:
> 
> 
> @rmcrackan : Yeah, the method used to download AAXC files are able to download every sort of Audible books. It's the same method which the iOS and Android Audible App uses. The drawback is that this method needs a unique key/IV pair for every file to decrypt them. You have to find a good method to store them for later use. If you loose them, you have to download the file again (if your account is valid or paid for Audible Plus titles)! In my eyes best option is to remove encryption immediately after downloading the file.

The decrypted license also has an expiration date. From my very limited testing, it looks like purchased content never expires (expires in year 3000) and audible Plus content expires when the billing cycle renews. So if you still have the aaxc file but need to get the decryption key again, you should be able to just request a license and use the returned decryption keys if you do it before the expiration date.

mkb79:
The key/IV and file combination can't expire. That's not possible. The expire date are only instruct the iOS or Audible App when they have to stop playing downloaded content.

Mbucari:
> 
> 
> The key/IV and file combination can't expire. That's not possible. The expire date are only instruct the iOS or Audible App when they have to stop playing downloaded content.

But Audible could change the keys on their end. As a hypothetical, say I add an Audible Plus book to my library and download the aaxc file in June. I also cancel my Plus membership in the same month. Then in August I reactivate my Plus membership. If I request a license for that Audible Plus book I added in June, will Audible send me the same axxc file encrypted with the same keys, or will the serve up a new aaxc with new keys?

mkb79:
> > The key/IV and file combination can't expire. That's not possible. The expire date are only instruct the iOS or Audible App when they have to stop playing downloaded content.
> 
> But Audible could change the keys on their end. As a hypothetical, say I add an Audible Plus book to my library and download the aaxc file in June. I also cancel my Plus membership in the same month. Then in August I reactivate my Plus membership. If I request a license for that Audible Plus book I added in June, will Audible send me the same axxc file encrypted with the same keys, or will the serve up a new aaxc with new keys?

That's a good question where I don't have an answer yet. But if you have the AAXC file and the key/iv pair you can decrypt the file at any time even your account are not exist anymore.

Mbucari:
> 
> 
> > > The key/IV and file combination can't expire. That's not possible. The expire date are only instruct the iOS or Audible App when they have to stop playing downloaded content.
> > 
> > 
> > But Audible could change the keys on their end. As a hypothetical, say I add an Audible Plus book to my library and download the aaxc file in June. I also cancel my Plus membership in the same month. Then in August I reactivate my Plus membership. If I request a license for that Audible Plus book I added in June, will Audible send me the same axxc file encrypted with the same keys, or will the serve up a new aaxc with new keys?
> 
> That's a good question where I don't have an answer yet. But if you have the AAXC file and the key/iv pair you can decrypt the file at any time even your account are not exist anymore.

I'm going to be able to test this in a few days. I've saved the aaxc file and license_response from an Audible Plus book and I cancelled my membership. It will expire in a few days, after which I'll reactivate my membership and see if Audible sends me the same key with a new expiration date.

Mbucari:
> 
> 
> Have you confirmed that decrypt still works without the cygwin files. I honestly don't remember why I added these years ago. I just know they were needed at the time.

They were

Mbucari:
> 
> 
> Have you confirmed that decrypt still works without the cygwin files. I honestly don't remember why I added these years ago. I just know they were needed at the time.

Yes. Those were all mp4trackdump dependencies. ffmpeg now does the decrypting.

rmcrackan:
> see if Audible sends me the same key with a new expiration date.

It seems like they'd have to do this. The keys are necessary for decryption. Since this happens on the client, it has to be independent of anything that audible could influence.

Mbucari:
> > if Audible sends me the same key with a new expiration date.
> 
> It seems like they'd have to do this. The keys are necessary for decryption. Since this happens on the client, it has to be independent of anything that audible could influence.

That's my guess too. But as I said earlier, I think if you re-request a license for the same title before the expiration date, Audible sends the same AAXC file with same keys. That is, they don't dynamically generate a new audiobook with new keys for each request. This is good because if you have an AAXC file but lost the keys, you can just request the license again to get the keys and you don't have to download the AAXC file again.

Mbucari:
> 
> 
> > > The key/IV and file combination can't expire. That's not possible. The expire date are only instruct the iOS or Audible App when they have to stop playing downloaded content.
> > 
> > 
> > But Audible could change the keys on their end. As a hypothetical, say I add an Audible Plus book to my library and download the aaxc file in June. I also cancel my Plus membership in the same month. Then in August I reactivate my Plus membership. If I request a license for that Audible Plus book I added in June, will Audible send me the same axxc file encrypted with the same keys, or will the serve up a new aaxc with new keys?
> 
> That's a good question where I don't have an answer yet. But if you have the AAXC file and the key/iv pair you can decrypt the file at any time even your account are not exist anymore.

I tested it, and they DO NOT change the keys even after subscription lapses and renews.

Issue #3, opened 2020-03-03
Doesn't work with Audible Australia


Pirateguybrush:
No option for Australian region in setup.

rmcrackan:
I have no access to the Australia site but I can make some educated guesses for how to make it work.

The one piece of information I can't find via google is Audible Australia's Marketplace Id. Each audible region has its own ID. Can you go to audible.com.au, inspect, and search for marketplaceid . Hopefully you'll find either MarketplaceId or ObfuscatedMarketplaceId followed by a ~13 character capitalized piece of text that you can report back to me.

rmcrackan:
I think I figured it out. Try the latest version: v3.1.8

Issue #2, opened 2020-02-17
unable to download library


draxal:
I saw mention of your solution on r/audible in a thread, and love the idea but I appear to be having an issue downloading the library.

When I try to import my library, it appears to login just fine and gets the book count and then starts failing. I believe this would be the relevant parts of the log file. Any ideas? 

------------------------------------------

2020-02-17 10:43:24.221 -07:00 [ERR] (at DtoImporterService.ImporterBase`1.Run(, )) Import error: post-validation importing
System.FormatException: Input string was not in a correct format.
   at System.Number.ThrowOverflowOrFormatException(ParsingStatus status, TypeCode type)
   at AudibleApiDTOs.Series.get_Index()
   at DtoImporterService.BookImporter.updateBook(Item item, Book book)
   at DtoImporterService.BookImporter.upsertBooks(IEnumerable`1 items)
   at DtoImporterService.BookImporter.DoImport(IEnumerable`1 items)
   at DtoImporterService.ImporterBase`1.Run[TResult](Func`2 func, T param)
2020-02-17 10:43:24.223 -07:00 [ERR] (at DtoImporterService.ImporterBase`1.Run(, )) Import error: post-validation importing
System.FormatException: Input string was not in a correct format.
   at System.Number.ThrowOverflowOrFormatException(ParsingStatus status, TypeCode type)
   at AudibleApiDTOs.Series.get_Index()
   at DtoImporterService.BookImporter.updateBook(Item item, Book book)
   at DtoImporterService.BookImporter.upsertBooks(IEnumerable`1 items)
   at DtoImporterService.BookImporter.DoImport(IEnumerable`1 items)
   at DtoImporterService.ImporterBase`1.Run[TResult](Func`2 func, T param)
   at DtoImporterService.ImporterBase`1.Import(T param)
   at DtoImporterService.LibraryImporter.DoImport(IEnumerable`1 items)
   at DtoImporterService.ImporterBase`1.Run[TResult](Func`2 func, T param)
2020-02-17 10:43:24.225 -07:00 [ERR] (at ApplicationServices.LibraryCommands+<ImportLibraryAsync>d__0.MoveNext()) Error importing library
System.FormatException: Input string was not in a correct format.
   at System.Number.ThrowOverflowOrFormatException(ParsingStatus status, TypeCode type)
   at AudibleApiDTOs.Series.get_Index()
   at DtoImporterService.BookImporter.updateBook(Item item, Book book)
[...]

Issue #1, opened 2019-12-04
Any way to select specific books?


webwizz2003:
Can you select specific books?  Am I missing something?  It looks like you have to back up all the books in your library at once.  It would be nice to be able to select the books you want to back up.

Thanks for making this.  It's coming along nicely.

rmcrackan:
Thanks for taking a look. I've got good news and bad news. You didn't miss anything, I just hadn't created that feature yet. I love the idea though so I created it today.

Now for the bad news: something about the latest Microsoft update (Visual Studio 16.4, Core 3.1) broke publishing. Ironically, this upgrade was supposed to be a *more* stable version of Core. I'm keeping an eye on an online thread where someone is already addressing this with them.

As soon as I can publish, this feature will be available in [v3.1-beta.9](https://github.com/rmcrackan/Libation/releases/tag/v3.1-beta.9). Or if you're comfortable with code, you can download it and run locally to get this feature immediately.

PR #30, opened 2021-06-24
Complete overhaul of download and decrypt.

Mbucari:
This is a biggie.

First, a couple of things.

- I moved all decryption and audiobook building into FileLiberator. AaxDecrypter isn't being used for anything anymore. The main reason is because I wanted use some AudibleApiDTOs types and AudibleApi resources and didn't want to add new references in AaxDecrypter. And frankly it just seems like the decryption routines belong in FileLiberator. Obviously reasonable minds can disagree, so feel free to re-organize as you see fit.
- I left some work for you because it wasn't clear to me how BackupBook is integrated with the UI. 

Major changes:

1. DecryptBook and DownloadBook are gone, replaced by DownloadDecryptBook. It wasn't clear to me how BackupBook is integrated with the UI, so I made a DownloadBookDummy to replace DownloadBook and that step is esentially skipped.
2. AaxToM4bConverter replaced by AaxcDownloadConverter.
2. The aaxc file is now downloaded, decrypted, and remuxed with tags and chapters all in one operation. The new FFMpegAaxcProcesser handles all downloading, decryption and audiobook building. Ffmpeg is actually downloading the aaxc, so no http client is necessary.
3. Since the aaxc file is never on disk, I made an IFileAbstraction for TagLib so it can read all metadata from the remote file through http.
4. EncodingInfo is unused because TagLib Tags has that info already.
5. Deleted DownloadedChapters and instead just used the AudibleApiDTOs.ChapterInfo

To Do:

- Choose whether to move new code into AaxDecrypter or delete AaxDecrypter project.
- Remove DownloadBook step from BackupBook, and remove all UI elements relating to that step.
- Probably add some more useful log information in AaxcDownloadConverter.
- TEST!


wtanksleyjr:
The AAX tags include some tags that I don't see in M4B, for example the Release Date (Mutagen calls this "rldt"). Are those being preserved by some means so we don't lose them permanently when pruning the encrypted downloads?

Mbucari:
> 
> 
> The AAX tags include some tags that I don't see in M4B, for example the Release Date (Mutagen calls this "rldt"). Are those being preserved by some means so we don't lose them permanently when pruning the encrypted downloads?

They are not as of right now. It looks like the tag library currently used in the project can detect rldt but does not copy the data. You should probably open an issue for that.

rmcrackan:
This is going to take some thought. There are a lot of cool ideas here. It's also a major shift to some goals. Before I spend time with the PR, let's discuss these points so we're on the same page. Sorry, this is going to get long in order to be clear:

Download and Decrypt separation:

* Practical #1: this is the big one: if download works but anything goes wrong with decrypt, you don't have to have an expensive download every time you retry decryption. If a user doesn't have a fast connection, this is a big deal.

* Practical #2: small reason: this allowed the option to retain the aax (now aaxc) file -- see the RetainAaxFiles config flag which is a hidden setting requested by one user who, if I remember right, liked to use Libation to download then use custom settings in inAudible to decrypt

* Practical #3: small reason: it looks like there's some metadata we don't retrieve that the user may want. This could be solved if there's a way to export all possible metadata into a structured file.

* History: they used to actually be separate steps. No longer relevant.

* Philosophy: separation of concerns: each piece does only what it needs to. Download, Decrypt, and Backup logic are all separate and can be modified cleanly. When I introduced downloading pdf.s, that was its own piece and the workflow in BackupBook just needed an extra few lines to incorporate it into the workflow without touching DownloadBook or DecryptBook

* "DownloadDecryptBook" -- If we do scrap everything else about this separation, may as well make this the new "BackupBook". It's less of a mouthful

FileLiberator and AaxDecrypter:

The reason for the separation of FileLiberator and AaxDecrypter is hinted at, but certainly not explicit, in the folder structure. Projects generally only refer to projects in a higher folder. The lower you go, the greater the dependencies. The domain ignorant utilities in particular are stand-alone tools with clear separate concerns -- the api knows nothing about decryption and the decryption knows nothing about the api or database etc. Something which needs knowledge of both probably belongs somewhere domain aware such as FileLiberator (albeit, probably with a better name). I can easily write programs against projects in the domain ignorant folder. For instance, until recent changes I had some very lightweight apps for querying the audible api and using AaxDecrypter for quick api tests which also supported full decryption. This was possible with no db and no domain logic. This gets ugly once you get down to the 'db aware' folder -- you need the full database and scaffolding of Libation.

Mbucari:
>     * Practical 1: this is the big one: if download works but anything goes wrong with decrypt, you don't have to have an expensive download every time you retry decryption. If a user doesn't have a fast connection, this is a big deal.

This should never be a problem. Since downloading, decrypting and remuxing aren't merely combined into one step, they're all one ffmpeg operation. If any part fails, the whole operation should fail. Also, I've been playing with ffmpeg more and figurted out a way to do it in one command so no piping is required. The only one way I could see the entire file downloading but failing to decrypt is if you passed the wrong decryption keys. The old AAX files had a hash so you could check that the activation bytes were valid, but aaxc doesn't (or ffmpeg doesn't support it). So it's possible to pass the wrong decryption keys and have ffmpeg read a garbled audio stream.  However, ffmpeg detects errors in the aac bitstream, and there are a LOT of errors if you supply the wrong decryption key. It would be simple to cancel the operation if an error is detected.

>     * Practical 2: small reason: this allowed the option to retain the aax (now aaxc) file -- see the RetainAaxFiles config flag which is a hidden setting requested by one user who, if I remember right, liked to use Libation to download then use custom settings in inAudible to decrypt

If we fix the metadata problem then the decrypted book should be identical to the aaxc file, but decrypted.. See my response to 3. You could add an option in settings to add downloaded chapter titles, and if that option isn't selected the output m4b file is literally a decrypted aaxc file - identical in every other way.

>     * Practical 3: small reason: it looks like there's some metadata we don't retrieve that the user may want. This could be solved if there's a way to export all possible metadata into a structured file.
I'm working on this. I've found a set of options that will essentially just decrypt the book in-place with all cover art, extra data stream(s), and ALL metadata. The output is a decrypted aaxc file with nothing else changed. The problem is there are some tags that FFMpeg doesn't recognize. It can copy them to output without knowing what they are, but it can't edit them or set them. So when you try to edit just the chapters (because aaxc file chapters don't match audible api chapters), it will overwrite the metadata section and overwrite all unknown tags. Even TagLib doesn't recognize all of them.

>     * Philosophy: separation of concerns: each piece does only what it needs to. Download, Decrypt, and Backup logic are all separate and can be modified cleanly. When I introduced downloading pdf.s, that was its own piece and the workflow in BackupBook just needed an extra few lines to incorporate it into the workflow without touching DownloadBook or DecryptBook

Think about the download-decrypt process as one step, like receiving data over a secure socket connection. You use an HttpClient in the project to download over https, but you wouldn't think about downloading the encrypted content then trying to decrypt it yourself. That's what's happening under the hood, but that step should be transparent to you, the user. I believe the same is true of audiobooks, especially if we can create a truly identical local copy, just decrypted, and leave the chapterize function as an option.

>     * "DownloadDecryptBook" -- If we do scrap everything else about this separation, may as well make this the new "BackupBook". It's less of a mouthful

Not a bad idea.

> 
> FileLiberator and AaxDecrypter:
> 
> The reason for the separation of FileLiberator and AaxDecrypter is hinted at, but certainly not explicit, in the folder structure. Projects generally only refer to projects in a higher folder. The lower you go, the greater the dependencies. The domain ignorant utilities in particular are stand-alone tools with clear separate concerns -- the api knows nothing about decryption and the decryption knows nothing about the api or database etc. Something which needs knowledge of both probably belongs somewhere domain aware such as FileLiberator (albeit, probably with a better name). I can easily write programs against projects in the domain ignorant folder. For instance, until recent changes I had some very lightweight apps for querying the audible api and using AaxDecrypter for quick api tests which also supported full decryption. This was possible with no db and no domain logic. This gets ugly once you get down to the 'db aware' folder -- you need the full database and scaffolding of Libation.

Yeah, that makes sense. All that's really needed is to just keep the old chapters type and pass the User-Agent as a parameter to the Create function. Actually, I discovered that ffmpeg supplies a default user-agent that seems to work, so I don't think it's necessary anymore.

rmcrackan:
I tried to be less verbose this time :) Was moderately successful

* Practical 1

RM: it's crazy that this is possible. If you can give it some solid testing, I'm willing to take the leap

* Practical 2

RM: sounds good. Predicated on 1 above

* Practical 3

RM: if this is irreconcilable, possibilities:

1) Can we export all metadata into a new json file? If this isn't possible because they are not recognized by ffmpeg, then:

2) New setting. First draft: "Chaperize"

- (default) SmartFix -- Just give me my stuff! Libation will try to fix tiny things like chapters name.  
- DecryptOnly -- Power users who want to crunch their own metadata. The book is decrypted only. All metadata will remain untouched, warts and all. This will leave in place all the obscure pieces of metadata which SmartFix could overwrite.

* Philosophy

RM: Makes sense. It's a new paradigm: you aren't combining 2 steps because there aren't 2 steps.

* FileLiberator and AaxDecrypter

RM: still feels like 3 libraries to me:

1) AudibleApi: All communication with audible. Provide C# wrapper. No dependencies.

2) Decryption: Output is the decrypted audio file and other misc files like cue, nfo. Input is anything needed to create them. Ideally, data gotten from the api could be the input here with no mandatory coupling. Even if I'm wrong and the new paradigm requires the decrypter to directly call the api, it feels like it should still be independent of Libation. No dependencies except for ffmpeg et al.

3) Libation: content/asset management. database. UI. settings. account management. 1-click options for things like refreshing library, download + decrypting books. search engine. etc...

Thoughts?

Mbucari:
OK, I think I've addressed everything.
Downloader seems very stable.
All aaxc metadata, except for chapter markers and titles, is restored in the m4b. If you want to preserve chapters too, that's easily accomplished with a setting and an existing FFMpegAaxcProcesser.ProcessBook overload.
All download code move back into AaxDecrypter, and no references to api needed.

Mbucari:
One thing to note on chapters: with these changes, if you want to add the downloaded chapters to the M4B you have to do it and download time. I could add that as an optional step later in the process so users could decide if they wanted their M4B metadata to exactly match the AAXC or if they want proper chapters. The reason I didn't do that is because I think adding chapters to the finished book could necessitate remuxing the whole file if there's not enough space in the allocated metadata section. That is a time-consuming process, especially for some big books like the 1.3 GB one I was testing.

rmcrackan:
This is all sounding really good. I'll dive into the new version when I'm at my work computer. Thank you for the separation between the api and decrypter.

Strangely, I received the email regarding chapters but I don't see those comments in this thread where github says they're supposed to be. I'm going to paste it again here so I can respond.

"One thing to note on chapters: with these changes, if you want to add the downloaded chapters to the M4B you have to do it and download time. I could add that as an optional step later in the process so users could decide if they wanted their M4B metadata to exactly match the AAXC or if they want proper chapters. The reason I didn't do that is because I think adding chapters to the finished book could necessitate remuxing the whole file if there's not enough space in the allocated metadata section. That is a time-consuming process, especially for some big books like the 1.3 GB one I was testing."

Sure, a setting is fine. Can you try it on one huge book and see if it blows up your resources? I don't think it's worth much investigation, but now I am curious.

FileLiberator/FileLiberator.csproj
```diff
@@ -13,4 +13,10 @@
     <ProjectReference Include="..\InternalUtilities\InternalUtilities.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="taglib-sharp">
+      <HintPath>DecryptLib\taglib-sharp.dll</HintPath>
```
rmcrackan, line 18: Obviously there's been a lot of churn lately so I want to clarify: this PR removed AtomicParsley and postproc-54.dll from the decrypter and added taglib-sharp.dll to the domain aware FileLiberator. I imagined FileLiberator going through the domain ignorant decrypter to do this.

FileLiberator/AaxcDownloadDecrypt/DownloadBookDummy.cs
```diff
@@ -0,0 +1,19 @@
...
+using Dinah.Core.ErrorHandling;
+using System.Threading.Tasks;
+
+namespace FileLiberator.AaxcDownloadDecrypt
+{
+    public class DownloadBookDummy : DownloadableBase
+    {
+        public override async Task<StatusHandler> ProcessItemAsync(LibraryBook libraryBook)
+        {
+            return new StatusHandler();
```
rmcrackan, line 11: minutia: Task.FromResult creates a "pre-computed" task which avoids incurring some the async state machine's overhead:

public override /* also 'async' no longer necessary */ Task<StatusHandler> ProcessItemAsync(LibraryBook libraryBook) => Task.FromResult(new StatusHandler());

rmcrackan:
I looked closer at the code today. Testing to come soon. This looks really good. Final points before merge:

1) see question about taglib-sharp

2) I agree about that setting to retain raw decrypt vs smart fix wherein we do not guarantee 100% metadata retention. Can you add that to settings and UI?

3) With the deletion of DecryptBook.cs, the RetainAaxFiles setting is now unused. It was a hidden setting anyway and was used by only 1 user. I can't find the conversation which was its impetus so I don't recall the feature's orig purpose. The user either wanted it because of metadata (which we'll solve with 2) or because they wanted to do their own converting without additional loss (and since we're keeping it pure, they still can). All this to say, can you kill Configuration.cs lines 86-92

Mbucari:
1) Correct about taglib. It's in AxxDecrypter and doesn't need to be exposed to FileLiberator. Amazingly, when I use taglib to copy all metadata from aaxc to m4b, the cover art comes with it. So atomicparsely is not needed. I don't know what postproc-54.dll belonged to, but it's not needed.

2) Raw decrypt still strips metadata. FFMpeg can not write metadata tags that it doesn't understand. But that's OK, because taglib can copy metadata it doesn't understand. That's what step 3 is doing, and it should always be done. So the only choice the user should make is whether Libation inserts chapter info retrieved by the api or retain chapter info embedded in the aaxc file. If they choose to retain aaxc chapters, then the output should be just a descrpted audiobook with all metadata and cover art. 

I'll try to add that option to settings.

3) will do
---

So I've been trying to figure out this cover art problem. I've figured out a workaround, but I still don't have a "good" solution. I narrowed down the cause and can reproduce it. Basically, if you use ffmpeg to just decrypt the aaxc and mux into m4b in one operation, most metadata and the cover art are retained in the file. We can use tablib to copy the missing metadata. This works flawlessly on my windows machine. Windows explorer shows the cover art in a thumbnail, and my media player (potplayer) loads the art too. But when I copy the file to my android phone, CX File Explorer doesn't load the cover art and neither does Smart Audiobook Player. Android VLC player does load cover art, though. Interestingly, if you copy the encryprted aaxc file to android and change the file extension to m4b, CX File Explorer DOES load the cover art.

The weir part is, if I
1) decode the audio to a raw aac stream
2) pipe that aac stream to a new ffmpeg instance
3) tell the new ffmpeg instance to load metadata and cover art from the aaxc file

...then it looks the same to Windows, but android has no trouble reading the cover art. So I think I'm going to go back to piping in FFMpegAaxcProcesser for now, and I'll ask about this in ffmpeg forums.

rmcrackan:
I just love metadata standards. and there are so many to love. I still get nightmares from my ambitious attempts about 15 years ago to create a universal mp3 tag editor. It ended it enlightenment, anger, and disgust in equal measure.

I'll hold off on a final review until you've resolved the metadata stuff to your satisfaction. I'm deferring that part completely to you since by now you're clearly the expert between the 2 of us.

Mbucari:
OK, I think it's ready for your full review.
Metadata and cover art is working flawlessly (albeit without an ideal solution).
Added user option to download chapter info and added migration to migrate_to_v5_0_0

One last think I'd like to add before going to release is an estimated remaining time to the downloader.

rmcrackan:
> One last think I'd like to add before going to release is an estimated remaining time to the downloader.

Agreed. Progress bars are hardly essential, but they sure are useful. I haven't yet looked at your details so I don't know if this is still in place -- my download progress bar included a timer so if there was no progress in a certain amount of time (I think it was a minute), an ugly red label would display an update of how long it's been frozen. I found this especially useful on days where audible and/or my isp was acting up. Then if the download continued, the label would be hidden until it happened again.

Mbucari:
OK, I'm done for the day. Check it out.

rmcrackan:
This is big enough, and has had enough incremental reviewing, that I'm going to complete the merge then start testing instead of the other way around which is my typical workflow. Tweaks and fixes can still happen before the big release. Hopefully the full release can happen soon.

rmcrackan:
@Mbucari -- I pushed some small changes incl one which caused all decrypts to crash when the title contained a colon (which is most of mine apparently)

Bug:

I own this title but I believe it's also on audible plus: https://www.audible.com/pd/B002V5H6F4

The cover art isn't found. When it's time to set cover image, we pass in (byte[])null :

> CoverImageFilepathDiscovered?.Invoke(this, aaxcDownloader.CoverArt); // cover art is null

> DecryptForm.SetCoverImage(null) // throws when attempting to turn byte[] => Image

Bug 1: there must be some other weird place that cover art can hide

Bug 2: when we're unable to find cover art, I imagine we'd still want to decrypt the book

Mbucari:
> 
> 
> @Mbucari -- I pushed some small changes incl one which caused all decrypts to crash when the title contained a colon (which is most of mine apparently)
> 
> Bug:
> 
> I own this title but I believe it's also on audible plus: https://www.audible.com/pd/B002V5H6F4
> 
> The cover art isn't found. When it's time to set cover image, we pass in (byte[])null :
> 
> > CoverImageFilepathDiscovered?.Invoke(this, aaxcDownloader.CoverArt); // cover art is null
> 
> > DecryptForm.SetCoverImage(null) // throws when attempting to turn byte[] => Image
> 
> Bug 1: there must be some other weird place that cover art can hide
> 
> Bug 2: when we're unable to find cover art, I imagine we'd still want to decrypt the book

I just checked that file, and it does not have cover art embedded in it. We could get cover art through the API and add it afterwards (this would probably be another option for the user). But we may want to pass the small cover art image used in the library to DownloadDecryptBook so it can use that instead of relying on it being in the aaxc file.

rmcrackan:
> I just checked that file, and it does not have cover art embedded in it. We could get cover art through the API and add it afterwards (this would probably be another option for the user). But we may want to pass the small cover art image used in the library to DownloadDecryptBook so it can use that instead of relying on it being in the aaxc file.

To keep it simple, let's just re-use that 1 setting: either the user wants us to fix this stuff or they don't. To that end, we could name it something better later.

PR #33, opened 2021-07-01
Work with updated Api Dtos and some minor readability edits.

Mbucari:


FileLiberator/DownloadPdf.cs
```diff
@@ -1,5 +1,4 @@
-﻿using System;
```
rmcrackan, line 1: Don't be surprised if I leave or re-add system, linq, or generic collections. It's not passive aggressive. These are the base tools I often expect from intellisense

AaxDecrypter/AaxcDownloadConverter.cs
```diff
@@ -147,13 +142,14 @@ public bool Step2_GetMetadata()
             RetrievedTags?.Invoke(this, aaxcTagLib);
             RetrievedCoverArt?.Invoke(this, coverArt);
 
-            return !isCanceled;
```
rmcrackan, line 150: Thanks, git. Wouldn't want that change to slip under the radar

Mbucari: Oops! I made last change as part of the last PR, and I also started working with the new DTOs. I reverted the DTO change before creating that PR, and I guess I took these out too.

rmcrackan: I was ragging on the git reporting, not you. For some reason it showed me the diff that this idential line was removed and added. To me, that should have been ignored. I've really got to search for an ignore-whitespace-only feature on github. It kills you with noise

wtanksleyjr: Not intending to get in the way, but when you reach the point of needing some alpha-testing I've got some experience both with conversion and coding (although not the languages you're using), and since I've got my own script-based AAX system running I'm both motivated to upgrade (AAXC is way better!), as well as able to double-check various results.

rmcrackan: That'd be awesome. I'm actually going to release 5.0 soon. Test away! I intend to silently upgrade for now and make an announcement on reddit after a bit more testing.

Of particular interest to you will be the new setting: Settings > Basic Settings > Allow Libation to fix up audiobook metadata. The name may later get some polish and spin but the idea is to support power users like you.

* Enabled (default): Libation attempts to clean up some fractured or incomplete metadata like cover art and chapters. Best efforts are made to retain all metadata but this is an overwrite/destructive process.
* Disabled: Libation only decrypts and all metadata is untouched.

PR #34, opened 2021-07-02
Minor code fixes and simplification

Mbucari:
- Added try blocks around Cue and NFO creation. If either throws an exception for any reason, the downloaded book will remain in temp file purgatory.
- Simplified Chapters constructor from aax file by using json instead of xml.

### DIFF
--- a/AaxDecrypter/AaxcDownloadConverter.cs
+++ b/AaxDecrypter/AaxcDownloadConverter.cs
@@ -211,13 +211,27 @@ namespace AaxDecrypter
 
         public bool Step5_CreateCue()
         {
-            File.WriteAllText(PathLib.ReplaceExtension(outputFileName, ".cue"), Cue.CreateContents(Path.GetFileName(outputFileName), downloadLicense.ChapterInfo));
+            try
+            {
+                File.WriteAllText(PathLib.ReplaceExtension(outputFileName, ".cue"), Cue.CreateContents(Path.GetFileName(outputFileName), downloadLicense.ChapterInfo));
+            }
+            catch (Exception ex)
+            {
+                //TODO Add log entry
+            }
             return !isCanceled;
         }
 
         public bool Step6_CreateNfo()
         {
-            File.WriteAllText(PathLib.ReplaceExtension(outputFileName, ".nfo"), NFO.CreateContents(AppName, aaxcTagLib, downloadLicense.ChapterInfo));
+            try
+            {
+                File.WriteAllText(PathLib.ReplaceExtension(outputFileName, ".nfo"), NFO.CreateContents(AppName, aaxcTagLib, downloadLicense.ChapterInfo));
+            }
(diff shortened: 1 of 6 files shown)


LibationLauncher/Program.cs
```diff
@@ -213,9 +213,10 @@ private static void migrate_to_v5_0_0()
 		{
 			var persistentDictionary = new PersistentDictionary(Configuration.Instance.SettingsFilePath);
 
-			if (persistentDictionary.GetString("AllowLibationFixup") is null)
+			var config = Configuration.Instance;
+			if (persistentDictionary.GetString(nameof(config.AllowLibationFixup)) is null)
             {
-				persistentDictionary.Set("AllowLibationFixup", true);
+				persistentDictionary.Set(nameof(config.AllowLibationFixup), true);
```
rmcrackan, line 219: Thanks. I'm a big fan of nameof. And: fewer string literals :)

AaxDecrypter/AaxcDownloadConverter.cs
```diff
@@ -211,13 +211,27 @@ public bool Step4_RestoreMetadata()
...
         public bool Step5_CreateCue()
         {
-            File.WriteAllText(PathLib.ReplaceExtension(outputFileName, ".cue"), Cue.CreateContents(Path.GetFileName(outputFileName), downloadLicense.ChapterInfo));
+            try
+            {
+                File.WriteAllText(PathLib.ReplaceExtension(outputFileName, ".cue"), Cue.CreateContents(Path.GetFileName(outputFileName), downloadLicense.ChapterInfo));
+            }
+            catch (Exception ex)
+            {
+                //TODO Add log entry
```
rmcrackan, line 220: If you're not ready to write out the serilog call then better to not have a try/catch swallow it so that at least the next layer up can catch it.

AaxDecrypter/AaxcDownloadConverter.cs
```diff
@@ -211,13 +211,27 @@ public bool Step4_RestoreMetadata()
...
         public bool Step6_CreateNfo()
         {
-            File.WriteAllText(PathLib.ReplaceExtension(outputFileName, ".nfo"), NFO.CreateContents(AppName, aaxcTagLib, downloadLicense.ChapterInfo));
+            try
+            {
+                File.WriteAllText(PathLib.ReplaceExtension(outputFileName, ".nfo"), NFO.CreateContents(AppName, aaxcTagLib, downloadLicense.ChapterInfo));
+            }
+            catch (Exception ex)
+            {
+                //TODO Add log entry
```
rmcrackan, line 233: Ditto about the future serilog call

AaxDecrypter/AaxcDownloadConverter.cs
Mbucari: OK, I added an error log.

Issue #37, opened 2021-07-02
JSON export missing part of title


wtanksleyjr:
I found an interesting bit of corrupt data in `B00CURDK9Y` (`Empires and Barbarians�� The Fall of Rome and the Birth of Europe`). I pasted that in from the tag itself as reported by mutagen, I can see two unknown characters, presumably bad UTF.

The interesting thing is that in the JSON export, the title cuts off with a single trailing space, so the subtitle is completely missing. Usually subtitle and title are colon-separated, but in this case there's that corrupt character in all of the title data.

https://www.audible.com/pd/Empires-and-Barbarians-Audiobook/B00CURDK9Y

Mbucari:
The Fall of Rome and the Birth of Europe is a subtitle, and the export doesn't (yet) export subtitles.

It is a Unicode character. Here's the mutagen readout:
`Empires and Barbarians\xa0: The Fall of Rome and the Birth of Europe (Unabridged)`

So I think @rmcrackan said he'd add subtitles to exports. As for the metadata tag itself, I'm inclined to leave them as-is instead of stripping data.

rmcrackan:
> I think @rmcrackan said he'd add subtitles to exports.

I do not remember rmcrackan saying this :)

I'm all for it if you'd like to give it a go. It seems likely that subtitles are already in a api/json field. Then it's just a matter of sticking it into the Book entity during library => db import

rmcrackan:
Closing this issue. Fixed in code. Will be included in next release

PR #40, opened 2021-07-03
Addressed Issue #37 and minor corrections

Mbucari:


DtoImporterService/BookImporter.cs
```diff
@@ -21,7 +21,7 @@ protected override int DoImport(IEnumerable<ImportItem> importItems)
 			new CategoryImporter(DbContext).Import(importItems);
 
 			// get distinct
-			var productIds = importItems.Select(i => i.DtoItem.ProductId).ToList();
```
rmcrackan, line 24: In several places in this PR, you removed ToList(). Would these areas benefit from deferred execution?

Mbucari: Probably not. It was just me imposing my style again, and for that I'm sorry!

rmcrackan: I understand the temptation. Use whichever convention you want on new code. I don't want to get into style wars though.

Issue #42, opened 2021-07-03
Libation crashes if Automated Backups form is closed before decrypted.


Mbucari:

rmcrackan:
This one was not yet addressed in the latest batch, correct?

Issue #39, opened 2021-07-03
Resume download if it gets interrupted.


Mbucari:
I think we should add the ability to resume downloading/decrypting the fire if it get interrupted (e.g. by unexpected restart, network error, or application error). Audible supports http Ranges so it's possible, and I actually don't think this would be too hard to do with a modified NetworkFileAbstraction.

rmcrackan:
Cool. I didn't know they supported that.

rmcrackan:
Wow, that was less of a code change than I thought there would be for this.

Issue #35, opened 2021-07-02
Thoughts about UI process of liberating single book


wtanksleyjr:
1. The UI affordance to liberate a specific book is extremely non-obvious on first download (why would clicking a stoplight make anything "go"?). What if it depicted a red button labeled with a download icon instead? The green light could become a green checkmark. If you have a yellow icon, a yellow-triangle would be iconic.
2.  When I click the stoplight to liberate a specific book, I find it extremely jarring that I have to answer "no" in order to get the software to do what I feel like I asked it to do. Would it work to rename the buttons from "yes/no" to "liberate all/liberate one"?
3. Additionally, the fact that the default choice (if I press Enter) is to process everything would be pretty bad for my 1500-book library -- perhaps some UI code to see how many books are pending and set the default to only the current one if there are more than, say, 10 books pending. (This would only matter if the user pressed "Enter", which happens to be how I like to interact with that kind of dialogue.)

rmcrackan:
1. The stoplight is going to get retired soon eventually because we've streamlined the app to no longer need the 'yellow' step -- so this metaphor no longer works. I like your ideas
2. Again: good idea
3. I don't want to put too much extra logic into this but I like the idea of switching the default

wtanksleyjr:
The solution in that commit looks fantastic!

PR #47, opened 2021-07-04
Addressed two issues and some minor fixed.

Mbucari:
Durably fixed Issue #42 .
Addressed #35 how we discussed, but I also found a use for the yellow light. Since adding resumable downloads as discussed in #39, the aaxc file is cached. Yellow indicates that a partial download exists.
Other minor fixes.

### DIFF
--- a/LibationWinForms/ProductsGrid.cs
+++ b/LibationWinForms/ProductsGrid.cs
@@ -182,15 +182,7 @@ namespace LibationWinForms
                 return;
             }
 
-            // not liberated: liberate
-            var msg
-                = "Liberate entire library instead?"
-                + "\r\n\r\nClick Yes to begin liberating your entire library"
-                + "\r\n\r\nClick No to liberate this book only";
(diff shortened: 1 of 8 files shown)


Mbucari:
Note, everything seems to work fine if you just have one decrypter window working at a time. You can fire up multiple instances, but the second instance to close causes a crash because it's trying to set a text box that's been disposed of. I haven't really explored what's going on under the hood too much, but obviously this is undesired behavior and should be addressed.

LibationWinForms/ProductsGrid.cs
```diff
@@ -182,15 +182,7 @@ private async void liberate_Click(object sender, DataGridViewCellEventArgs e)
                 return;
             }
 
-            // not liberated: liberate
```
rmcrackan, line 185: I concede that your solution is probably what we need here.

The original idea was: I assumed if people clicked the stop light, it was because it's a big obvious thing and they didn't know about liberating their entire library since that's in a less obvious menu. To me 'full library' seems less obvious and more preferable. Sounds like my intuition hasn't been in sync with others' experience though.

PR #51, opened 2021-07-11
Replaced FFMpeg decryptor and taglib with AAXClean

Mbucari:
OK, I think she's ready for Prime Time!

[AAXClean ](https://github.com/Mbucari/AAXClean) decrypts on the fly from an input stream to an output stream. It supports adding custom chapters and file-level metadata, but by default, it retains all file-level and stream-level metadata. The FFMpeg decryption speedup was netting me ~1300x - 2000x increase. AAXClean gets me ~5000x - 8000x. I backed up 22GB of books in 8:42.  But **most importantly**, cover art displays on android!

If you use this, you won't need FFMpeg or taglib anymore.


rmcrackan:
Very cool looking project! And it jives perfectly with my philosophies: let the expert do what they do best and loose coupling

> If you use this, you won't need FFMpeg or taglib anymore.

I don't see these in your new project. How will these dependencies go away?

Mbucari:
`How will these dependencies go away?`
At it's core, AAXClean is a specialized Mpeg-4 parser. It implements just enough of those specifications. To read and decrypt aavd audio streams, and read / write chapter info and file-level metadata.

rmcrackan:
Makes sense. Are you concerned with manually keeping up with future ffmpeg progress rather than depending on them? If yes, feel free to remove those dependencies when you’re confident you’re done with them

RE mp3s: I assumed single pass would just be some ffmpeg arg.s, but doing it as a 2nd pass isn’t a bit deal

Mbucari:
I'm keeping ffmpeg for now because of future mp3 conversion. But it's an awfully big dependency for just that.

Regarding mp3 conversion, how do you envision it working? Will it be a checkbox in settings that, if checked, only spits out mp3's and no M4B's? Or are you thinking a separate action that will batch convert M4B's to mp3? Also, convert to one mp3 file or split the book into multiple mp3 files by chapters?

rmcrackan:
> I'm keeping ffmpeg for now because of future mp3 conversion. But it's an awfully big dependency for just that.
> 
> Regarding mp3 conversion, how do you envision it working? Will it be a checkbox in settings that, if checked, only spits out mp3's and no M4B's? Or are you thinking a separate action that will batch convert M4B's to mp3? Also, convert to one mp3 file or split the book into multiple mp3 files by chapters?

Just saw these questions. See conversation in FB messager

PR #52, opened 2021-07-11
Minor fix.

Mbucari:
Need to retrieve ChapterInfo from the decrypted Mp4File in case Libation isn't fixing up chapters.


Mbucari:
Be sure to pull in new changes from AAXClean.

PR #53, opened 2021-07-13
Removed Ffmpeg and taglib. Updated AAXClean.

Mbucari:
I did a lot of tweaking on AAXClean and it appears very stable. It's possible there's a book out there that will not work with it, but I've tested my entire library with it without incident.

Next step is mp3 encoder using lame.


wtanksleyjr:
I'm willing to run my 1500 books through this again, so when we're ready
for a release let's go!

Last time it took 3 full nights and every time I stopped it (to let the
kids use the Windows machine) the book I was on joined a queue of "retry"
books, which at the time was a problem because they actually never got
finished. Fortunately at some point you guys fixed that, so everything
finished. This will almost certainly get the same kind of testing, plus the
usual Kindle Unlimited Narration and family sharing.

-Wm

rmcrackan:
Too cool. Depending on which recent version of Libation you used, the next version may be faster still :) Thanks for putting it through the paces

Mbucari:
@wtanksleyjr 
Do you have any audio books that are larger than 4 GB? I don't know if any of that large even exist, but if they do I know for a fact they won't work. Test for any larger than 2 gigabytes also. They should work, but there's a chance they won't. Depends on the choices audible made while encoding.

rmcrackan:
My biggest is Sherlock Holmes ~1.7gb

wtanksleyjr:
I don't have any bigger than 4, but I have 1 that clocks in at 3.8 GB, I'll
check if my phone will play it.

Oh, and the Cycle of Arawn is 1.8G.

Maybe I'll buy that Wheel of Time boxset. (Just kidding, there's no such
thing.)

-Wm

wtanksleyjr:
Quick comment: my C drive ran out of space (due to kids installing games on
the SSD instead of the big HD), and Libation, which was using that both for
temp space and for config files, just instadied, no popup or anything. I
don't think that's a bug per se, but just making a quick observation.

-Wm

Mbucari:
Thanks! It's fast, right? My speed is network limited by my 700 Mbps connection.

wtanksleyjr:
It's fast, but I'm limited the same way you are. Maybe I could dig out one
of my really old laptops and see if I could get it to be CPU-bound :) . But
more seriously, this round of changes (to support aaxc) has made Libation
absolutely dominate the competition, including my own hand-tuned scripts.
This change is absolutely useful for elegance and removing redundancy, but
compared to the big strategy change before, it's a rounding error.

I haven't been able to test the 3.8GB file. Still pending, sorry -- I
forgot to restart the conversion process last time.

-Wm

Mbucari:
@wtanksleyjr 
How's it coming?
I hope you have an unlimited data plan, because I might ask you to do the same thing again soon lol.

wtanksleyjr:
Seriously, man! ;)

It's a TON faster, seriously. The best way to see how fast is to finish the
conversion and then add some more books (for example get some Kindle
Unlimited With Narration), and watch how fast it burns through them.

Yes, I've tested all the variations I can think of. I've found only one
problem: one specific book just won't download no matter what I do. I need
to get my stuff together to make a coherent report.

-Wm

PR #54, opened 2021-07-15
Added Mp3 support to AAXClean.

Mbucari:
AAXClean can now decrypt directly to mp3 or convert an existing m4b to mp3. Look at the readme and theese changes for usage pattern.

Mp3s are encoded in average bitrate mode, with the target bitrate equal to the input aaxc file's average bitrate. I have no plans to offer customized encoding options, but it wouldn't be that hard to do.

Right now MP3s are one big file. I do plan to add support for splitting the mp3 into single chapter files.

### DIFF
--- a/AaxDecrypter/AaxcDownloadConverter.cs
+++ b/AaxDecrypter/AaxcDownloadConverter.cs
@@ -177,7 +177,7 @@ namespace AaxDecrypter
         {
             OutputFormat format = OutputFormat.Mp4a;
 
-            DecryptProgressUpdate?.Invoke(this, int.MaxValue);
+            DecryptProgressUpdate?.Invoke(this, 0);
 
             if (File.Exists(outputFileName))
                 FileExt.SafeDelete(outputFileName);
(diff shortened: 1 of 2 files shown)


rmcrackan:
I don't see anything about mp3s. Is this PR just paving the way so future mp3 stuff is possible?

While you're in that file: I can't tell, is outDir still used? It creates a dir but I don't see where anything gets written into that dir

I thought I set alerts so I could see when you updated AAXClean but the github settings don't appear to support notifications for commits,. My workaround was to watch your commits in my rss reader. In case it's useful to you, you can watch my progress with https://github.com/rmcrackan/Libation/commits/master.atom , etc

Mbucari:
Correct, this is just paving the way for that. The option is there and right now it's set to only output in Mp4a. I leave it to you how to best integrate that.

I think you're right, outdir is no longer needed.

rmcrackan:
I cleaned up outdir. Is there a way to leverage your new mp3 feature for existing m4b files?

Mbucari:
Yes. Use 
new Mp4File(m4bfile).ConvertToMp3()

Or something like that. Not in a position to look at it now.

PR #69, opened 2021-07-24
Added logging of download license.

Mbucari:


AaxDecrypter/AaxcDownloadConverter.cs
```diff
@@ -51,6 +53,11 @@ public AaxcDownloadConverter(string outFileName, string cacheDirectory, Download
             downloadLicense = ArgumentValidator.EnsureNotNull(dlLic, nameof(dlLic));
             OutputFormat = outputFormat;
 
+            if (Serilog.Log.Logger.IsVerboseEnabled())
```
rmcrackan, line 56: No conditional required

Mbucari: Ok. I was just copying the logging pattern you used elsewhere.

rmcrackan:
> Ok. I was just copying the logging pattern you used elsewhere

No worries. I definitely use that pattern around non-logging things such as: if (is verbose) MessageBox.Show (warning message)

If I used it anywhere around logging commands, it was unintentional.

PR #72, opened 2021-07-27
Added a background file system scanner.

Mbucari:
AudibleFileStorage now uses the new BackgroundFileSystem to keep track of all files in the BooksDirectory. This will minimize the number of full file enumerations required, and all file enumerations will happen on a background thread.

### DIFF
--- a/FileManager/BackgroundFileSystem.cs
+++ b/FileManager/BackgroundFileSystem.cs
@@ -95,18 +95,24 @@ namespace FileManager
 
         private void RemovePath(string path)
         {
-            var pathsToRemove = fsCache.Where(p => p.StartsWith(path)).ToArray();
+            lock (fsCache)
+            {
+                var pathsToRemove = fsCache.Where(p => p.StartsWith(path)).ToArray();
 
-            foreach (var p in pathsToRemove)
-                fsCache.Remove(p);
+                foreach (var p in pathsToRemove)
+                    fsCache.Remove(p);
+            }
         }
 
         private void AddPath(string path)
         {
-            if (File.GetAttributes(path).HasFlag(FileAttributes.Directory))
-                fsCache.AddRange(Directory.EnumerateFiles(path, SearchPattern, SearchOption));
-            else
-                fsCache.Add(path);
+            lock (fsCache)
+            {
+                if (File.GetAttributes(path).HasFlag(FileAttributes.Directory))
+                    fsCache.AddRange(Directory.EnumerateFiles(path, SearchPattern, SearchOption));
+                else
+                    fsCache.Add(path);
+            }
         }
 
         #endregion
--- a/FileManager/BackgroundFileSystem.cs
+++ b/FileManager/BackgroundFileSystem.cs
@@ -1,0 +1,107 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace FileManager
+{
+    class BackgroundFileSystem
+    {
+        public string RootDirectory { get; private set; }
+        public string SearchPattern { get; private set; }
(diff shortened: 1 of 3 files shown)


FileManager/BackgroundFileSystem.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            //Wait for background scanner to terminate before reinitializing.
+            backgroundScanner?.Wait();
+            backgroundScanner = new Task(BackgroundScanner);
+            backgroundScanner.Start();
+        }
+
+        private void FileSystemWatcher_Error(object sender, ErrorEventArgs e)
+        {
+            Init(RootDirectory, SearchPattern, SearchOption);
```
rmcrackan, line 62: This confuses me

rmcrackan:
Funny. I was just today designing a way to get rid of nearly all of this in favor of tracking state. I'm on the fence about it. Yours looks like a good improvement on the current method.

FileManager/BackgroundFileSystem.cs
Mbucari: There's a possibility that FileSystemWatcher has an error while waiting for a change in the watched directory. If that happens, BackgroundFileSystem can't be sure it didn't miss a change, so we need to re-initialize the whole thing.

rmcrackan: Are you sure the old scanner can't survive and continue running?

FileManager/BackgroundFileSystem.cs
```diff
@@ -0,0 +1,117 @@
...
+            directoryChangesEvents?.Dispose();
+            fileSystemWatcher?.Dispose();
+
+            fsCache = Directory.EnumerateFiles(RootDirectory, SearchPattern, SearchOption).ToList();
+
+            directoryChangesEvents = new BlockingCollection<FileSystemEventArgs>();
+            fileSystemWatcher = new FileSystemWatcher(RootDirectory);
+            fileSystemWatcher.Created += FileSystemWatcher_Changed;
+            fileSystemWatcher.Deleted += FileSystemWatcher_Changed;
+            fileSystemWatcher.Renamed += FileSystemWatcher_Changed;
```
rmcrackan, line 49: Last time I worked with system file watcher, these were fired a lot, esp anywhere with a lot of changes. Instead of adding to the queue, I suggest add if not already present. Else the queue can contain a lot of duplicates and waste a lot of time scanning the same dir back to back

FileManager/BackgroundFileSystem.cs
Mbucari: https://docs.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher.error?view=net-5.0

"Occurs when the instance of FileSystemWatcher is unable to continue monitoring changes or when the internal buffer overflows."

Mbucari:
>I was just today designing a way to get rid of nearly all of this in favor of tracking state. 
It's a good idea, and it's got my vote.

rmcrackan:
> It's a good idea, and it's got my vote.

Thanks. Just the design summary is over 60 lines long so it's not a quick fix. After we finish the conversation about frequent queue-ing I'll push yours through and (hopefully) get around to mine soon-ish.

At the end of this weekend I'm going to be unavailable for almost 3 weeks so I don't want to rush through a major redesign of something like that.

FileManager/BackgroundFileSystem.cs
Mbucari: What I observed playing around with it is that there are only redundant firings if you subscribe to FileSystemWatcher.Changed, because it gets fired at the same time as Created, Deleted, and Renamed. Buf if you don't subscribe to created, then each of the other 3 events fire only once per change as expected. Renaming a file or directiory fires once. Deleting a directory, including all its contents, fires Deleted only once. Adding a directory, including one with contents, fires Created only once.

rmcrackan: Cool. Maybe that's what I had been doing wrong at the time

PR #74, opened 2021-07-28
Added Convert all M4b to Mp3 action.

Mbucari:
The "Liberate" tool strip menu now contains "Convert all M4b to Mp3". It uses the existing decryption infrastructure and is a minor change.  The only issue is how it integrates with the storage system, which is in flux right now.  Until you get that sorted, you may want to remove the tool menu entry in release builds.

One more thing, while making this I discovered that the m4b files made by AAXClean had an error in them.  It's nothing major and almost certainly wouldn't cause users any problems, but technically the files were incorrect. Depending how how fragilethe media player/editor is, it could cause a player crash at the end of the file. This error was introduced in [this commit](https://github.com/Mbucari/AAXClean/commit/11b4ff67ff2983d515539943f3d05096d21e680a) and was fixed [here](https://github.com/Mbucari/AAXClean/commit/bfaa450529ce97d62a1fbd3c0953a9a55a35b442).

### DIFF
--- a/FileLiberator/ConvertToMp3.cs
+++ b/FileLiberator/ConvertToMp3.cs
@@ -1,0 +1,108 @@
+﻿using AAXClean;
+using DataLayer;
+using Dinah.Core.ErrorHandling;
+using Dinah.Core.IO;
+using FileManager;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace FileLiberator
+{
+    public class ConvertToMp3 : IDecryptable
+    {
+
+        private Mp4File m4bBook;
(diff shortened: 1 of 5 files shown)


rmcrackan:
> The "Liberate" tool strip menu now contains "Convert all M4b to Mp3"

This looks like a perfect first version. If there's user feedback, any changes can grow from there.

> The only issue is how it integrates with the storage system, which is in flux right now.

No issue for me. This won't affect my migration plan.

> Until you get that sorted, you may want to remove the tool menu entry in release builds.

I'm going to do this so I can release it after my vacation and only release mostly bug fixes for the rest of the week.

> m4b files made by AAXClean had an error in them

I pulled your changes and will include it with the next release.

PR #85, opened 2021-08-12
A LOT of code churn. Mostly UI.

Mbucari:
1) Major ```ProductsGrid``` and ```GridEntry``` refactoring. The most significant changes are:
	1. ```GridEntry``` properties are now the display values, and sorting happens through a background implementation. This is the reverse of how ```GridEntry``` behaved before, and it allows for much less hacky UI code. It also enabled me to easily add sorting to the button columns. Unfortunately (or not), I had to modify ```SortableBindingList``` to work with the new approach. It's not all that different really, and if Libation is the only place you're using it then maybe you can update Dinah with the new version (```SortableBindingList2```) and its related dependencies.
	2. Added custom ```DataGridViewButtonColumn``` types to handle drawing images onto the buttons. This allowed me to add the 'Liberate' and 'Edit Tags' columns to the data source, so the ```DataGridView``` is now completely auto generated with no columns that aren't in the binding source.
	3. Because of i and ii, ```ProductsGrid``` no longer handles drawing/painting for the ```DataGridView```. These changes also pave the way towards a cross-platform UI.

2) I added ```RemoveBooksDialog```. It is a ```DataGridView``` of ```RemovableGridEntry``` (derrived from ```GridEntry``` with a ```bool``` '```Removed```' property). This view has a checkbox column for selecting books to remove. When loaded, it scans the Audible account(s) library and compares it to Libation's library.  The ```Removed``` checkbox will be checked for any books in Libation that weren't found in Audible. The user can check/uncheck any books s/he likes and remove them from the library.

3) Removed ```WindowsDesktopUtilities```. ```WinAudibleImageServer``` was only used in 2 places and was easily replaced by ```PictureStorage```. I modified ```PictureStorage``` by changing the ```PictureCached``` event handler to send ```PictureCachedEventArgs```, which contains the ```PictureDefinition``` and the picture in ```byte[]```, and by adding the ```GetPictureSynchronously``` method.  I also changed the background downloader to use a ```BlockingCollection<PictureDefinition>``` queue. I was occasionally getting crossthread exceptions in the old timer implementation, though Libation silently swallowed them. I also removed the 700ms delay between downloads, so all images download at pipe speed.

4) ```DownloadDecryptBook.RequestCoverArt``` event handler now works synchronously, so the cover art is retrieved before Step 2 begins. This will allow retrieved cover art to be added to MP3s, which wasn't available before.

5) Added xml filetype tags to top of all resx files in the hopes that VS will stop auto-generating huge comment blocks.

6) I noticed you use tabs instead of spaces in your editor, so I changed mine to tabs as well. Hopefully this will cut down on the number of diffs.

7) I redesigned ```ProcessorAutomationController```. The exposed methods are unchanged, but the internals have been completely redesigned. 

	1) ```BackupBook``` has been removed. PDFs are backed up by chaining after ```DownloadDecryptBook.Completed```
	2) The ```FileLiberator``` interfaces have changed as follows:

		* ```IDownloadable``` is now ```IStreamable```
		* ```IDecryptable``` is now ```IAudioDecodable```
		* ```IDownloadableProcessable``` is gone.  ```BackupBook``` was the only type that inherited ```IProcessable``` but not ```IStreamable```.  Without it, all ```IProcessable``` are also ```IStreamable```, so ```IProcessable``` now inherits ```IStreamable```.

		```IDecryptable.DecryptBegin```, ```IDecryptable.UpdateProgress```, ```IDecryptable.UpdateRemainingTime``` and ```IDecryptable.DecryptCompleted``` have been removed. Instead, ```IAudioDecodable``` implementers use ```IStreamable.StreamingBegin```, ```IStreamable.StreamingProgressChanged```, ```IStreamable.StreamingTimeRemaining```, and ```IStreamable.StreamingCompleted```.

	3) The new ```BookLiberation``` foms are:

		* ```LiberationBaseForm```
		* ```AudioDecodeForm``` (based on former ```DecryptForm```)
		* ```DownloadForm``` (based on former ```DownloadForm```)
		* ```PdfDownloadForm```
		* ```AudioDecryptForm```
		* ```AudioConvertForm```

		And this is their inheritance structure:
		```
		Form
		 |
		 V		
		LiberationBaseForm
		 |    |
		 |    V
		 |  DownloadForm
		 |    |
		 |    V
		 |  PdfDownloadForm
		 V
		AudioDecodeForm
		 |    |
		 |    V
		 |  AudioConvertForm
		 V
		AudioDecryptForm
		```

		```LiberationBaseForm``` subscribes to, unsubscribes from, and handles all events in ```IStreamable``` derived types.
		All Form event handlers are virtual and overridden as necessary by descendants.

	4) Inside ```ProcessorAutomationController```, there is a generic static creator for binding ```FileLiberator``` processors to ```BookLiberation``` forms called ```CreateProcessable```.
		```CreateProcessable``` creates a new ```IProcessable``` that is wired up to a new ```LiberationBaseForm``` on ```IProcessable.Begin```. The form is shown on ```IStreamable.StreamingBegin``` , is closed on ```IStreamable.StreamingCompleted```, and disposed on ```IProcessable.Completed```. ```CreateProcessable``` returns the new ```IProcessable```.

8) There are a couple of places where I needed to use Control.Invoke but couldn't. I ended up making SynchronizeInvoker to take care of those cases. One thing that would be nice is if you changed Dinah.Core.Windows.Forms.ControlExt.UIThread to accept ISynchronizeInvoke instead of a Control. It'll behave the same, but I'll be able to use that extension instead in these two cases. Or better still, make a new extension in Dinah.Core for multi-threading, since ISynchronizeInvoke isn't windows-specific like Control.

9) I improved thread safety throughout the project.

In closing, I think I paid down a LOT of technical debt in this one!


rmcrackan:
Holy cow. I'm honestly not sure where to even begin with reviewing this one.

Low hanging fruit: "I noticed you use tabs instead of spaces in your editor". I mostly use my work computer because of the better version of VS they provide. My previous team required spaces, current team requires tabs. So there's a mix of these styles throughout

Mbucari:
>  I'm honestly not sure where to even begin with reviewing this one.

3, 4, and 5 should be easiest to review.

3 is all contained in PictureStorage.
4 mostly affects AaxcDownloadConverter, but depends on the changes made in 3.

After those, I'd look at 1. Affected files are:

- GridEntry
- ProductsGrid
- AsyncNotifyPropertyChanged (new)
- DataGridViewImageButtonColumn (new)
- EditTagsDataGridViewImageButtonColumn (new)
- IMemberComparable (new)
- LiberateDataGridViewImageButtonColumn (new)
- MemberComparer (new)
- ObjectComparer<T> (new)
- SortableBindingList2<T> (new)
- SynchronizeInvoker (new, think of it as enabling Control.Invoke without a Control).

Most of the new files are very small, single-responsibility classes and should be easy to review. Start by looking at ProductsGrid, then GridEntry. ProductsGrid will lead you to ```SortableBindingList2<T>```, ```MemberComparer```, ```IMemberComparable```, and the ```DataGridViewImageButtonColumn``` and ```DataGridViewImageButtonCell``` classes.  Then look at GridEntry. It will lead you to everything else.

Then go on to 7, which is contained in the ```LibationWinForms.BookLiberation``` and ```FileLiberator``` namespaces. Start by looking at the ```FileLiberator``` interfaces, then move on to ```ProcessorAutomationController```, then to ```LiberationBaseForm```. Once you have a handle on how ```LiberationBaseForm``` interacts with the ```FileLiberator``` interfaces, the rest of the ```LibationWinForms.BookLiberation``` should be easy to grasp.

Mbucari:
I may be helpful to start looking at the new files instead of the diffs.

rmcrackan:
> removed the 700ms delay between downloads, so all images download at pipe speed

Early on, a lot of effort was made to throttle calls to amazon. They can be vicious about blocking the IPs of scrapers. I don't think getting rid of this throttle will hurt but it makes me a bit nervous.

rmcrackan:
> new extension in Dinah.Core for multi-threading, since ISynchronizeInvoke isn't windows-specific like Control

You're welcome to move it to the Dinah.Core repo if you want. I have no strong preference

Mbucari:
> They can be vicious about blocking the IPs of scrapers. I don't think getting rid of this throttle will hurt but it makes me a bit nervous.

My thinking is that, as you browse your audible library, your browser downloads cover images for all products on the page as quickly as it can. That's 20-50 thumbnails per page. The same thumbnails are also used in their store, where you can load dozens per page. So I don't think it's necessary, especially since images are cached and a large number of requests only happens the first time you import your library.  But if you decide to re-introducing throttling, I'd suggest a 1-2 second gap after every 50 downloads.

rmcrackan:
> you can update Dinah with the new version (SortableBindingList2) and its related dependencies

Done

rmcrackan:
The changes in 7 look good. It was definitely time to better centralize that controller wire up logic.

One piece of minutia: in CreateProcessable , completedAction subscribes to Completed even if null. I've never tried to subscribe a null before. Have to verified that it's not a problem?

Mbucari:
>  Completed even if null. I've never tried to subscribe a null before. Have to verified that it's not a problem?

I did test that, and it does not cause an exception. I'm assuming the runtime is using some variation of event?.Invoke() internally since Microsoft introduced that feature.

PR #87, opened 2021-08-20
Make sure DataGridView updates the display immediately after Details are changed.

Mbucari:
I made GridEntry.NotifyPropertyChanged public and use it from ProductsGrid to update rows as data are changed.

I also did a little refactoring an some minor fixes.

### DIFF
--- a/LibationWinForms/LiberateDataGridViewImageButtonColumn.cs
+++ b/LibationWinForms/LiberateDataGridViewImageButtonColumn.cs
@@ -22,12 +22,42 @@ namespace LibationWinForms
 
 			if (value is (LiberatedState liberatedState, PdfState pdfState))
 			{
-				(string mouseoverText, Bitmap buttonImage) = GridEntry.GetLiberateDisplay(liberatedState, pdfState);
+				(string mouseoverText, Bitmap buttonImage) = GetLiberateDisplay(liberatedState, pdfState);
 
 				DrawButtonImage(graphics, buttonImage, cellBounds);
 
 				ToolTipText = mouseoverText;
 			}
+		}
+
+		private static (string mouseoverText, Bitmap buttonImage) GetLiberateDisplay(LiberatedState liberatedStatus, PdfState pdfStatus)
+		{
+			(string libState, string image_lib) = liberatedStatus switch
+			{
+				LiberatedState.Liberated => ("Liberated", "green"),
+				LiberatedState.PartialDownload => ("File has been at least\r\npartially downloaded", "yellow"),
+				LiberatedState.NotDownloaded => ("Book NOT downloaded", "red"),
+				_ => throw new Exception("Unexpected liberation state")
+			};
+
+			(string pdfState, string image_pdf) = pdfStatus switch
+			{
+				PdfState.Downloaded => ("\r\nPDF downloaded", "_pdf_yes"),
+				PdfState.NotDownloaded => ("\r\nPDF NOT downloaded", "_pdf_no"),
+				PdfState.NoPdf => ("", ""),
+				_ => throw new Exception("Unexpected PDF state")
+			};
+
+			var mouseoverText = libState + pdfState;
+
+			if (liberatedStatus == LiberatedState.NotDownloaded ||
+				liberatedStatus == LiberatedState.PartialDownload ||
+				pdfStatus == PdfState.NotDownloaded)
+				mouseoverText += "\r\nClick to complete";
+
+			var buttonImage = (Bitmap)Properties.Resources.ResourceManager.GetObject($"liberate_{image_lib}{image_pdf}");
+
+			return (mouseoverText, buttonImage);
 		}
 	}
 }
--- a/LibationWinForms/ProductsGrid.cs
+++ b/LibationWinForms/ProductsGrid.cs
@@ -67,7 +67,7 @@ namespace LibationWinForms
 					await Liberate_Click(liveGridEntry);
 					break;
 				case nameof(liveGridEntry.DisplayTags):
-					Details_Click(liveGridEntry.LibraryBook);
+					Details_Click(liveGridEntry);
 					break;
 			}
 		}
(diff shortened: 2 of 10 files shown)


Mbucari:
Is there a need for both LiberatedStatus and LiberatedState? If not, can we ditch LiberatedStatus and just use LiberatedState and PdfState?

LibationWinForms/ProductsGrid.cs
```diff
@@ -149,10 +152,10 @@ public void Display()
 
 		public void RefreshRow(string productId)
 		{
-			var rowIndex = getRowIndex((ge) => ge.AudibleProductId == productId);
+			var liveGridEntry = getGridEntry((ge) => ge.AudibleProductId == productId);
 
-			// update cells incl Liberate button text
-			_dataGridView.InvalidateRow(rowIndex);
```
rmcrackan, line 155: Where is the row invalidated/re-drawn now?

rmcrackan:
> Is there a need for both LiberatedStatus and LiberatedState? If not, can we ditch LiberatedStatus and just use LiberatedState and PdfState?

Good question. There's a subtle difference as explained in the comment above LiberatedState. However, I fear it's too subtle. I'm in the middle of a lot of changes at the moment. After this, I think you're right and they should be combined. For now, please leave separate so it doesn't walk on my in-progress code

LibationWinForms/ProductsGrid.cs
Mbucari: GridEntry implements [INotifyPropertyChanged](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.inotifypropertychanged?view=net-5.0). When added to a BindidingList, the list subscribes to the PropertyChanged event in every GridEntry. Firing the [PropertyChanged](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.inotifypropertychanged.propertychanged?view=net-5.0) triggers the DataGridView to update the view for GridEntry. If a property name is specified (as it is below), DataGridView will only update the cell bound to that property. Since RefreshRow is only used to update display of the Liberate column, we only need to notify that the Liberate property has changed, and DataGridView will redraw it. If you want to notify all properties in the GridEntry, don't specify a property name.

Mbucari: RefreshRow might more accurately be named RefreshLiberate.

Mbucari:
> > Is there a need for both LiberatedStatus and LiberatedState? If not, can we ditch LiberatedStatus and just use LiberatedState and PdfState?
> 
> Good question. There's a subtle difference as explained in the comment above LiberatedState. However, I fear it's too subtle. I'm in the middle of a lot of changes at the moment. After this, I think you're right and they should be combined. For now, please leave separate so it doesn't walk on my in-progress code

What I'm trying to move towards is a true model-view-viewmodel implementation, where Library is the model,  ProductsGrid is the view, and GridEntry is the view model. Right now the view is coupled to the model too tightly. For example, prior to this PR, Details_Click updates the model, then it updates the view.  After this PR, Details_Click updates the model, then tells the view model it's been updated, then the view model tells the view to update. What ultimately should happen is Details_Click updates the view model, and then the view model update the model and notifies the view that it's changed. This is my goal once you finish up your storage redesign and I know what I'm working with.

Mbucari:
Fixed #88 in last commit. I only have one book with a PDF, so I didn't realize the error lol.

Mbucari:
Scratch that. You shouldn't need to create a new DownloadPdf IProcessable after each book. I need to fix chaining, but that last commit wasn't correct.

rmcrackan:
Thanks for the pdf fix.

The important parts of my redesign are complete.

I fear I've gotten myself into a pickle with LiberatedStatus, LiberatedState, PdfState. We need something to represent liberated, not liberated, error, and null/not-applicable (for pdf). We do not need this to represent partially-liberated since that is purely a function of whether there's a partial file present. Since these are enums, existing references to them in json or db uses their int value. I haven't looked yet but maybe we'll get lucky and I was smart with the int.s. Otherwise a migration will be needed to combine these.

In any case: my other stuff is done and you're free to mess with these. I'll be offline for a while but my goal is to ratify your PR and publish a new release today or tomorrow.

Mbucari:
This isn't making sense. After every book downloads, it tries to download a pdf by calling DownloadPdf.ProcessAsync. ProcessAsync first fires IProcessible.Begin, which creates a new PdfDownloadForm, and then calls ProcessItemAsync, wich ends up firing IStreamable.StreamingBegin which should show the form. But somehow, IProcessable.Begin isn't creating as new PdfDownloadForm form and the persistant DownloadPdf is accessing the old, disposed form. It doesn;'t make sense because DownloadDecryptBook uses AudioDecryptForm in the exact same way, but there's no problem with them. The error must be caused by how I'm chaining pdf download after the book download, but I don't see it and unfortunately I only have one book in my library that has a PDF, so I can't replicate the error. Can you try debugging this?

Mbucari:
OK, maybe this fixed it. I was unsubscribing all IStreamable events on FormClosed. If the form wasn't shown, it would never fire FormClosed and StreamingBegin would try to show the old, disposed form. I changed LiberateBaseForm to unsubscribe from all IStreamable events on Disposed.

LibationWinForms/LiberateDataGridViewImageButtonColumn.cs
```diff
@@ -22,12 +22,42 @@ protected override void Paint(Graphics graphics, Rectangle clipBounds, Rectangle
...
+				PdfState.Downloaded => ("\r\nPDF downloaded", "_pdf_yes"),
+				PdfState.NotDownloaded => ("\r\nPDF NOT downloaded", "_pdf_no"),
+				PdfState.NoPdf => ("", ""),
+				_ => throw new Exception("Unexpected PDF state")
+			};
+
+			var mouseoverText = libState + pdfState;
+
+			if (liberatedStatus == LiberatedState.NotDownloaded ||
+				liberatedStatus == LiberatedState.PartialDownload ||
```
rmcrackan, line 54: Oh, right. PartialDownload isn't a *saved* state, but it's still a representable condition. I guess it should also be included in the same enum so that places like this can use it

PR #90, opened 2021-08-22
Fully implemented the MVVM pattern

Mbucari:
1) ```LiberatedState``` and ```PdfState``` are gone. ```LiberatedStatus``` takes their place. ```LiberatedStatus``` has a ```PartialDownload``` value, but that value will not be written to the database.
2) ```UserDefinedItem``` has a static event, ```ItemChanged```, that fires when the values of ```Tags```, ```BookStatus```, or ```PdfStatus``` change. ```UserDefinedItem``` is responsible for checking that values have actually changed and will not fire ```ItemChanged``` unless there's actually been a change.
3) ```FileLiberator``` is now domain ignorant. When books or PDFs are downloaded, it sets ```UserDefinedItem``` statuses.
4) ```GridEntry``` (the view model), subscribes to ```UserDefinedItem.ItemChanged```. When it receives a change notification (from the model), it writes the change to the database and notifies ```ProductsGrid``` (the view) of the change so it can redraw.
5) With all that, there's no need to manually refresh the view after each download or after manually editing properties because the model notifies the view model of any changes, and the view model notifies the view.

Flowchart

Now that I look at this flow chart, it's more of a composite of MVVM and MVC


rmcrackan:
MVC and MVVM are both solid patterns and I have no problem doing a hybrid. They're both better than my previous MVWTF. Nice work. And nice diagram too. I feel embarrassed for leaving some of this ~5 year old slop in here for so long.

ApplicationServices/LibraryCommands.cs
```diff
@@ -161,37 +154,22 @@ private static async Task<int> importIntoDbAsync(List<ImportItem> importItems)
...
+
+		public static int UpdateTags(Book libraryBook, string newTags)
 		{
 			try
 			{
 				using var context = DbContexts.GetContext();
 
-				var udi = book.UserDefinedItem;
-
-				var tagsChanged = udi.Tags != newTags;
```
rmcrackan, line 173: The purity of this change is attractive. However we now have up to 3 discrete db calls -- if the user changes tags, book liberation, and pdf liberation. If we later add further editing on that form, each of those will also use a new db call as well as their own transaction.

FileLiberator/DownloadDecryptBook.cs
```diff
@@ -170,11 +173,14 @@ var dest
                     Cue.UpdateFileName(f, audioFileName);
 
                 File.Move(f.FullName, dest);
+
+                movedAudioFile |= AudibleFileStorage.Audio.IsFileTypeMatch(f);
```
rmcrackan, line 177: Good idea. I'm a bit fan of this kind of logic.

LibationWinForms/BookLiberation/ProcessorAutomationController.cs
```diff
@@ -305,7 +304,7 @@ public BackupLoop(LogMe logMe, IProcessable processable, AutomatedBackupsForm au
 		protected override async Task RunAsync()
 		{
 			// support for 'skip this time only' requires state. iterators provide this state for free. therefore: use foreach/iterator here
-			foreach (var libraryBook in Processable.GetValidLibraryBooks())
+			foreach (var libraryBook in Processable.GetValidLibraryBooks(ApplicationServices.DbContexts.GetContext().GetLibrary_Flat_NoTracking()))
```
rmcrackan, line 307: Curious about why you moved this here. It's certainly not right or wrong; I'm just curious about the choice.

ApplicationServices/LibraryCommands.cs
Mbucari: Since grid entry is now the only piece interacting with the database, I can expose Begin and End transaction methods to cut down on overhead.

LibationWinForms/GridEntry.cs
```diff
@@ -26,12 +29,15 @@ internal class GridEntry : AsyncNotifyPropertyChanged, IMemberComparable
 
 		private Book Book => LibraryBook.Book;
 		private Image _cover;
+		private Action _refilter;
```
rmcrackan, line 32: Personal preference: mark this as {get;} or readonly so it can't accidentally be overwritten elsewhere

LibationWinForms/BookLiberation/ProcessorAutomationController.cs
Mbucari: that DB call was previously inside IProcessibleExt.GetValidLibraryBooks. moving it into LibationWinForms allowed me to remove the ApplicationServices dependency from FileLiberator and make it DB ignorant.

LibationWinForms/GridEntry.cs
```diff
@@ -70,8 +76,48 @@ private void PictureStorage_PictureCached(object sender, FileManager.PictureCach
...
+
+		/// <summary>
+		/// This event handler receives notifications from the model that it has changed.
+		/// Save to the database and notify the view that it's changed.
+		/// </summary>
+		private void UserDefinedItem_ItemChanged(object sender, string itemName)
+		{
+			var udi = sender as UserDefinedItem;
+
+			if (udi.Book.AudibleProductId != LibraryBook.Book.AudibleProductId)
```
rmcrackan, line 89: These are fine as is, but in case it helps: LibraryBook.Book is aliased above as just Book since it's used so frequently.

LibationWinForms/GridEntry.cs
```diff
@@ -96,8 +142,22 @@ private set
 		public string Category { get; }
 		public string Misc { get; }
 		public string Description { get; }
-		public string DisplayTags => string.Join("\r\n", Book.UserDefinedItem.TagsEnumerated);
```
rmcrackan, line 99: This is an interesting one. I put each tag on its own line here with the assumption that each book will have no more than 3 or 4 tags. In this use case, it's much easier to read when they're on their own line. With your change I assume they will be space delimited instead of new line delimited. I prefer them on their own line but I'm open to conversation if there's a reason you think it should be otherwise.

LibationWinForms/GridEntry.cs
```diff
@@ -200,5 +260,12 @@ private static string GetMiscDisplay(LibraryBook libraryBook)
 		}
 
 		#endregion
+
+		~GridEntry()
+		{
```
rmcrackan, line 265: I've meant to bring it up before. I'm warry on destructors because GC can be such an unpredictable beast. It looks like you have more experience with them in practice though. How do you feel these compare to more explicit things like Dispose or OnClose?

ApplicationServices/LibraryCommands.cs
rmcrackan: That would help. Or a Transaction Scope. But it's still 3 separate db calls rather than 1. Obviously 1 vs 3 is trivial, but I'm concerned that it establishes a pattern that will only grow with future expansion, rather than being 1 update call.

LibationWinForms/BookLiberation/ProcessorAutomationController.cs
rmcrackan: Ah, good call.

LibationWinForms/GridEntry.cs
Mbucari: I wouldn't say I'm that experienced with them, but I think a finalizer is better used here than making GridEntry Disposable because:
1) There shouldn't be a problem with those event handlers if GridEntry lives longer than expected. They'll still fire, but they won't do anything since GridEntry has been unbound from the view.
2) The creator of a Disposable object is responsible for disposing of it, so ProductsGrid or Form1 would be responsible for iterating through the bindings list and manually disposing of every GridEntry each time the grid is reloaded.

LibationWinForms/GridEntry.cs
Mbucari: Gotcha. I'll change the getter back, but leave the setter so the editor can easily change them.

ApplicationServices/LibraryCommands.cs
Mbucari: OK, I added BeginEdit and EndEdit to GridEntry, and LibraryCommands only has a single method for updating the database for UserDefinedItem: UpdateUserDefinedItem.

rmcrackan:
Since this PR looked like pure architecture, I'm not going to push a new release yet. Lookin' good

PR #101, opened 2021-09-03
Fixed cross thread bug and Enumerable change bug.

Mbucari:
I don't know why these bugs just started showing up now, but they weren't happening when I made my last PR but they're happening every run now. 

Also fixed RemoveBooksDialog counter not updating.

### DIFF
--- a/LibationWinForms/Dialogs/RemoveBooksDialog.cs
+++ b/LibationWinForms/Dialogs/RemoveBooksDialog.cs
@@ -58,7 +58,7 @@ namespace LibationWinForms.Dialogs
 			{
 				var removedBooks = await LibraryCommands.FindInactiveBooks((account) => new WinformResponder(account), _libraryBooks, _accounts);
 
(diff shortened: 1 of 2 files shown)


Mbucari:
Be mindful that your UIThread() extension is asynchronous. It's fine for setting display info, or any fire-and-forget application, but sometimes you need to invoke Synchronously, and for that you should just use this.Invoke.

I mention this because there was only one cross-thread error inside setGrid, but when I used UIThread() I got other weird errors caused by the action completing after the rest of the method had executed. To work around that, I just invoked the call to setGrid.  But this is something that may come up again.

rmcrackan:
Good point. There are actually sync and async options: UIThread and InvokeIfRequired. They could really use being refactored, renamed, replaced throughout, and then these 2 marked as deprecated

rmcrackan:
Fixed. Old methods marked obsolete. Now: UIThreadSync and UIThreadAsync. I would like to have used UIThread and UIThreadAsync but UIThread was already async so that seemed like a bad idea

PR #103, opened 2021-09-04
Minor fix and changes for form size and location persistance.

Mbucari:
- FindInactiveBooks now calls the new logRestart method.
- Adjusted RemoveBooksDialog's default form size and widened the Remove column so you can see the sort arrow.
-  Added Form extension methods RestoreSizeAndLocation and SaveSizeAndLocation.
- RemoveBooksDialog, AudioConvertForm  and AudioDecryptForm now save their sizes and positions.

### DIFF
--- a/ApplicationServices/LibraryCommands.cs
+++ b/ApplicationServices/LibraryCommands.cs
@@ -18,6 +18,8 @@ namespace ApplicationServices
 
 		public static async Task<List<LibraryBook>> FindInactiveBooks(Func<Account, ILoginCallback> loginCallbackFactoryFunc, List<LibraryBook> existingLibrary, params Account[] accounts)
 		{
+			logRestart();
+
 			//These are the minimum response groups required for the
 			//library scanner to pass all validation and filtering.
 			LibraryResponseGroups =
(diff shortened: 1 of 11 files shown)


ApplicationServices/LibraryCommands.cs
```diff
@@ -18,6 +18,8 @@ public static class LibraryCommands
 
 		public static async Task<List<LibraryBook>> FindInactiveBooks(Func<Account, ILoginCallback> loginCallbackFactoryFunc, List<LibraryBook> existingLibrary, params Account[] accounts)
 		{
+			logRestart();
```
rmcrackan, line 21: glad to see you like my poor man's benchmarker that I forgot to delete :)

Mbucari: It was hard to miss when it was throwing exceptions!

rmcrackan: Oh shit. It didn’t do that to me. I guess I need to push through your fix to a release so that’s not what other users see

Mbucari: Yeah, it was crashing on the first logTime because __log was empty, so log..Last() threw an exception. It wouldn't happen if, before launching RemoveBooksDialog, you ran import books at least once.

Issue #104, opened 2021-09-07
MP3 Decode Cuts Audiobook Duration To Around 4 Hours


sgmax:
Using the MP3 encoding option cuts off the audiobook at around 4 hours (ish). I attach the log for my books - I noticed this yesterday when I tried to decode eight books to mp3 and ended up with durations around the 4 hour mark for all eight., 
I re-ran the downloads/decrypts in .M4B and the audiobook duration was set correctly (between 7 to 9+ hrs duration).
I also tried the download decode with and without the "Fix up metadata" option checked - the .M4B option works in both modes.
So it is definitely the MP3 decode that is stopping around halfway through. As the lengths are all in the ballpark of 4 hrs (plus or minus), I'm guessing a temporary counter/timer variable is to blame.
My default MP3 decoder is Lame_v3.100 (although it looks as if you use your own decoder, so this may be extraneous info).
I attach a verbose log for the MP3 decode (which I ran for just 2 of the decodes as an FYI).
I repeated these two decodes for an MP4 decode with the "Fix up metadata" option checked. 
Let me know if you need additional information.
Just wanted to say that I LOVE the software app - this is really helpful - many thanks!
[Log202109.log](https://github.com/rmcrackan/Libation/files/7123498/Log202109.log)
![Libation-Settings  MP3](https://user-images.githubusercontent.com/28791733/132390888-a24733b1-0cc9-4248-9f10-3771a528211c.jpg)
![Library Extract  showing lengths of books](https://user-images.githubusercontent.com/28791733/132390898-ea15fcd3-6da5-462f-9443-fe32ac48c8dd.JPG)
![Metadata-B002V8KQU8-mp3](https://user-images.githubusercontent.com/28791733/132390922-31eb2d8f-607a-448a-a421-734f16b81f38.JPG)
![Metadata-B002V8KQU8-m4b](https://user-images.githubusercontent.com/28791733/132390924-e316fc6d-7d04-4c06-befd-e869791ffe22.JPG)
![Metadata-B005KBQUKS-mp3](https://user-images.githubusercontent.com/28791733/132390925-ba6b5640-8e41-48a1-84ad-c1a7fa04ac90.JPG)
![Metadata-B005KBQUKS-m4b](https://user-images.githubusercontent.com/28791733/132390928-947716d4-809b-4c8e-b64d-156458bda061.JPG)

Mbucari:
[...]

PR #116, opened 2021-09-15
Correctly write mp3 Xing header.

Mbucari:
Solves #104


rmcrackan:
Awesome. Thanks again

PR #123, opened 2021-09-24
Add support for unencrypted mp3 audiobooks.

Mbucari:
Added new abstract class ```AudiobookDownloadBase``` that does most of what ```AaxcDownloadConverter``` used to do.

```AaxcDownloadConverter``` and ```Mp3AudiobookDownloader``` inherit ```AudiobookDownloadBase ```

Also, I moved logging into LiberationBaseForms's event handlers.  My idea is that the controller's events should only have one event handler, based in LiberationBaseForms. All work done on those events should be carried out by overriding those event handlers.  I won't be mad if you decide you want to change it back (:

### DIFF
--- a/AaxDecrypter/AaxcDownloadConverter.cs
+++ b/AaxDecrypter/AaxcDownloadConverter.cs
@@ -7,7 +7,7 @@ using System.IO;
 
 namespace AaxDecrypter
 {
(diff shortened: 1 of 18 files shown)


rmcrackan:
Thanks for the explanation. I was trying to wrap my brain around the code last night and couldn't tell what was going on at the time. That makes sense.

PR #127, opened 2021-09-26
To mulitple files

seanke:
I have just created a quick proof of concept. Interested to know if there's anyone who has looked into saving audiobooks into multiple files. If not, and you'd be interested in merging a feature that does so, I'll press on with it.

### DIFF
--- a/AaxDecrypter/AaxcDownloadConverter.cs
+++ b/AaxDecrypter/AaxcDownloadConverter.cs
@@ -102,8 +102,33 @@ namespace AaxDecrypter
             aaxFile.SetDecryptionKey(downloadLicense.AudibleKey, downloadLicense.AudibleIV);
             
             aaxFile.ConversionProgressUpdate += AaxFile_ConversionProgressUpdate;
+            if(OutputFormat == OutputFormat.M4b) 
+                ConvertToMultiMp4b();
+            else 
+                ConvertToMultiMp3();
+            aaxFile.ConversionProgressUpdate -= AaxFile_ConversionProgressUpdate;
+
+            aaxFile.Close();
+
+            return true;
+        }
+
+        private void ConvertToMultiMp4b()
+        {
             var chapterCount = 0;
-            //TODO make this work with m4b files
+            aaxFile.ConvertToMultiMp4a(downloadLicense.ChapterInfo, newSplitCallback =>
+            {
+                chapterCount++;
+                var fileName = Path.ChangeExtension(outputFileName, $"{chapterCount}.m4b");
+                if (File.Exists(fileName))
+                    FileExt.SafeDelete(fileName);
+                newSplitCallback.OutputFile = File.Open(fileName, FileMode.OpenOrCreate);
+            });
+        }
+        
+        private void ConvertToMultiMp3()
+        {
+            var chapterCount = 0;
             aaxFile.ConvertToMultiMp3(downloadLicense.ChapterInfo, newSplitCallback =>
             {
                 chapterCount++;
@@ -111,24 +136,8 @@ namespace AaxDecrypter
                 if (File.Exists(fileName))
                     FileExt.SafeDelete(fileName);
                 newSplitCallback.OutputFile = File.Open(fileName, FileMode.OpenOrCreate);
-                //TODO clean up all this junk
-                newSplitCallback.LameConfig = new NAudio.Lame.LameConfig
-                {
-                    ID3 = new NAudio.Lame.ID3TagData()
-                    {
-                        Track = chapterCount.ToString(),
-                        Artist = aaxFile.AppleTags.AlbumArtists,
-                        Album = aaxFile.AppleTags.Album,
-                        Title = aaxFile.AppleTags.Title,
-                        Genre = aaxFile.AppleTags.Generes
-                    }
-                };
+                newSplitCallback.LameConfig.ID3.Track = chapterCount.ToString();
             });
-            aaxFile.ConversionProgressUpdate -= AaxFile_ConversionProgressUpdate;
-
-            aaxFile.Close();
-
-            return true;
         }
 
         private void AaxFile_ConversionProgressUpdate(object sender, ConversionProgressEventArgs e)
--- a/AaxDecrypter/AaxcDownloadConverter.cs
+++ b/AaxDecrypter/AaxcDownloadConverter.cs
@@ -25,8 +25,12 @@ namespace AaxDecrypter
                 Name = "Download and Convert Aaxc To " + OutputFormat,
 
                 ["Step 1: Get Aaxc Metadata"] = Step1_GetMetadata,
-                ["Step 2: Download Decrypted Audiobook"] = splitFileByChapters ? Step2_DownloadAudiobookAsMultipleFilesPerChapter : Step2_DownloadAudiobookAsSingleFile,
-                ["Step 3: Create Cue"] = Step3_CreateCue,
+                ["Step 2: Download Decrypted Audiobook"] = splitFileByChapters
+                    ? Step2_DownloadAudiobookAsMultipleFilesPerChapter
+                    : Step2_DownloadAudiobookAsSingleFile,
+                ["Step 3: Create Cue"] = splitFileByChapters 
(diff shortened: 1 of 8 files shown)


rmcrackan:
Thanks! No efforts have been made on this so far. Feel free to continue. If this goes to completion I'll be glad to expand Setting to include a single-file vs multi-file setting. It will need to be a setting because I for sure want to keep all of my downloads as single file.

My efforts into encoding and decryption largely stagnated a few years ago when I got aax => m4b stable. Since then a volunteer / partner / reverse-engineering wizard has stepped in to provide the recent advances in decryption options. I'm including him here in case he wants to take a look.

@Mbucari  Let me know if you have any opinions on this. I value your input in this of all areas.

Mbucari:
AAXClean already supports splitting mp4's and mp3's by chapters. It's a really simple process, and I have a usage example in the readme:

https://github.com/Mbucari/AAXClean/blob/master/README.md#multipart-conversion-example

AaxDecrypter/AaxcDownloadConverter.cs
```diff
@@ -67,25 +67,40 @@ protected override bool Step2_DownloadAudiobook()
...
-            var outputFile =  File.Open(outputFileName, FileMode.OpenOrCreate, FileAccess.ReadWrite);
+            var chapterCount = 0;
+            //aaxFile.ConvertToMultiMp4a(downloadLicense.ChapterInfo, newSplitCallback =>
+            aaxFile.ConvertToMultiMp3(downloadLicense.ChapterInfo, newSplitCallback =>
+            {
+                chapterCount++;
+                var fileName = Path.ChangeExtension(outputFileName, $"{chapterCount}.mp3");
+                if (File.Exists(fileName))
+                    FileExt.SafeDelete(fileName);
+                newSplitCallback.OutputFile = File.Open(fileName, FileMode.OpenOrCreate);
```
Mbucari, line 81: You do need to give the output file write access. I updated the usage example on AAXClean.

AaxDecrypter/AaxcDownloadConverter.cs
```diff
@@ -67,25 +67,40 @@ protected override bool Step2_DownloadAudiobook()
...
+                newSplitCallback.LameConfig = new NAudio.Lame.LameConfig
+                {
```
Mbucari, line 83: When you override LameConfig, you need to specify encoding parameters. If you only want to adjust ID3 tags, don't reinitialize LameConfig. By default, MP3s are encoded to average bitrate of the input source and down sampled tomorrow.

Mbucari:
I think if you implement this future, you should do it in the existing AaxcDownloadConverter. Probably wouldn't be that much code. Maybe make a Step2_DownloadAudiobookSingle() and a Step2_DownloadAudiobookMultipart(), and assign the appropriate method to the step runner.

seanke:
Thanks for the input! I'll keep working on it. By the way, the codebase is really nice to work with.

rmcrackan:
@seanke , @Mbucari 

Annoyingly, I'm not getting email notification for a lot of these conversations. Maybe tag me in a response when you 2 think this is ready for actual integration.

AaxDecrypter/AaxcDownloadConverter.cs
Mbucari: A couple more things on NewSplitCallback. When you call AacToMp3MultipartFilter, it initializes LameConfig with default encoding parameters and ID3 tags.  The default ID3 tags are from here:
https://github.com/Mbucari/AAXClean/blob/e7cf26757a0f311be62010d5ba2116e3579e60d6/AudioFilters/AacToMp3Filter.cs#L40

The encoding settings and tags persist between callbacks, so the minimum you need to change is the output file stream and the track number.

AaxDecrypter/AudiobookDownloadBase.cs
```diff
@@ -59,7 +59,7 @@ public AudiobookDownloadBase(string outFileName, string cacheDirectory, Download
 
 		public abstract void Cancel();
 		protected abstract int GetSpeedup(TimeSpan elapsed);
-		protected abstract bool Step2_DownloadAudiobook();
+		protected abstract bool Step2_DownloadAudiobookAsSingleFile(); //TODO do I need to change this??
```
Mbucari, line 62: Yup.

AaxDecrypter/UnencryptedAudiobookDownloader.cs
```diff
@@ -21,7 +21,7 @@ public UnencryptedAudiobookDownloader(string outFileName, string cacheDirectory,
 				Name = "Download Mp3 Audiobook",
 
 				["Step 1: Get Mp3 Metadata"] = Step1_GetMetadata,
-				["Step 2: Download Audiobook"] = Step2_DownloadAudiobook,
+				["Step 2: Download Audiobook"] = Step2_DownloadAudiobookAsSingleFile,
 				["Step 3: Create Cue"] = Step3_CreateCue,
```
Mbucari, line 25: As I think about this, a CUE file doesn't make sense if you're splitting the audio at cue boundaries.

rmcrackan: Agreed. And this is looking more and more like we can ditch the 'steps' paradigm

```
step 1: metadata
step 2:
    if multi
        split
    else
        process as single
        create cue
```

Mbucari:
@rmcrackan looks like it's done. Have you reviewed?

Mbucari:
@rmcrackan looks like a good first version. I haven't pulled it in for testing, but I see no reason it won't work. The only thing I'm not sure about is DownloadDecryptBook.MoveFilesToBooksDir() no longer using AudibleFileStorage.IsFileTypeMatch().

FileManager/Configuration.cs
```diff
@@ -101,6 +101,13 @@ public bool DecryptToLossy
             get => persistentDictionary.GetNonString<bool>(nameof(DecryptToLossy));
             set => persistentDictionary.SetNonString(nameof(DecryptToLossy), value);
         }
+        
+        [Description("Decrypt to lossy format?")]
+        public bool SplitFilesByChapter
```
Mbucari, line 106: Can you please update the description?

FileLiberator/DownloadDecryptBook.cs
```diff
@@ -153,8 +155,8 @@ private static (string destinationDir, bool movedAudioFile) MoveFilesToBooksDir(
             foreach (var f in sortedFiles)
             {
                 var dest
-                    = AudibleFileStorage.Audio.IsFileTypeMatch(f)
```
rmcrackan, line 156: IsFileTypeMatch should probably still be used here

rmcrackan:
@Mbucari I agree, this looks like a good solution, especially as a first draft. Thanks for your eyes also

@seanke It looks like there are just a few minor things left. After they're addressed, ping me here and I'll do a final review before accepting. Thanks for adding this feature!

FileLiberator/DownloadDecryptBook.cs
seanke: Nice one! Have used that instead.

FileManager/Configuration.cs
seanke: Opps, nice catch. Fixed.

seanke:
@rmcrackan thanks! I have updated the code accordingly.

rmcrackan:
Thanks @seanke for your great work and @Mbucari for the extra attention with the decode stuff!

Mbucari:
*slaps forehead*
One thing I completely forgot about is that sometimes chapters have zero length (or close to it).  I can't think of any example at the moment, but I've seen it happen. It's usually when a book has Chapters grouped into 'Parts'. For example, "Part 1" is a chapter marker, but so it "Chapter 1", and they both point to the same timestamp.

seanke:
@rmcrackan @Mbucari what happens when you run the code over a book with one if these empty chapters?

rmcrackan:
Thank you for this explanation. It makes these cases much more clear for me. I'm copy/pasting this example into the code comments

PR #132, opened 2021-10-05
Convert IProcessable to abstract class Processable.

Mbucari:
As I described in my email, I converted IProcessable to abstract class Processable. IProcessableExt methods are now in Processable. 

Place your file cache update code inside Processable.OnStreamingCompleted().


Mbucari:
I also converted IStreamable and IAudioDecodable to abstract classes, moved event logging from LiberationBaseForm into those classes, and renamed LiberationBaseForm event handlers.

rmcrackan:
I like the new architecture, esp that we didn't sacrifice the good layer separation that we've got going on. Reviewing now.

I've tried hard to keep these as interfaces but I think you're right and practicality trumps purity. And if I really need the interfaces back later, I can extract it from the abstracts.

FileLiberator/Processable.cs
```diff
@@ -0,0 +1,75 @@
...
+            ? await ProcessAsync(libraryBook)
+            : new StatusHandler();
+
+        /// <returns>True == Valid</returns>
+        public abstract bool Validate(LibraryBook libraryBook);
+
+        /// <returns>True == success</returns>
+        public abstract Task<StatusHandler> ProcessAsync(LibraryBook libraryBook);
+
+        public void OnBegin(LibraryBook libraryBook)
```
rmcrackan, line 59: Same: public => protected

FileLiberator/Streamable.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace FileLiberator
+{
+    public abstract class Streamable
+    {
+        public event EventHandler<string> StreamingBegin;
+        public event EventHandler<DownloadProgress> StreamingProgressChanged;
+        public event EventHandler<TimeSpan> StreamingTimeRemaining;
+        public event EventHandler<string> StreamingCompleted;
+        public void OnStreamingBegin(string filePath)
```
rmcrackan, line 12: Same: public => protected

FileLiberator/AudioDecodable.cs
```diff
@@ -0,0 +1,40 @@
...
+    public abstract class AudioDecodable : Processable
+    {
+        public event EventHandler<Action<byte[]>> RequestCoverArt;
+        public event EventHandler<string> TitleDiscovered;
+        public event EventHandler<string> AuthorsDiscovered;
+        public event EventHandler<string> NarratorsDiscovered;
+        public event EventHandler<byte[]> CoverImageDiscovered;
+        public abstract void Cancel();
+
+        public void OnRequestCoverArt(Action<byte[]> setCoverArtDel)
```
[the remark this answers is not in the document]
Mbucari: They need to be public because LiberationBaseForm subscribes to them. You can't fire events in a base class from a derived class.

LibationWinForms/BookLiberation/BaseForms/LiberationBaseForm.cs
```diff
@@ -133,40 +133,30 @@ private void UnsubscribeAudioDecodable(object sender, EventArgs e)
 		/// (ie. shown) because Control doesn't get a window handle until it is Shown.
 		/// </summary>
 		private void OnStreamingBeginShow(object sender, string beginString) => Invoker.UIThreadAsync(Show);
-		
+
 		#endregion
 
-		#region IStreamable event handlers
```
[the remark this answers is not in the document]
Mbucari: Logging was moved into Processable, AudioDecodable, and Streamable classes.

FileLiberator/DownloadDecryptBook.cs
```diff
@@ -214,11 +201,11 @@ var title
                 throw new Exception(errorString("Locale"));
         }
 
-        public bool Validate(LibraryBook libraryBook) => !libraryBook.Book.Audio_Exists;
+        public override bool Validate(LibraryBook libraryBook) => !libraryBook.Book.Audio_Exists;
```
Mbucari, line 204: Is there a reason validate and Validate can't be combined?

rmcrackan: Different workflow logic. The private one should prob be renamed.

Public Validate is its Processable workflow: 'is this book a candidate for download/decrypt? if already liberated, then: no'.

Private validate: We know the book is a candidate for liberation. Do internal validation about the book's metadata.

If the public method also processed the private checks, one of these would be true for invalid metadata:

* return false. Indistinguishable from a book which has already been liberated. Bad metadata is hidden.
* throw. This implies there was an error seeing if this book is candidate for liberation. Possible, but it feels like it overloads what the public Validate is intended to check

PR #134, opened 2021-10-06
Fix splitting audiobooks on chapters

Mbucari:
Fixed issue described in #127 re chapters of zero length. A new split file will only be created if it is at least 15 seconds long.

Also made change to settings dialog so that SplitFilesByChapter can only be enabled if AllowLibationFixup is also enabled.


Mbucari:
I screwed this up. Don't push a release until I fix it.

PR #233, opened 2022-04-25
Update Libation to work with new AAXClean.Codecs

Mbucari:
Pretty minor stuff, but I did have to mark almost every project as windows-only because AAXClean.Codecs is windows-only.  Technically, AAXClean was windows-only before, but it just wasn't designated as such.  AAXClean is now platform independent.


rmcrackan:
Muchas gracias

Issue #207, opened 2022-03-04
Enhancement: ability to remove "This is Audible" audio bookends and other ideas


DepressMode:
It would be nice to have the capability to remove the poor-quality "This is Audible." and "Audible hopes you have enjoyed this program." messages at the very beginning/end of each Audible book. They just don't sound good and our books will never be truly liberated until we've wiped the name of its keeper. I know I've seen this capability in the inAudible v1.97 software (I can attach the files or a magnet link if desired). Libation is easily superior in almost every way, but the feature I miss most is the ability to remove those messages automatically. It should be relatively simple considering the messages are always the same for each audiobook. 

A couple other suggestions: 
The option for simple audio mixing (mono vs stereo when available, etc.)
Individual converting of .M4B to .MP3 rather than entire library at once (perhaps in the tag edit menu)
.NFO files for each file version
Automatically download .JPG of thumbnail art.
Integrate chapters into .MP3 conversions

rmcrackan:
> Remove beginning/ending audible sound bites ... Simple audio mixing

Direct audio manipulation is a can of worms I don't want to open. I'm trying to stick to the approach of everyone do what they do best. My focus is freeing audible libraries. Other people already specialize at audio manipulation.

> .NFO files for each file version

Used to have these but got rid of them. Metadata is available via exports

> Automatically download .JPG of thumbnail art.

Good idea. Tracking this one in #182 

> Individual converting of .M4B to .MP3 rather than entire library at once ... Integrate chapters into .MP3 conversions

The m4b => mp3 is a convenience feature; not a core offering. I like the idea of allowing it to target files. I currently don't intend to extend its functionality. I recommend finding a file splitter program to use with the generated cue file. Or re-import with mp3 and chapter-izing settings enabled.

rmcrackan:
@DepressMode A lot of new stuff added in [v7.2](https://github.com/rmcrackan/Libation/releases/tag/v7.2.0). I think you'll be pleased

PR #238, opened 2022-05-08
Add option to remove Audible branding audio

Mbucari:
Addresses #207 with a new setting and some minor changes to the downloadAudiobookAsync routine.

Nothing too fancy here.  AAXClean just discards AAC audio frames before the first chapter and after the last chapter.  The whole file is still downloaded and read.

### DIFF
--- a/AaxDecrypter/AaxDecrypter.csproj
+++ b/AaxDecrypter/AaxDecrypter.csproj
@@ -5,8 +5,8 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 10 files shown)


rmcrackan:
This looks great. Thank you so much.

One minor thing to fix before I approve: please add the setting new to LibationScaffolding.PopulateMissingConfigValues with default: false

Mbucari:
You're very welcome.  I added the default setting and also did a little cleanup on DownloadDecryptBook.

Issue #223, opened 2022-04-06
Remove "(Unabridged)" from the title?


macr0dev:
One of these days when you feel like added more enhancements (I know you said you were busy and only doing minor things and fixes) it would be great to have the option to not have the "**(Unabridged)**" part of the title.  I know that it's in the aax files, it just would be great to be able to remove it.  It's not part of the title on the audible site, and I really don't understand why they shoe-horn it into the aax file.

Reasons?  When I use the albums with booksonic, the embedded search function won't work because it includes that work.  When I try to use it with the new app AudioBookShelf, it won't auto match for the same reason.  Although they should really be using a fuzzy search....

Thanks!

Issue #187, opened 2022-01-19
Feature Request: Allow saving the original aax


TomboAhi:
I would like an option to  download and save the original encrypted .aax file, as is, without any decryption or conversion.

rmcrackan:
Interesting. Can I ask what advantage this has over the decrypted .m4b file? The decrypted m4b file is the same audio quality, only without the drm.

TomboAhi:
Downloading a book is significantly faster than decrypting it. So having the option to not decrypt will allow for the download of a large library as well as keeping the library in sync with your local copy would be significantly faster because the added process of decryption is not taking place. 

It would also affords the flexibility to decrypt the books independently, as needed in the format needed at the time. It will always be there as the aax and if I want book A as an M4b for an audiobook player but then want another book as MP3 for use in the car, I can only do what I need when I need.

Kazaa3:
I would like to add that depending on the country laws of the user, the legal classification of the backup could be different. A simple backup is legal almost everywhere in the world. The .m4b conversion could possibly mean circumventing an effective copy protection measure, so the situation is less clear (although of course it's sad that we live in a world where an honest buyer may have to worry about such things). But saving the original Audible files could give a user this security. 

Btw: An extension of this feature could be a post batch conversion of the original files as well as a NAS storage path for the archive that is syncronized around by newly added files.

wtanksleyjr:
I remember asking for the app to preserve AAX files. This was back when the
app used to download them --- it doesn't anymore.

Aside from that: I'm not sure where you heard that. Downloading a book is
vastly slower than decrypting it. Decrypting is extremely simple and
straightforward, a matter of memory access only, while downloading requires
network access. I think you might be thinking of the old rainbow-table hash
attack that older tools like OpenAudible use; this application doesn't use
any attack, it simply uses Audible's API to get the files and the proper
decryption key.

You might be right that the legal classification could be different, I'm
not a lawyer and even lawyers have to be specialists to do that kind of
thing. I know it's no different in the US code -- so long as the software
CAN decrypt the files it's a circumvention device, even if it doesn't
actually DO the decryption. It's notable, therefore, for the sake of laws
that concern themselves with this, that downloading the AAX files is
officially supported by their own public website for the purpose of loading
them into iTunes. I would suggest, though, that what you want if you're
trying to be safe is probably to stick with what's on their public website
and not to use an app like this.

-Wm

TomboAhi:
[...]

Issue #144, opened 2021-10-28
Partial / Complete files in `%temp%\Libation` are not tidied / reused


attie:
While figuring out #143, I had a book download and decrypt, but then fail to move into its final position.

This file was left as `%temp%\Libation\DecryptInProgress\x.x.m4b`.

Upon re-running the download/decrypt, these files were ignored (not reused or removed), the book was re-downloaded, re-decrypted and the new files had a ` (1)` appended (e.g: `x.x (1).m4b`).

I don't imagine these files should be left behind during normal operation (but if an error or perhaps if a force-quit occurs?)...
Therefore, it might be nice to try handling these files - e.g: either verify for completeness (difficult) / delete on retry (easy) / delete on failure (not ideal) / etc...

rmcrackan:
I'm experimenting with tracking enhancements and feature requests using github issues instead of a local file.

# Clean up old bad files

* On error: delete. Clean up after your own process when possible.
* Delete old bad files on start. This will catch places where the previous isn't possible.

rmcrackan:
Added in [v7.2](https://github.com/rmcrackan/Libation/releases/tag/v7.2.0)

PR #239, opened 2022-05-08
Fix temp file storage/deletion + a few new features

Mbucari:
- .tmp and .json files are again properly stored in DownloadsInProgressDirectory, and the are again used if resuming a download.
- Yellow stoplight works again.
- All audio files stored in DecryptInProgressDirectory are deleted if the conversion fails #144
- Added option to retain aax file #187
- Add option to remove "(Unabridged)" from metadata tags #223
- Added options for mp3 encoding
- Add option to create cue sheet
- Prevent same book being decrypted more than once at a time
- Made a new settings tab for all audio file editing/fixing options.

image

### DIFF
--- a/LibationWinForms/grid/GridEntry.cs
+++ b/LibationWinForms/grid/GridEntry.cs
@@ -10,6 +10,7 @@ using Dinah.Core.DataBinding;
 using Dinah.Core;
 using Dinah.Core.Drawing;
 using LibationFileManager;
+using System.Threading.Tasks;
 
 namespace LibationWinForms
 {
@@ -39,6 +40,7 @@ namespace LibationWinForms
 			}
 		}
 
+		public bool DownloadInProgress { get; private set; }
 		public string ProductRating { get; private set; }
 		public string PurchaseDate { get; private set; }
 		public string MyRating { get; private set; }
@@ -76,6 +78,16 @@ namespace LibationWinForms
 		private Book Book => LibraryBook.Book;
 
 		public GridEntry(LibraryBook libraryBook) => setLibraryBook(libraryBook);
+
+		public async Task DownloadBook()
+		{
+			if (!DownloadInProgress)
+			{
+				DownloadInProgress = true;
+				await BookLiberation.ProcessorAutomationController.BackupSingleBookAsync(LibraryBook);
+				DownloadInProgress = false;
+			}
+		}
 
 		public void UpdateLibraryBook(LibraryBook libraryBook)
 		{
--- a/LibationWinForms/grid/GridEntry.cs
+++ b/LibationWinForms/grid/GridEntry.cs
@@ -83,9 +83,15 @@ namespace LibationWinForms
 		{
 			if (!DownloadInProgress)
 			{
-				DownloadInProgress = true;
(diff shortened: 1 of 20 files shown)


LibationWinForms/grid/GridEntry.cs
```diff
@@ -77,6 +79,16 @@ public string DisplayTags
 
 		public GridEntry(LibraryBook libraryBook) => setLibraryBook(libraryBook);
 
+		public async Task DownloadBook()
+		{
+			if (!DownloadInProgress)
+			{
+				DownloadInProgress = true;
```
rmcrackan, line 86: I recommend wrapping this like, or similar:

    if (!DownloadInProgress)
    {
        try
        {
            DownloadInProgress = true;
            await BookLiberation.ProcessorAutomationController.BackupSingleBookAsync(LibraryBook);
        }
        finally
        {
            DownloadInProgress = false;
        }
    }

rmcrackan:
[Dude](https://i.kym-cdn.com/photos/images/original/000/971/686/891.jpg). This is seriously impressive for 1 day's work. The hell do you need me for? I had 1 very minor comment and other than that it's ready to rock.

Mbucari:
Ha! Thank you! When I get on a roll it's hard to stop!

I made the change you recommended, and I also added code to save the decryption key to a file if the user is saving the aax file. Cheers!

Mbucari:
I think you're referring to the DownloadLicense constructor.  In that case, DownloadLicense was used for podcast-type audio files that were delivered as unencrypted mp3s, so null keys didn't mattwer.  Here, we only want to write the aaxc file and key file if AudiobookDownloadBase is actually downloading an aaxc file.

rmcrackan:
> save the decryption key

Good call. I had another comment but it's just curiosity. I doubt it's blocking so I'll put this through.

I'm on the cusp of enabling library auto-update and I've got another request that should be low hanging fruit too. Once work slacks off just a tiny bit I expect to release both of those. The auto-update turned out to be surprisingly hard due to choices I made years ago about how the grid is updated -- oops. I refactored that recently, as well as significantly speeding up database import. Between your work and mine, it's going to be a *very* fine month for features.

Mbucari:
I noticed the speed increase! It's VERY welcome. thanks!

rmcrackan:
tl;dr: skip this entirely if you don't care about implementation details about the speed increase

This is why you benchmark. The speed increase came from the most unlikely place imaginable. In the import, I pull all relevant items into an in-memory cache that Entity Framework has then I use that. WAY faster than hitting the database: in memory (rather than hitting the db each time) and O(sqrt(n)) -- stupid fast. Except it wasn't. That step was *destroying* my benchmarks -- wtf! Turns out, those benefits only applied if you queried by a certain field which I wasn't. I already indexed by my needed fields, so hitting the db itself was twice as fast as that stupid in-mem table. Then I added a switched to using my own in-memory hashtable to mirror the database's index tables. Long story short: what was taking 20,000-40,000 ms -- after my own hashtables -- was down to ONE ms

Mbucari:
That's crazy.  And +1 to benchmarking. TVS's built-in profiler has helped me track down several nasty performance bottlenecks!

Issue #182, opened 2022-01-10
Save cover image along side audio book


BigGuyWithBeard:
Hi, 
Can we have an option to save a jpg next ot the audio book, eg folder.jpg.  then in windows explorer, we can see thumbnail of the book cover.

rmcrackan:
@Mbucari I swear I saw this when I was reviewing the PR but now I don't see this new setting. Am I going nuts?

Mbucari:
I didn't add a setting for it. Libation saves a full size image for every book it downloads.  Probably should make it optional, huh.

Mbucari:
The picture download method is in FileLiberator.Processable, but it probably should be in FileLiberator.DownloadDecryptBook because that's where all other config options are read and executed.

rmcrackan:
Yeah, it would be better as optional. And I assume it should be named the same as the file. Or the book title in the case of chapterized. RE location: you're probably right there also. You've been in that part of the code a lot more recently than me. If you have another change in you, go for it. Otherwise, it might just have to wait until ... who knows.

Mbucari:
Yea, I'll do that here soon.  I also had something else in the pipe that's nearly ready.

rmcrackan:
@BigGuyWithBeard this is technically in the current version now (non-optional) and in the next version it will be available as an optional setting. So I'm going to close out this issue

PR #240, opened 2022-05-10
Lots of new UI stuff!

Mbucari:
- Added Libation cup icon to most forms
- Faster grid sorting by Liberated status
- Products Grid is now fully customizable (see Figure 1) Users can:
  - Rearrange columns
  - Resize most columns
  - Hide columns
- When a user clicks on the description grid cell, a textbox pops up showing the full description (see Figure 2).
- Download 1215x1215 cover art after successfully backing up the book. (See AudibleApi PR) #182

## Figure 1
image

## Figure 2
image


Source/AaxDecrypter/AudiobookDownloadBase.cs
```diff
@@ -140,11 +140,6 @@ protected bool Step_Cleanup()
 				else
 					FileUtility.SaferDelete(TempFilePath);
 			}
-			else
-			{
-				FileUtility.SaferDelete(OutputFileName);
```
Mbucari, line 145: All temp file deletions on failure are handled by DownloadDecryptBook.ProcessAsync

Source/LibationWinForms/Dialogs/SettingsDialog.AudioSettings.cs
```diff
@@ -17,7 +17,6 @@ private void lameTargetRb_CheckedChanged(object sender, EventArgs e)
 		private void LameMatchSourceBRCbox_CheckedChanged(object sender, EventArgs e)
 		{
 			lameBitrateTb.Enabled = !LameMatchSourceBRCbox.Checked;
-			lameConstantBitrateCbox.Enabled = !LameMatchSourceBRCbox.Checked;
```
Mbucari, line 20: These options are not mutually exclusive

Source/LibationWinForms/Form1.cs
```diff
@@ -262,8 +262,8 @@ private void configAndInitAutoScan()
 					.Where(a => a.LibraryScan)
 					.ToArray();
 
-					// in autoScan, new books SHALL NOT show dialog
-					await LibraryCommands.ImportAccountAsync(Login.WinformLoginChoiceEager.ApiExtendedFunc, accounts);
+				// in autoScan, new books SHALL NOT show dialog
+				await Invoke(async () => await LibraryCommands.ImportAccountAsync(Login.WinformLoginChoiceEager.ApiExtendedFunc, accounts));
```
Mbucari, line 266: This was throwing a cross-thread error after it's first 5 minutes elapsed.

Source/LibationWinForms/grid/ProductsGrid.cs
```diff
@@ -225,5 +244,102 @@ public void Filter(string searchString)
 		#region DataGridView Macro
 		private GridEntry getGridEntry(int rowIndex) => _dataGridView.GetBoundItem<GridEntry>(rowIndex);
 		#endregion
+
+		#region Column Customizations
+
+		protected override void OnVisibleChanged(EventArgs e)
+		{
```
Mbucari, line 251: Do all column customization in OnVisibleChanged. Weird things can happen in some edge cases if these tasks are done inside the constructor or OnLoad, and since this is a Control not a Form, we don't have access to a Shown event.

Source/LibationWinForms/grid/GridEntry.cs
```diff
@@ -61,10 +65,12 @@ public string DisplayTags
 		// these 2 values being in 1 field is the trick behind getting the liberated+pdf 'stoplight' icon to draw. See: LiberateDataGridViewImageButtonCell.Paint
 		public (LiberatedStatus BookStatus, LiberatedStatus? PdfStatus) Liberate
 		{
-			get => (LibraryCommands.Liberated_Status(LibraryBook.Book), LibraryCommands.Pdf_Status(LibraryBook.Book));
+			get => (_bookStatus, _pdfStatus);
 			
 			set
 			{
+				_bookStatus = value.BookStatus;
+				_pdfStatus = value.PdfStatus;
```
Mbucari, line 73: Cache book and pdf statuses for faster sorting by liberated status

rmcrackan:
Don't check in LibationContext.db. It looks like it was in there already which means one of us did it accidentally before

Source/LibationFileManager/PictureStorage.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace LibationFileManager
 {
-	public enum PictureSize { _80x80 = 80, _300x300 = 300, _500x500 = 500 }
+	public enum PictureSize { _80x80 = 80, _300x300 = 300, _500x500 = 500 , _1215x1215 = 1215 }
```
rmcrackan, line 11: Per other conversation: as this code is here, this will apply the 1215 size to old/smaller/reduced-size image ID

Source/LibationWinForms/Form1.cs
rmcrackan: Damn. It didn't do that for me. Thanks for the fix

Source/LibationFileManager/PictureStorage.cs
Mbucari: It only matters if you're requesting an image with that size.  I removed it for now because the cover should be the Large native size, not reduced to 1215.

Mbucari:
I'm done here.

Mbucari:
Updated AudibleApi package reference.

PR #242, opened 2022-05-10
Add option to download cover art & full-size cover art viewer

Mbucari:


Source/FileManager/BackgroundFileSystem.cs
```diff
@@ -134,6 +134,7 @@ private void RemovePath(string path)
 
         private void AddPath(string path)
         {
+            if (!File.Exists(path)) return;
```
Mbucari, line 137: SaveFileDialog does something weird.  It triggers the scanner to think there's a file, but there isn't one.

Source/LibationWinForms/grid/ImageDisplay.cs
```diff
@@ -0,0 +1,118 @@
...
+using LibationFileManager;
+using System;
+using System.Drawing;
+using System.IO;
+using System.Windows.Forms;
+
+namespace LibationWinForms
+{
+	public partial class ImageDisplay : Form
+	{
```
Mbucari, line 11: All the weird overrides in here are to make the form's aspect ratio always match the picture's aspect ratio.

Source/LibationWinForms/Dialogs/SettingsDialog.Designer.cs
```diff
@@ -28,957 +28,972 @@ protected override void Dispose(bool disposing)
 		/// </summary>
 		private void InitializeComponent()
 		{
-            this.booksLocationDescLbl = new System.Windows.Forms.Label();
```
Mbucari, line 31: Tabs vs spaces...

Source/FileLiberator/DownloadDecryptBook.cs
```diff
@@ -278,5 +278,34 @@ private static bool moveFilesToBooksDir(LibraryBook libraryBook, List<FilePathCa
...
+                var picBytes = PictureStorage.GetPictureSynchronously(new PictureDefinition(picId, size));
+
+                if (picBytes.Length > 0)
+                    File.WriteAllBytes(coverPath, picBytes);
+            }
+            catch (Exception ex)
+            {
+                //Failure to download cover art should not be
+                //considered a failure to download the book
+                Serilog.Log.Logger.Error(ex.Message);
```
rmcrackan, line 306: I generally recommend using an Error() overload that takes the exception as a param. It'll yield more helpful logs

Source/FileManager/BackgroundFileSystem.cs
rmcrackan: Good catch. That is indeed weird

Source/LibationWinForms/grid/ImageDisplay.cs
rmcrackan: Ah. Can you add that comment in-code somewhere here? Otherwise I'm bound to forget

Source/LibationWinForms/grid/ProductsGrid.cs
```diff
@@ -62,14 +65,37 @@ private async void DataGridView_CellContentClick(object sender, DataGridViewCell
...
+			else if (e.ColumnIndex == descriptionGVColumn.Index)
 				Description_Click(getGridEntry(e.RowIndex), _dataGridView.GetCellDisplayRectangle(e.ColumnIndex, e.RowIndex, false));
+			else if (e.ColumnIndex == coverGVColumn.Index)
+				await Cover_Click(getGridEntry(e.RowIndex));
+		}
+
+		private async Task Cover_Click(GridEntry liveGridEntry)
+		{
+			var picDefinition = new PictureDefinition(liveGridEntry.LibraryBook.Book.PictureLarge, PictureSize.Native);
+			var picDlTask = Task.Run(() => PictureStorage.GetPictureSynchronously(picDefinition));
```
rmcrackan, line 81: Does this need an await ?

rmcrackan:
Unrelated to the changes in this PR: I've nixed db persistence which is a side effect of other changes. eg: in GridEntry, UserDefinedItem_ItemChanged no longer calls Commit. When I was implementing batch features, performance was awful because the client couldn't define its own transactions; there were too many Commit-like calls resulting incidentally. In-memory things can still propagate with events/notifications, but the code initiating the change should define when persistence happens. To that end, I had to be diligent with enforcing this pattern, for example in DownloadDecryptBook.ProcessAsync

    libraryBook.Book.UserDefinedItem.BookStatus = LiberatedStatus.Liberated;
    ApplicationServices.LibraryCommands.UpdateUserDefinedItem(libraryBook.Book);

Mbucari:
> Unrelated to the changes in this PR: I've nixed db persistence which is a side effect of other changes. 

Good to know.

Source/LibationWinForms/grid/ProductsGrid.cs
rmcrackan: Does this need `await Task.Run`... ?

(Forgive me if this is repetitive, I'm fairly unfamiliar with how github does things for collaboration. I'm used to working solo on github or in a team not on github)

Issue #169, opened 2021-12-04
Request - Replace individual status windows with a queue docked in the main window


TipsyOne:
When manually starting a download a new window opens to display its status. If another is started its window gets stacked on top of the previous ones. This can cause a mess of windows if several are running at once.  

I would suggest creating a docked queue window that allows prioritizing active/upcoming downloads, and a toggle between simultaneous and sequential downloads.

rmcrackan:
Addressed in new version [v7.6](https://github.com/rmcrackan/Libation/releases/latest)

PR #243, opened 2022-05-12
Minor mods for future UI changes

Mbucari:
I'm working on #169.

I copied Form1 to Form1_New and am making my modifications there.  My plan is that when I'm ready for you to look at it, you can just accept my PR and do testing in your IDE by swapping in Form1_New for Form1 inside of Main() instead of trying to figure it out by just reading the code. At that point, everything should be new code and your review process before accepting the PR should be quicker.

That said, these are some minor changes to the existing architecture that my new US mods need to work.  Since they're not breaking, I thought you could just merge them now.


Mbucari:
Here's what it's looking like:

image
image

rmcrackan:
Wow, that's pretty cool. Logistically, how will I continue to be able to do work that touches Form1 while you're doing this though? Do we first need a refactor of Form1 to segregate things in different files to reduce conflicts?

Mbucari:
My thinking is that you can continue making changes to Form1 in the meantime.  When I'm ready to push this feature, I'll stitch it to the latest version of Form1 in your repo.

The whole thing is pretty loosely coupled to Form1.  Right now it's just:
 - That button to the right of "Filter" and its click handler.
 - Once more menu item in the Liberate menu and its click handler
 - An event handler for ProductGrid.LiberateClicked (added in this PR)

But if you could accept this PR, it would make this final feature push much smoother because it will just be new stuff and no conflicts.

Mbucari:
Never mind! After going through some merge hell I realized you were most certainly right! I pushed a change with Form1 set up to work with the docking queue.  A lot of the new stuff is in here, but none of it is wired up so you can release it without anyone knowing it exists and without it causing problems.

If you want to see the docking container pop in and out (which is pretty nifty if I do say so), just uncomment line 29 in the Form1 constructor.

rmcrackan:
Sure, I just released so I'll stop now. My next big thing is that refactor. Once I start that, merging will be basically impossible for the main 3 front-end classes. Keep your new code as isolated as possible. I mean, you're always good about that but especially now before that refactor.

Source/LibationWinForms/grid/ProductsGrid.cs
```diff
@@ -271,6 +273,9 @@ internal List<DataLayer.LibraryBook> GetVisible()
 			.Select(row => ((GridEntry)row.DataBoundItem).LibraryBook)
 			.ToList();
 
+		internal IReadOnlyList<GridEntry> AllEntries
+		=> bindingList;
```
rmcrackan, line 277: Ditto: Also looks unused?

Source/LibationWinForms/LibationWinForms.csproj
```diff
@@ -38,6 +38,9 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Compile Update="Form1_New.cs">
```
rmcrackan: I don't see a Form1_New.cs

Source/LibationWinForms/grid/ProductsGrid.cs
```diff
@@ -37,6 +38,7 @@ namespace LibationWinForms
 
 	public partial class ProductsGrid : UserControl
 	{
+		internal event EventHandler<GridEntry> LiberateClicked;
```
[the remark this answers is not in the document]
Mbucari: Yeah, both will be used eventually (or something like them), but I can remove them for now.

My vision for the overall architecture is to go full MVVM, with GridEntry being the viewmodel.  UI elements should only interact with the database through GridEntry.  That may not be possible and I understand if you have different priorities, but doing it that way would allow for much greater separation of concerns and looser coupling.

Source/LibationWinForms/LibationWinForms.csproj
Mbucari: Yeah, that was a holdover that I removed.

rmcrackan:
Can you change the folder and namespace too? I think that's probably the last item.

Mbucari:
All done

rmcrackan:
This is interesting. How will this work when default liberating the entire library? Currently there's no queue as such. Or arguably: the entire library is the queue every time. With thousands of books each time, I could imagine that crippling the app. I haven't delved into it yet. Does it just show the top 'n' and then "...and 1,234 more items" ?

rmcrackan:
The tray starts open and is stuck open for me. The `<<<` button in your screenshot is not present for me

Mbucari:
Use the  "Pop Out" button on the bottom right of the window.

For it to start closed (whish it should be until this feature is ready to roll out), uncomment this line in Form1 constructor.

			//splitContainer1.Panel2Collapsed = true;

Mbucari:
I removed the hide function.  I can put it back easily enough later.

PR #244, opened 2022-05-14
New Processing Queue

Mbucari:
Here it is!

I think the functionality is very intuitive, so not a whole lot of instruction is needed. Key points are:

- The whole library is loaded at once, and only controls that are in view are rendered.  I made a new VirtualFlowControl to handle this. After our chitchat re INotifyPropertyChanged, i realized that I had implemented something nearly identical without realizing, so I just implemented INotifyPropertyChanged instead. As an analouge to help you understand, ProcessQueueControl is like the DataGridView, ProcessBook is like GridEntry, and ProcessBookControl is like a row in DataGridView.
- ~~All books must be added on the UI thread.  They aren't rendered as they're added so CPU shouldn't be hit too hard, but the internal objects must be created on the UI thread.  To help with this, methods that accept IEnumerable<LibraryBook> automatically synchronize with the UI thread.~~ not true anymore. They only need to be queued on the UI thread. Also, I'm not liking the TrackedQueue. I'm going to try to get away with just a single list. I think that combined with the stored statuses will be enough and greatly simplify things.
- I think the best way to evaluate this feature is to just try it and ask questions.

image

image


Source/LibationWinForms/Form1.Liberate.cs
```diff
@@ -7,11 +9,14 @@ public partial class Form1
 	{
 		private void Configure_Liberate() { }
 
-		private async void beginBookBackupsToolStripMenuItem_Click(object sender, EventArgs e)
-			=> await BookLiberation.ProcessorAutomationController.BackupAllBooksAsync();
+		//GetLibrary_Flat_NoTracking() may take a long time on a hugh library. so run in new thread 
+		private void beginBookBackupsToolStripMenuItem_Click(object sender, EventArgs e)
+			=> Task.Run(() => processBookQueue1.AddDownloadDecrypt(ApplicationServices.DbContexts.GetLibrary_Flat_NoTracking()
```
rmcrackan, line 14: [Eliding](https://blog.stephencleary.com/2016/12/eliding-async-await.html) should be used carefully and I'm wary of using it with UI event delegates. If you've tested these this runs correctly as async, we can leave it though.

Source/LibationWinForms/grid/ProductsGrid.cs
```diff
@@ -144,8 +146,7 @@ private static async Task Liberate_Click(GridEntry liveGridEntry)
 				return;
 			}
 
-			// else: liberate
-			await liveGridEntry.DownloadBook();
+			LiberateClicked?.Invoke(this, liveGridEntry.LibraryBook);
```
rmcrackan, line 149: Curious why you opted for .Invoke instead of await

rmcrackan:
1. 

> Also, I'm not liking the TrackedQueue. I'm going to try to get away with just a single list.

Is this still true? That is: are related changes pending?

2. 

Re INotifyPropertyChanged/MVVM, I've done some research this weekend so I can understand it better. Braindump notes coming soon but [1] I haven't yet decided where to add these comments and [2] I don't want to risk complicating your PR for something trivial like this.

3. 

In a previous PR, [this](https://github.com/rmcrackan/Libation/blob/master/Source/LibationWinForms/grid/ProductsGrid.cs#L92-L109) created the `picDlTask` task at ln 92 but doesn't `await` it until ln 109. It's not wrong but I'm curious about the thinking behind this unconventional choice. ie: I feel like I'm missing something

Mbucari:
1. Not true anymore. I'm bacvk to thinking it's the right choice for the job. A mysterious OutORange exception I was experiencing was caused by bad update logic in the UI, not the structure itself.
2. Eagerly awaiting (heh)
3. Because Task.Run starts the task. It's running on its own thread. It gets to do work before you need it.  It's not that unconventional.  Microsoft describes the practice in their intro to async/await.  It allows out of order execution, which is what I was going for. If you await every async task, you're still doing all tasks in order and the only benefit you have is not locking down the UI thread.

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/

> In many scenarios, you want to start several independent tasks immediately. Then, as each task finishes, you can continue other work that's ready. In the breakfast analogy, that's how you get breakfast done more quickly. You also get everything done close to the same time. You'll get a hot breakfast.

rmcrackan:
Sounds reasonable.

Last question for this PR. Again: nothing wrong with this, I'm just curious: why this change?

ProductsGrid. old:

> await liveGridEntry.DownloadBook();

new:

LiberateClicked?.Invoke(this, liveGridEntry.LibraryBook);

Mbucari:
I have it set up so that clicking on a stoplight enqueues that book in the ProcessQueue. I wish I could link to a line of code here, but since I cant, look at Source/LibationWinForms/Form1.ProcessQueue.cs#15

Configure_ProcessQueue() subscribes to LiberateClicked, and handles the event by adding that LibraryBook to the ProcessQueueControl.

rmcrackan:
n/m I'm a friggin idiot. Somehow I thought you changed an await to invoking that same method differently without the `async` keyword. The code is entirely different. I think I drank too much last night...

Source/LibationWinForms/ProcessQueue/VirtualFlowControl.cs
```diff
@@ -190,7 +190,10 @@ private void AdjustScrollBar()
 				vScrollBar1.LargeChange = LargeScrollChange;
 
 				//https://stackoverflow.com/a/2882878/3335599
-				vScrollBar1.Maximum = VirtualHeight + vScrollBar1.LargeChange - 1;
```
rmcrackan: Seems like there's still a lot of code churn. Do you want me to accept the PR now? or give you more time to tweak?

PR #245, opened 2022-05-15
Fixed scaling issue.

Mbucari:
Don't know how I broke it again. I must have accidentally reverted a change at some point.  I tested this fix with 100%, 125% and 150% scaling and it works


Mbucari:
Fuck. No, stuff is off screen at 100% WTH?

PR #249, opened 2022-05-16
Add FilterableSortableBindingList to handle filtering the DataGridView

Mbucari:
```FilterableSortableBindingList``` extends ```SortableBindingList<GridEntry>``` and implements ```IBindingListView```

It handles filtering through the ```IBindingListView.Filter``` property using ```SearchEngineCommands.Search```

All affected methods in ```ProductsGrid``` have been updated to use this interface.

Doing filtering this way solves the problem you described in your email, adheres to Microsoft's design principles, and makes ProductsGrid less messy.

### CODE REVIEW

**Comment by Mbucari** [comment]
Alight. I've pushed the changes you recommended in your email.

File: Source/LibationWinForms/Form1.Liberate.cs [hunk 1]
```diff
@@ -11,12 +11,18 @@ public partial class Form1
 
 		//GetLibrary_Flat_NoTracking() may take a long time on a hugh library. so run in new thread 
 		private async void beginBookBackupsToolStripMenuItem_Click(object sender, EventArgs e)
-			=> await Task.Run(() => processBookQueue1.AddDownloadDecrypt(ApplicationServices.DbContexts.GetLibrary_Flat_NoTracking()
+		{
+			SetQueueCollapseState(false);
```
rmcrackan, line 15: Great minds think alike. I was going to suggest this too but I forgot to mention it when I was writing that email

Source/LibationWinForms/grid/GridEntry.cs
```diff
@@ -160,10 +160,12 @@ private void UserDefinedItem_ItemChanged(object sender, string itemName)
 					break;
 				case nameof(udi.BookStatus):
 					Book.UserDefinedItem.BookStatus = udi.BookStatus;
+					_bookStatus = udi.BookStatus;
```
rmcrackan, line 163: Huh... I could have sworn this was already in there. I hope I didn't take it out by mistake

Mbucari: Who knows. This section has been touched so much over the last week lol.

Source/LibationWinForms/grid/FilterableSortableBindingList.cs
```diff
@@ -0,0 +1,99 @@
...
+		/// Items that were removed from the base list due to filtering
+		/// </summary>
+		private readonly List<GridEntry> FilterRemoved = new();
+		private string FilterString;
+		private Action Sort;
+		public FilterableSortableBindingList(IEnumerable<GridEntry> enumeration) : base(enumeration)
+		{
+			//This is only necessary because SortableBindingList doesn't expose Sort()
+			//You should make SortableBindingList.Sort protected and remove reflection
+			var method = typeof(SortableBindingList<GridEntry>).GetMethod("Sort", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
```
rmcrackan, line 34: Pushing that update now. Nuget builds are slow this time of day

rmcrackan:
While I'm waiting on those nuget changes to percolate up the chain, can you give me the quick version of what changed in products grid?

Source/LibationWinForms/grid/FilterableSortableBindingList.cs
rmcrackan: Push complete. Merge to see it

Mbucari:
> While I'm waiting on those nuget changes to percolate up the chain, can you give me the quick version of what changed in products grid?

- All filtering hackery removed.  Filtering is done by using BindingSource.Filter.
  - It is no longer required to re-filter the list after sorting
  - Filter(0
  - Display() is simplified.  No ordering or the library books is done here.
  - updateGrid() and Filter() are simplified.  They have the same responsibilities as before, but now they're executed using the proper filter methods
- All handling of Liberate_Click is moved to Form1.ProcessQueue

Mbucari:
I'm done with all comments.

Source/LibationWinForms/grid/FilterableSortableBindingList.cs
Mbucari: Christ I'm slow today.  I didn't understand I needed to merge your changes. Done now.

rmcrackan:
Kids are all finally in bed. I can look at this again.

I see that InvalidateRow is gone. How does this new version handle `existingItem.UpdateLibraryBook(libraryBook)` which updates all GridEntry fields without mutating the list which contains it?

Mbucari:
It's still there. Line 190

Mbucari:
I'm sorry, I misunderstood. You don't need to invalidate a row when the library book changes because GridEntry subscribes to UserDefinedItem.ItemChanged and notifies the view to update.

rmcrackan:
> I'm sorry, I misunderstood. You don't need to invalidate a row when the library book changes because GridEntry subscribes to UserDefinedItem.ItemChanged and notifies the view to update.

One of the more common fields I need updated is user rating (aka "My Rating") which is not part of UserDefinedItem. It seems like this new way wouldn't work for that field. It's not a "UserDefinedItem" because as far as Libation is concerned, this is data from audible not from the user. I think a few other fields can infrequently change also like series, category, and cover art -- I'd have to verify those in BookImporter though.

The "My Rating" field is a big one for me. One of my more common uses of Libation is to rate a book on audible, do a new scan, then my default filter hides those books, then pick my next listen from the visible items.

Mbucari:
Looking at GridEntry, it looks like the LibraryBookUpdated event is only fired from the constructor. It's basically unused.

That said, ALL fields will update when the DataGridView is redrawn/painted. NotifyPropertyChanged just forces the view to update when there was no other reason for it to update (eg scrolling or mouse over)

Mbucari:
MVVM is both push and pull.  When the view knows that it is out of date (eg by scrolling fields into view that were previously hidden), it pulls the current state from the view model. When the view model knows that the view is out of date (eg because a model value was changed), it pushes the current state to the view.

rmcrackan:
LibraryBookUpdated happens to be fired in the ctor, but obviously nothing could have subscribed to it yet. The event's real purpose was:

UpdateLibraryBook \> setLibraryBook \> LibraryBookUpdated \> InvalidateRow

> ALL fields will update when the DataGridView is redrawn/painted. NotifyPropertyChanged just forces the view to update when there was no other reason for it to update

Are you sure about that? I thought that's why they field name was included in the pattern: so only that field would be updated

Mbucari:
> Are you sure about that?

I posted more details before you asked your question. In case you didn't see, look above your last comment.

If you want to see an under-the-hood example, look at ProcessQueue. I did something very similar to what I described above.. When VirtualFlowControl scrolls, it fires RequestData becaus eit knows that there are new books in the field of view, and it needs to update the view to show them.  When liberation or queue state changes, ProcessBook fires PropertyChanged so the view can show the changed state.

rmcrackan:
I'll do some testing tomorrow if I remember. What I'm finding looks like the purpose of the PropertyChangedEventArgs(propertyName) is so only that propertyName is updated. I assumed that's why GridEntry.UserDefinedItem_ItemChanged specified each property instead of having a non-specific catch-all at the end.

Mbucari:
>... So only that propertyName is updated
Right. It's a waste of resources if you tell the view that it needs to update everything for that view model entry. But it's also the only way that the view model can tell the view to update (push).

rmcrackan:
My understanding was: bindinglist.add and bindinglist.remove know that it can only be a full row change. PropertyChangedEventArgs updates only the specified field. I did some testing which seemed to bear this out. This is why I wired up the InvalidateRow.

It's possible that I simply implemented it wrong. This isn't a paradigm I'm used to. In the backend, which is where I spend my work life, completely different patterns are used.

Mbucari:
You could consider invalidating a row a form of push. It's really telling the view that it is time to pull in the current state from the view model. But like I said before, a simple mouse over or scroll in the view will do the same thing. But yeah, I think the best approach right now is just to do some testing and see if everything behaves as you expect/want it to behave. If it doesn't come, we can adjust then.

rmcrackan:
You were right -- [test code](https://github.com/rmcrackan/Libation/files/8704753/binding_test.zip). This defied my expectations. Summary: **if** the specified field exists, update only that field in the UI. Else: update all. See comments in `Name`

Issue #251, opened 2022-05-18
Download Stuck


Tb3986:
Hello,
i would like to report a bug very annoying. Download is stuck and don't won't to move all files from temp folder to the final one.

![Untitled](https://user-images.githubusercontent.com/45364981/169017405-0fed09f4-d0c2-4dab-aa16-599353741299.png)

With the new update, 3 downloads, 3 stucks.

Here the log

https://we.tl/t-H5OKsApPez

Regards

EDIT:
Tried to use another temp folder and final folder, same issue.

rmcrackan:
@Tb3986 Be happy to look into this. That site isn't letting me download your log file. Can you attach it here on github directly? You can drag/drop it into the text box.

@Mbucari Ya know those lack of bug reports we were talking about... :worried:

Tb3986:
[Log202205_001.log](https://github.com/rmcrackan/Libation/files/8716152/Log202205_001.log)

Done

Tb3986:
> @Tb3986 Can you please change your logging level to "Verbose", retry the same book, and post those logs?

[Log202205.log](https://github.com/rmcrackan/Libation/files/8718110/Log202205.log)

Mbucari:
Fixed in #252

rmcrackan:
Fixed in latest version [v7.6.2](https://github.com/rmcrackan/Libation/releases/latest)

PR #252, opened 2022-05-18
Fixed Issue #251

Mbucari:
This was a bug in AAXClean with some multipart files that would happen if the ending timestamp of the last chapter from metadata didn't match the timestamp of the last audio frame in the file.

### DIFF
--- a/Source/LibationWinForms/Form1.ProcessQueue.cs
+++ b/Source/LibationWinForms/Form1.ProcessQueue.cs
@@ -23,7 +23,18 @@ namespace LibationWinForms
 
 		private void ProductsGrid_LiberateClicked(object sender, LibraryBook e)
 		{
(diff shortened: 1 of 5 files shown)


Source/LibationWinForms/Form1.ProcessQueue.cs
```diff
@@ -23,17 +23,7 @@ private void Configure_ProcessQueue()
 
 		private void ProductsGrid_LiberateClicked(object sender, LibraryBook e)
 		{
```
rmcrackan, line 25: Per email conversation: I think nothing is needed here. I dunno what the hell went wrong this morning

Mbucari: Right. Reverted.

PR #253, opened 2022-05-18
Revert that unnecessary change

Mbucari:
Might want to accept this PR before releasing.  Just realized that if the pdf isn't downloaded but the audiobook is found, the PDF download will not be queued.


rmcrackan:
D'oh! Let's see if I can do another release before anyone notices :)

rmcrackan:
First time I've ever released two versions in 5 minutes, lol

PR #254, opened 2022-05-18
Update installer

Mbucari:
I made a package installer here: https://github.com/Mbucari/Upgrader

It's nothing fancy.  It just takes 3 command line arguments:

[zip file url] [extraction directory] [optional: launch exe]

If the user clicks "yes" to upgrade, Libations will download Upgrader from my repo, launch it with those arguments, then exit.


rmcrackan:
I'm trying to go through the possibilities: If you extract to your same folder, it'll try to overwrite but fail because you've got files locked. If you extract elsewhere, then you'll have the old program still running when the new one launches. I guess this will eliminate a few clicks. I'll think it over. Really I'd like to have something like autoupdater.net or really anything battle-hardened. I've written my own upgrader before and the recovery edge cases are brutal.

btw -- you have some libation stuff in that repo's json file

Mbucari:
I put Upgrader.exe in the master repo so that the hard link in Libation is always the most recent version. It's also a standalone exe.

Issue #154, opened 2021-11-11
Episodes in grid


rmcrackan:
## New way to display episodes with parents

Currently every episode is shown individually and is a complete eyesore. Episodes without parent: still show as a normal row. Episodes with parent:

Parent has single grid entry/row. Replace stoplight with "+" or "-" image. Button expands/collapses children in new rows beneath. Actually just shows or hides them. They'll need to already be in the correct order (addressed below).

Only parents are affected by sorting and filtering.
* Sort: When expanded, episodes always appear in series order below parent irrespective of other grid sorting
* Filter: If parent is shown, all children are available to be shown. If parent is hidden, all children are hidden

sorting idea 1: https://www.codeproject.com/Articles/18440/DataGridView-Multi-column-Sort

sorting idea 2: bindingSource.Sort = "[Street Name], [Street Direction], [Street Number]";

### Sorting example
* Metamorphosis. podcast analyzing Kafka. child. episode 1. series id 789
* Metamorphosis. podcast butterflies. child. episode 2. series id 123
* ABC book
* Metamorphosis. podcast butterflies. parent. series id 123
* XYZ book
* Metamorphosis. podcast butterflies. child. episode 1. series id 123
* Metamorphosis. book by Ovid. no episode series
* Metamorphosis. podcast analyzing Kafka. child. episode 2. series id 789
* Metamorphosis. podcast analyzing Kafka. parent. series id 789

#### Sort level 1: user-defined. Eg: click column

CreateMemberValueDictionary() will have to use same values for parents and all children

* **ABC book**
* Metamorphosis. podcast analyzing Kafka. child. episode 1. series id 789
* Metamorphosis. podcast butterflies. child. episode 2. series id 123
* Metamorphosis. podcast butterflies. parent. series id 123
* Metamorphosis. podcast butterflies. child. episode 1. series id 123
* Metamorphosis. book by Ovid. no episode series
* Metamorphosis. podcast analyzing Kafka. child. episode 2. series id 789
* Metamorphosis. podcast analyzing Kafka. parent. series id 789
* **XYZ book**

#### Sort level 2: episode series. User has no control over this and subsequent sort levels

* ABC book
* **Metamorphosis. book by Ovid. no episode series**
* Metamorphosis. podcast butterflies. child. episode 2. series id 123
* Metamorphosis. podcast butterflies. parent. series id 123
* Metamorphosis. podcast butterflies. child. episode 1. series id 123
[...]

PR #255, opened 2022-05-23
Implemented Episode grouping and refactored ProductsGrid

Mbucari:
- Episodes grouped by Series
  - I implemented this pretty much how you had described in your original post in #154 
  - ```GridEntry``` is now abstract and is extended by ```LibraryBookEntry``` and ```SeriesEntry```
  - Filtering filters on individual episodes
  - Expanding and collapsing uses the filtering infrastructure.  If children are collapsed, they are considered filtered out.
  - I had to edit ```SortableBindingList```.  The new version is ```SortableBindingList1```
- Refactored ```ProductsGrid``` into ```ProductsGrid``` and ```ProductsDisplay```
  - ```ProductsGrid``` now handles all direct interaction with the DataGridView, such as
    - Column customization
    - Data binding and updating
    - Filtering
    - Cell click handling and firing column-specific click events
  - ```ProductsDisplay``` handles column-specific click events and is what Form1 interacts with
  - Fixed the designer error so now the grid is added to Form1 in design mode.
-  Fixed the filtering error you described in #154

TODO:  The only series-specific information available is what's ioncluded with the book info.  I looked at @mkb79's api documentation, and I don't see a way to query info about a series directly.  This is a problem because series have their own overall rating, description, and cover art that differs from the values in any episode.  Right now,  SeriesEntry just takes that info from the Books.  But it would be better to query and display that info from the series itself.


Source/LibationWinForms/Form1.cs
```diff
@@ -12,24 +12,10 @@ namespace LibationWinForms
 {
 	public partial class Form1 : Form
 	{
-		private ProductsGrid productsGrid { get; }
-
 		public Form1()
 		{
 			InitializeComponent();
 
-			if (this.DesignMode)
```
Mbucari, line 21: From my research, this will never be true inside a constructor.

rmcrackan: ctor and Load are called in DesignMode for inherited forms. To be consistent, either the ctor check can be removed or I should add defensive checks on each Load subscription

Source/AppScaffolding/AppScaffolding.csproj
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup>
     <TargetFramework>net6.0-windows</TargetFramework>
-    <Version>7.7.1.1</Version>
```
rmcrackan, line 6: Version shouldn't get smaller

Source/LibationWinForms/Form1.QuickFilters.cs
```diff
@@ -56,5 +50,11 @@ private void firstFilterIsDefaultToolStripMenuItem_Click(object sender, EventArg
 		private void addQuickFilterBtn_Click(object sender, EventArgs e) => QuickFilters.Add(this.filterSearchTb.Text);
 
 		private void editQuickFiltersToolStripMenuItem_Click(object sender, EventArgs e) => new EditQuickFilters().ShowDialog();
+
+		private void productsDisplay_InitialLoaded(object sender, EventArgs e)
```
rmcrackan, line 54: Where does this get called?

Mbucari: It's called in the designer generated code.

Source/AppScaffolding/AppScaffolding.csproj
Mbucari: Yeah, I'll revert this change. This is a leftover from when I was testing the automatic upgrader.

Source/LibationWinForms/grid/SortableBindingList1.cs
```diff
@@ -0,0 +1,87 @@
+﻿using Dinah.Core.DataBinding;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+
+namespace LibationWinForms
+{
+	internal class SortableBindingList1<T> : BindingList<T> where T : class, IMemberComparable, IHierarchical<T>
```
rmcrackan, line 9: Should I just remove the sortable binding stuff from dinah.core?

Source/LibationWinForms/grid/GridEntry.cs
```diff
@@ -1,101 +1,81 @@
...
 {
-	/// <summary>
-	/// The View Model for a LibraryBook
-	/// </summary>
-	internal class GridEntry : AsyncNotifyPropertyChanged, IMemberComparable
+	public interface IHierarchical<T> where T : class
 	{
-		#region implementation properties NOT exposed to the view
-		// hide from public fields from Data Source GUI with [Browsable(false)]
+		T Parent { get; }
```
rmcrackan, line 15: I was hoping to diligently pour through the changes in this class but I gotta be honest, there's too much changed. It'll be easier to pull it locally.

Source/LibationWinForms/grid/SortableBindingList1.cs
Mbucari: Probably. If it's going to be useful to you for other projects then keep it, but at this point it's so tailored to the libation implementation that it doesn't really make sense for it to be generic anymore.

Source/LibationWinForms/grid/GridEntry.cs
Mbucari: Understood. Moving SortableBindingList into libation and making it non-generic will allow for cleaner code, and this interface can go away.

rmcrackan: Let me know of anything you want moved from dinah.core.databinding => libation and I'll kill it in dinah.core. It's only being used in Libation.

Source/LibationWinForms/Form1.QuickFilters.cs
rmcrackan: Makes sense. By default the merge tool in github hides files with huge changes which always includes autogen stuff like that

rmcrackan:
I'll hopefully give this the attention it deserves tomorrow. I'm looking forward to seeing what it looks like

Source/LibationWinForms/Form1.cs
Mbucari: Looks like this was the only instance of DesignMode being checked in a Form constructor. Everywhere else it's only checked in the Load event handler.

Source/LibationWinForms/grid/GridEntry.cs
Mbucari: Keep everything in Dinah for now.  It;'s still used by RemobeBooksDialog.  We can migrate that DataGridView to the new binding later, but quite frankly I don't want to deal with that right now.

rmcrackan:
> TODO: The only series-specific information available...

tl;dr: Don't sweat it. What you did is great. Audible's hierarchy on this is a f--king mess.

The long version:

They have 3 paradigms for what a 'series' is. I'm done playing whack-a-mole with this particular issue. There's a book series, eg: Harry Potter. There's a newer type of series for podcasts. And there's an old version of this podcast-like series that they used to use before their recent actual podcast offering -- in these you can own a child without owning the parent and there might be some other minutiae. These last 2 both have parent-child relationships. The parent has a book-like entity that's not really a book but can have ratings. And depending on audible keeps tweaking their UIs, sometimes it's easier or harder to rate parents vs children.

Mbucari:
> Gotta say, my dude, this is pretty damn slick. Very smooth

So glad you like it! I ran into so many weird errors while putting this together, and I'm still a little worried that something is going to break in a weird way for no apparent reason.  It just needs a lot of testing to ferret out those edge cases.


>     1. ideally children should always be sorted by series index. 

Done. Previously, children were sorted in the same way as the whole grid but would remain grouped together beneath their parent.  Switching back and forth between that approach and always sorting by index is easy enough.

>     2. SortableBindingList, SortableBindingList1, etc

That will come if/when ```RemoveBooksDialog``` no longer uses ```SortableBindingList``` from Dinah. Right now it does, and I'm inclined to just leave it.  I merged ```SortableBindingList1``` and ```FilterableSortableBindingList``` into a new single class: ```GridEntryBindingList```

>     3. general grid bug, not feature-specific: create a filter to omit a certain tag

Fixed.  Inside ```LibraryBookEntry.UserDefinedItem_ItemChanged```, I call ```SearchEngineCommands.UpdateBookTags(Book);``` and ```SearchEngineCommands.UpdateLiberatedStatus(Book);```. Then inside ```GridEntryBindingList```, I re-filter on item changed.

>     4. UX: scroll so parent "+" row is near the top. Click +. It jumps to the bottom. Is this fixable?

Fixed.  The reason it behaved as it did was to try to avoid an error that would occur if I didn't reset the whole list. Apparently some refactoring I did at some point in the past day fixed that problem. See what I mean about weird errors?

>     5. Children: lack of any indentation is confusing, esp if parent is scrolled offscreen. Help me brainstorm. You know my design skills are hot garbage but here are some initial mock ups. Thoughts?

Please don't take offense, but those are all super ugly lol.  I colored the background of series book entries a light blue.  It's not very pretty, but it does the job and looks OK.

rmcrackan:
> I'm still a little worried that something is going to break in a weird way for no apparent reason

You don't need to worry it might happen; it's a certainty. :) All we can do is test and make the code easier to read than it is to write.

> Please don't take offense, but those are all super ugly lol

They are for sure an abomination. Just starting the brainstorming.

> I colored the background of series book entries a light blue

Looks like we replied here at almost the same time. See my feedback above on that:

> creative. Let's go more pale. At first I thought I was witnessing a bug where everything got highlighted and I couldn't figure out why I was unable to de-select it

rmcrackan:
> Please don't take offense, but those are all super ugly lol

You aren't in awe at my amazing design and photoshop skills?! :surprised-pikachu:

Mbucari:
> You aren't in awe at my amazing design and photoshop skills?! :surprised-pikachu:

they're about as good as mine, as evidence by the process queue up-down arrows and my plus/minus icons.  Feels like something straight off Softpedia circa 2005.

Mbucari:
> Let's go more pale. 

Done. Went with a very pale green to contrast with selection blue.

PR #257, opened 2022-05-24
Fix hang on grid update

Mbucari:
The problem was caused by my fix to the filtering issue you described in item 3 of https://github.com/rmcrackan/Libation/pull/255#issuecomment-1136038580

The fix was that when an item in the binding list changed,  re-filter the entire library and check the book that changed against the search results. However, when UpdateGrid runs, it updates almost every book in the library and triggers the item changed, so SearchEngineCommands.Search was re-run for every single book in the library.

The fix is to introduce the ```SuspendFilteringOnUpdate``` property to ```GridEntryBindingList```. Set it to true when we're updating the display, then back to false when we're finished.

### DIFF
--- a/Source/LibationWinForms/Form1.VisibleBooks.cs
+++ b/Source/LibationWinForms/Form1.VisibleBooks.cs
@@ -33,17 +33,11 @@ namespace LibationWinForms
 				{
 					liberateVisibleToolStripMenuItem.Format(notLiberated);
 					liberateVisibleToolStripMenuItem.Enabled = true;
-
-					liberateVisible2ToolStripMenuItem.Format(notLiberated);
-					liberateVisible2ToolStripMenuItem.Enabled = true;
 				}
 				else
 				{
 					liberateVisibleToolStripMenuItem.Text = "All visible books are liberated";
 					liberateVisibleToolStripMenuItem.Enabled = false;
-
-					liberateVisible2ToolStripMenuItem.Text = "All visible books are liberated";
-					liberateVisible2ToolStripMenuItem.Enabled = false;
 				}
 			});
 		}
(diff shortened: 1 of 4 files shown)


Source/LibationWinForms/Form1.VisibleBooks.cs
```diff
@@ -33,17 +33,11 @@ void setLiberatedVisibleMenuItem()
 				{
 					liberateVisibleToolStripMenuItem.Format(notLiberated);
 					liberateVisibleToolStripMenuItem.Enabled = true;
-
-					liberateVisible2ToolStripMenuItem.Format(notLiberated);
-					liberateVisible2ToolStripMenuItem.Enabled = true;
```
rmcrackan, line 38: Why were the liberateVisible2ToolStripMenuItem lines removed?

Mbucari: Because I'm an idiot and didn't realize they were different from liberateVisibleToolStripMenuItem. Reverted.

rmcrackan: In fairness: I knew that stupid name would bite me one day. I'll change it soon

rmcrackan:
Thank you for the quick fix! Testing looks great.

Mbucari:
Woot! VS CPU profiler to the rescue again!

Source/LibationWinForms/Form1.VisibleBooks.cs
rmcrackan: renamed -- https://github.com/rmcrackan/Libation/commit/7987dfb81959c486c1b15c1d3787d2234aa7e134

PR #260, opened 2022-05-26
Throttle episode scanning to 10 concurrent scans.

Mbucari:


Mbucari:
I learned a hell of a lot about async programming during this update. Still a bit of a mindfuck though.  So many invisible walls I can't see very well by just reading the code. I'm glad I finally discovered semaphores, though! I can't believe I hadn't looked into them before now.  They'll make a lot of what I do much easier!

rmcrackan:
Tread lightly with semaphores and mutexes. Are they good? Without a doubt! But they are also seductive and can make you think you need them when something more simple is better.

Mbucari:
> But they are also seductive

Undoubtedly. I'm already brainstorming how I can use them more lol. But in these cases they are absolutely necessary. I tried out more than five different async architectures to get this to work, and a semaphore was the only solution that was not a complete hack and framework abuse.

rmcrackan:
Before I looked at your code, I was about to recommend looking into the ridiculously named "SemaphoreSlim" -- looks like you beat me to it :)

Source/AudibleUtilities/ApiExtended.cs
```diff
@@ -119,32 +120,33 @@ private async Task<List<Item>> getItemsAsync(LibraryOptions libraryOptions, bool
 		{
 			var items = new List<Item>();
 
-			Serilog.Log.Logger.Debug("Begin library scan");
+			Serilog.Log.Logger.Debug("Beginning library scan.");
 
 			List<Task<List<Item>>> getChildEpisodesTasks = new();
 
-			int count = 0;
+			int count = 0, maxConcurrentEpisodeScans = 10;
```
rmcrackan, line 127: Let's drop this to 5. Ten would *probably* be ok but I want to be conservative. I live in fear of amazon's reputation for IP banning. I have a very long history with site scrapping and they are infamous. Granted, we're piggy backing on audible's mobile api instead of scraping, but I still worry.

In the very first version of what would eventually be Libation, I routed all calls through a "Humanizer" which would actually add a random wait time between each call. That proved to be overkill but it shows my commitment to the idea.

Mbucari: Done

Mbucari:
I started this update with the goal of exceeding the 10,000 title limit, but along the way the goalpost moved to speed and concurrency.  My limited tests _suggest_ that this will exceed the 10,000 book limit, but I have no way of testing that.  At 50 titles per page, that's >200 pages required to exceed 10,000 titles.  I've tested up to 54 pages with success.  I just do not have enough books to test 200+ pages. But you do.

In ApiExtended.getItemsAsync, change this line:

```C#
await foreach (var item in Api.GetLibraryItemAsyncEnumerable(libraryOptions))
```

To this...

```C#
await foreach (var item in Api.GetLibraryItemAsyncEnumerable(libraryOptions, 5, 10))
```

And see if you can get all your items.

rmcrackan:
This seems like a huge success. Any final changes before I merge?

Mbucari:
Great!

No more changes on my end for now. Get er done.

PR #264, opened 2022-05-26
Fix for episodes with no series link

### LINKED ISSUE #263: Getting an error when opening 7.10.0.2
A6D959FE-7D64-4FF0-93F9-FD50E3C2EB0A

Hitting continue only displays blank screen.

Mbucari:
This will fix error #263

The problem, as I explained in the issue thread, is that if a parent had 0 episode, the parent was added to the database and it had no series link. This can happen if a podcast has only 1 episode, eg https://www.audible.com/pd/Episode-1-Podcast/B08ZXVZ7KQ

In that case, the parent is it's own child. For this special case where a series has no children, Libation clones the parent and adds it as a child.

getItemsAsync will add all parents to the db as well as their children.  Parents' ASINs are changed to "SERIES_[ASIN]" to better distinguish it from items that actually contain content, and so that in the special case where a parent is its own child, the ASINs do not collide.

Then in products grid, instead of checking ContentType is ContentType.Episode, use the new IsEpisodeChild().  There's a corresponding IsEpisodeParent() which I plan to use in the future to get series-specific info as I described in the TODO note in #255

### DIFF
--- a/Source/AudibleUtilities/ApiExtended.cs
+++ b/Source/AudibleUtilities/ApiExtended.cs
@@ -137,6 +137,7 @@ namespace AudibleUtilities
 					//Add the parent to the library because it contains the series
 					//description, series rating, and series cover art which differ
 					//from the individual episodes' values.
+					item.Series = new Series[]{ new Series { Asin = item.Asin, Title = item.TitleWithSubtitle } };
 					items.Add(item);
 				}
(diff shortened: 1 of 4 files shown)


rmcrackan:
I like what I see. Thanks for the quick turnaround while I was away

Issue #265, opened 2022-05-27
263 post mortem


rmcrackan:
Let's discuss this after #263 is fully resolved

1) what's the right solution?
2) are there implications of rolling that back the SERIES_ hack? Esp how it will affect this user in particular?
3) what went wrong?
3A) I think the short version is: audible changed something weird and unpredictable. The nature of reverse engineering is being reactionary. What could we do differently? probably nothing
3B) exceptions weren't handled correctly. I'll address the exceptions part soon
3C) a user was left in a bad state. Is there a way to help a user out of a bad state? does Libation need to deploy with a helper app? Like: something where we say: open LibationOhShit.exe and paste in `rollback to v7.0` or `update Books set brain=false where foo=bar`

Mbucari:
1. I want to add Parents to the library. I do not want to alter their Asins.  I think ContentType should have a new enum field: ContentType.SeriesParent, and only Items with RelationshipToProduct.Parent should be assigned that enum value. 
This leaves the problem of the special case where a parent is also its own child. I'd be inclined to add the item twice, once as ContentType.SeriesParent and once as ContentType.Episode, but the DB implementation prevents duplicate ASINS. Is that an insurmountable problem? I remember a while back you said ASINs are not PKs, so maybe we can have dups.

2. I tested this bu removing the "SERIES_" prepend, and the effect is that those items remain in the db but don't affect anything. There is a duplication problem which I hadn't notice before (doh!) that happens if the db was upserted once with [ASIN] and once with SERIES_[ASIN]. That can be fixed with Remove Books Dialog. But nothing breaks.

3. The problem was 3-fold.

First, Audible has a poor system for indexing podcasts.  For most podcasts, there is a parent ASIN which is retrieved from the library api endpoint, and children ASINs that are retrieved from the catalog api endpoint.  What they SHOULD have done for podcasts with only 1 episode is to make 2 separate ASINs: 1 for the series and 1 for the episode.  What they ACTUALLY did was make 1 ASIN for that episode, and no ASIN for a containing series.

Second, whem the importer found an episode type that had no children, it just added the book to the library with no Series information.  This was a bad idea because, even though it wasn't linked to any other ASINs, it is a series **type** that should have a series identifier.

Third, when I made the new ProductsGrid I relied on a (now proven wrong) assumption that any ContentType.Episode book would have at least 1 SeriesBook.  That didn't happen because of the choice made in "Second" above.

Mbucari:
Re No.1

[...]

PR #268, opened 2022-06-07
Address issues in 263

Mbucari:
There's a lot to unpack here, so I'll just go buy bullets and hope I get everything.

1. Added a migration to remove books and series that have the "SERIES_" prefix from the db. The effect of this migration, along with new Parent content type (Items 2 and 4 below) is that when this version is launched, no episodes will appear in the grid.  All episodes are still in the db along with any `UserDefinedItem`, but they are not added to the grid because they don't have any parents.  This is solved after the first library sync, after which all episodes will be back in the grid and intact.
2. Added `Parent` to `ContentType` enum with value of 4.  Flags are not needed (as explained in 3.2), but I thought I'd preserve the options for flags in the future.
3. Edited `ApiExtended` to import episodes and series parents
   1. Using the new `Item.IsSeriesParent` [from my AudibleApi PR](https://github.com/rmcrackan/AudibleApi/pull/28), `getItemsAsync` retrieves individual episodes as well as the series parent.
   2. I was wrong about [the first part of what went wrong](https://github.com/rmcrackan/Libation/issues/265#issuecomment-1139205952). What ACTUALLY went wrong is that an individual **episode** of a podcast was added to the library and not the podcast **series**. When a series is added to the library, the Api's library endpoint returns just the series item and not any of the series' episodes.  But users are also allowed to just add individual episodes and not whole series to their library. When they do that, the library endpoint returns individual episodes. To handle these cases inside Libation, I added code in the `getChildEpisodesAsync` to retrieve the series parent from the catalog endpoint and add it to the library. So there is no need to make `ContentType` private, and no DB migration is required. As far as I can tell at this time, there are no orphan episodes in the Audible catalog.
   3. `GetLibrary_Flat_NoTracking()` now has an overload to include `ContentType.Parent` in the returned books. Default is to not include parents. The only place where we query the library with parents is in `ProductsDisplay.Display()`
4. Modified GridView classes to use new `ContentType.Parent`
   1. Modified `SeriesEntry` to use series-specific values for everything except `PurchaseDate` and `Length`.
   2. Refactored `SeriesEntry`, `LibraryBookEntry` and `GridEntry` for more code reuse.
   4. Clicking on a `SeriesEntry` cover and description now launch windows for full display like they already do for `LibraryBookEntry`
   5. Refactored `ProductsGrid.UpdateGrid` for clarity and safety, and to use the the new parent content type.
   6. Moved queries into QueryExtensions and added some new ones.
5. Sped-up adding multiple books to `ProcessQueueControl`. UI no longer hangs. This should also resolve an uncommon error that could happen while liberating a large batch if the first book in queue completed before the queue was finished populating.

### DIFF
--- a/Source/LibationWinForms/ProcessQueue/TrackedQueue[T].cs
+++ b/Source/LibationWinForms/ProcessQueue/TrackedQueue[T].cs
@@ -226,5 +226,14 @@ namespace LibationWinForms.ProcessQueue
 				QueuededCountChanged?.Invoke(this, _queued.Count);
 			}
 		}
+
+		public void Enqueue(IEnumerable<T> item)
+		{
+			lock (lockObject)
+			{
+				_queued.AddRange(item);
+				QueuededCountChanged?.Invoke(this, _queued.Count);
+			}
+		}
 	}
 }
--- a/Source/LibationWinForms/ProcessQueue/TrackedQueue[T].cs
+++ b/Source/LibationWinForms/ProcessQueue/TrackedQueue[T].cs
@@ -181,7 +181,7 @@ namespace LibationWinForms.ProcessQueue
 
 		public bool MoveNext()
 		{
-			int queuedCount, completedCount = 0;
+			int completedCount = 0, queuedCount = 0;
 			bool completedChanged = false;
 			try
 			{
@@ -208,8 +208,8 @@ namespace LibationWinForms.ProcessQueue
 			finally
 			{
 				if (completedChanged)
-					CompletedCountChanged?.Invoke(this, _completed.Count);
-				QueuededCountChanged?.Invoke(this, _queued.Count);
+					CompletedCountChanged?.Invoke(this, completedCount);
+				QueuededCountChanged?.Invoke(this, queuedCount);
 			}
 		}
 
--- a/Source/LibationWinForms/ProcessQueue/TrackedQueue[T].cs
+++ b/Source/LibationWinForms/ProcessQueue/TrackedQueue[T].cs
@@ -85,13 +85,18 @@ namespace LibationWinForms.ProcessQueue
 
 		public bool RemoveQueued(T item)
 		{
+			bool itemsRemoved;
+			int queuedCount;
+
 			lock (lockObject)
 			{
-				bool removed = _queued.Remove(item);
-				if (removed)
-					QueuededCountChanged?.Invoke(this, _queued.Count);
-				return removed;
+				itemsRemoved = _queued.Remove(item);
+				queuedCount = _queued.Count;
 			}
+
+			if (itemsRemoved)
+				QueuededCountChanged?.Invoke(this, queuedCount);
+			return itemsRemoved;
 		}
 
 		public void ClearCurrent()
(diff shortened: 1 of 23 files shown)


rmcrackan:
AudibleApi nuget is updated -- you can update that now

rmcrackan:
> Flags

I have personal to-do items to: convert ContentType to a flag and to hide this impl. detail (private/protected EF-friendly field). Are either of these still needed? It seems like no since the stuff I was thinking of is now in QueryExtensions. If we do make them flags later, I'll refactor this so it's a canonical part of Book and hide the flag

> What ACTUALLY went wrong is that an individual episode of a podcast was added to the library and not the podcast series.

Damn, I've run across this before but I thought I'd fixed it. Maybe in the recent churn it was overlooked. Oh well. Glad it's fixed again

> As far as I can tell at this time, there are no orphan episodes in the Audible catalog

I'm sure audible's undocumented api will find some way to eff this up later lol

> Moved queries into QueryExtensions and added some new ones.

These look great. I think this file should live in the data layer though.

> Sped-up adding multiple books to ProcessQueueControl

Sweeeeeet. Love me some speed increases. Thanks!

Mbucari:
> Flags. Are either of these still needed? 

Nope! We just have a new content type. Otherwise everything is pretty much as it was.

> I think this file should live in the data layer though.

Done. Obviously we need to keep queries operating on GridEntry types inside LibationWinForms, but all other are moved to DataLayer.

rmcrackan:
> Obviously we need to keep queries operating on GridEntry types inside LibationWinForms, but all other are moved to DataLayer.

Good call. I must have missed those.

rmcrackan:
Since this is no longer in 'draft', is it ready for prime time?

Mbucari:
> is it ready for prime time?

Yup! This is probably my most polished PR in a long time.  I think I had gotten a little sloppy, but this one is very tight.

Source/LibationWinForms/ProcessQueue/TrackedQueue[T].cs
```diff
@@ -226,5 +226,14 @@ public void Enqueue(T item)
 				QueuededCountChanged?.Invoke(this, _queued.Count);
 			}
 		}
+
+		public void Enqueue(IEnumerable<T> item)
+		{
+			lock (lockObject)
+			{
+				_queued.AddRange(item);
+				QueuededCountChanged?.Invoke(this, _queued.Count);
```
Comment (rmcrackan) [inline Source/LibationWinForms/ProcessQueue/TrackedQueue[T].cs:235]: Firing an event inside a lock can deadlock -- please move event outside of lock

Comment (Mbucari) [reply Source/LibationWinForms/ProcessQueue/TrackedQueue[T].cs]: Done.

Comment (rmcrackan) [reply Source/LibationWinForms/ProcessQueue/TrackedQueue[T].cs]: Thanks for finding those others also

Mbucari:
Even though there's nothing really new here, I think you should make this a release. I noticed that if I added 2 episodes from the same series (without adding the series), Libation would interpret it as 2 different series and duplicate it in the grid view. It doesn't cause an error, but it is a display bug.

rmcrackan:
> release

Yup. Working on that as we speak

Issue #270, opened 2022-06-11
7.10+ does not update grid with new books after library import


flysalad:
Versions 7.10.2 and 7.11 both fail to update the grid to show new books when using the import function. I have to restart Libation to get the grid to show the new books. Other functions I use seem to be fine.

[Log202206.log](https://github.com/rmcrackan/Libation/files/8883081/Log202206.log)

Mbucari:
@rmcrackan This one is a bit sticky.

The offending episode (of course it's an episode) is B01ERWOYXU. It's in his library, but it has been removed from audible.

The way Libation retrieves parents for individual episodes is by looking for parent's asin in the episode's relationships, querying the parent from the catalog api endpoint, then adding the parent to the database. Prior to 7.11, individual episodes weren't assigned a series with a parent asin. This isn't a problem when episodes are still in the user's library, because a fresh library scan will find and add the series. The problem for him is that episode is no longer on audible, so its db entry has no series link with parent asin, and we'll never be able to find it again.

Moving forward as audible removes more episodes from podcasts, this will not be a problem because the episodes' series link will already have been added to the db.

But for this legacy problem, the only way I can think to handle this is to add a catch-all series for orphaned episodes. Maybe add it in a migration.

Mbucari:
@flysalad Can you please share your LibationContext.db file with me? I'd like to see how your episodes are being stored after 2 attempts to fix this problem.

The file contains all your library books, custom tags and liberated status, but no personally identifying information.

flysalad:
Zipped LibationContext.db atttached, as requested.

[LibationContext.zip](https://github.com/rmcrackan/Libation/files/8883258/LibationContext.zip)

Mbucari:
Thank you @flysalad. And please delete that link, as I was wrong, it does contain your email address. Sorry! The DB guy is asleep.

Mbucari:
@rmcrackan 

OK, there's are even more legacy issues here than I realized. So back in the day before podcasts were supported, all items that were IsEpisodes were discarded from the import operation.  But for a long time, IsEpisodes worked like:

```C#
=> Relationships?.Any(r => r.RelationshipToProduct == RelationshipToProduct.Child && r.RelationshipType == RelationshipType.Episode)
			?? false;
```

Remember [my explanation of relationships](https://github.com/rmcrackan/AudibleApi/pull/28#issue-1263872961)? By only checking if an item had a relationship to a child, the importer:

[...]

PR #271, opened 2022-06-11
Move Remove Books function into main grid, added more db migrations and fixups for episodes

Mbucari:
## First the most pressing issue, episode series continue to cause pain.
This pr solves https://github.com/rmcrackan/Libation/issues/270.  I've tested that DB as it was given to me, and the migrations work well.  Only one series with 6 episodes was removed from the db because it is no longer on audible.

My fix is exactly as I outlined in that issue.:

- Detect series parents that were added to the database as books with ContentType.Episode, and change them to ContentType.Parent
- Then detect series parents that were added to the database as books with ContentType.Parent but are missing a series link, and add the link.
- For remaining orphans, scan catalog for all parent asins found in the orphans' series link.
  - If parent is found in the catalog, add it to the db as a series parent.
  - If a parent is not found, remove all episodes referencing that parent's asin from the db.

No episodes in that db made it past those migrations without having a series parent in the database.

Additionally, I've added more logging and catches around queries so that errors don't break the display.  If somehow an orphan does find its way through to the display logic, it'll be logged and ignored, and the rest of the library will display properly.

## Move Remove Books function into main grid

I added a checkbox column to the grid view called "Remove" and a corresponding Remove property to GridEntry. The checkbox column is tri-state using a new enum value.  This allows series to have an intermediate check state when some of the children are checked.

The scan and remove books process begins like it used to, but clicking Import > Remove...
From there, the Remove column is shown and the user can begin checking items as the library scan runs.  Once the scan finishes, libation will check the remove box for every item not returned by the library, but only books that haven't been liberated. Users may still remove liberated books, but they won't be auto-selected for removal.

When the remove column is shown, a "Remove \# Books" button appears in the tool strip, and a new "Done Removing Books" menu item appears in the Import menu. Clicking "Done Removing Books" will hide the Remove column and tool strip button.

## Import/Export from/to audible-cli.

I added a new class `AudibleUtilities.Mkb79Auth` which does all the heavy lifting of converting between itself and `Account`. 

In AccountsDialog, the grid view has a new button column called "Export". When a button in that column is clicked, that row's Account is converted to Mkb79Auth and user is prompted to save the json file.
On the bottom of AccountsDialog, I added a button to import from a json file.


Source/ApplicationServices/LibraryCommands.cs
```diff
@@ -27,10 +27,17 @@ static LibraryCommands()
...
-        public static async Task<List<LibraryBook>> FindInactiveBooks(Func<Account, Task<ApiExtended>> apiExtendedfunc, List<LibraryBook> existingLibrary, params Account[] accounts)
+        public static async Task<List<LibraryBook>> FindInactiveBooks(Func<Account, Task<ApiExtended>> apiExtendedfunc, IEnumerable<LibraryBook> existingLibrary, params Account[] accounts)
 		{
 			logRestart();
 
+			lock (_lock)
+			{
+				if (Scanning)
+					return new();
+				ScanBegin?.Invoke(null, accounts.Length);
```
Mbucari, line 38: Not sure about this one. I copied your code ImportAccountAsync, but I remember your warning about not doing this on my last PR.  You good with these?

rmcrackan: I'd just put the event invocation after the lock. Did I screw this up somewhere?

Mbucari: Yeah, one method down. I just copied you lol

rmcrackan: Darn. I don't want to step on your current work

Source/AppScaffolding/LibationScaffolding.cs
```diff
@@ -405,25 +405,58 @@ public static void migrate_to_v6_6_9(Configuration config)
 
 		public static void migrate_from_7_10_1(Configuration config)
 		{
```
Mbucari, line 407: I figure this stays in for a few release cycles. Gets the regular and irregular users databases migrated. Then it can be deleted. After that, the app still won't crash with a bad DB, but if they want all their stuff shown they have to delete and resync.

rmcrackan:
> On the bottom of AccountsDialog, I added a button to import from a json file.

Very cool. I hope this is useful to Marcus. For import/export, do we default to the user folder that he mentioned?

rmcrackan:
I finally went through the dozens of emails and code changes. After you read my previous notes, let me know when to pull the trigger. I'm ready when you are. fyi -- this will be Libation 8

Mbucari:
> I hope this is useful to Marcus.

I think it'll be useful to other users as well as an account backup and restore function through the gui. 

> do we default to the user folder that he mentioned?

We do now.

> Test the crap out of it.

I did. On her db, on my db, on empty db. Also, I've been doing all mu debugging with this migration running every launch. No problems yet!  Based on test results and just analyzing how the queries work, I don't see there being a problem. However, out of an abundance of caution I wrapped that migration code in try/catch block. If it fails, an entry is added to the config and it won't be attempted again. 

I did try to find a way to early return, but it's not possible without adding more queries that do basically the same thing, save for writing to the db.

Mbucari:
I think I addressed all your comments, but let me know if I missed something from another thread or an email.  Otherwise I'm done, and it's ready.

Mbucari:
I'm not sure the new "Remove # Books" button in the bottom status strip is obvious enough.  Run it and let me know what you think.

Another option is to add a button over the top of the search query text box.  Search is disabled while remove books is active, so there's a spot in the ui for it.

Mbucari:
I moved the remove books button.

PR #273, opened 2022-06-13
Add option to save episodes to series parent

Mbucari:
I hope I made this in time before Libation 8 goes live.  I've wanted this feature for a long time, and now that series parents are actually in the db as books, it's possible.

### DIFF
--- a/Source/AppScaffolding/AppScaffolding.csproj
+++ b/Source/AppScaffolding/AppScaffolding.csproj
@@ -3,7 +3,7 @@
 
   <PropertyGroup>
     <TargetFramework>net6.0-windows</TargetFramework>
-    <Version>7.11.0.1</Version>
+    <Version>7.11.0.90</Version>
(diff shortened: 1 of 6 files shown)


rmcrackan:
How will this play with the if-series template tags?

Mbucari:
I don't understand how the <if series tag is supposed to work<if seties>.

I used that tag in my folder settings like so:
<title short> [<id>] <if series><series><-if series>

And this is the result:
C:​\Users​\mbuca​\Libation​\Books\A Study in Scarlet [123456789] _if series_Sherlock Holmes_-if series_\A Study in Scarlet_ A Sherlock Holmes Novel [123456789].m4b

It doesn't seem to work as intended

Mbucari:
It works as you'd expect.  The parent folder is a series, so the tags inside the `<if series>` condition are applied to it as well as to the episode's folder inside of the parent.

Issue #186, opened 2022-01-11
FEATURE REQUEST: Ability to set tag naming paterns the same way I can set file/folder patterns


oucil:
It's great that I can use variables for the file naming conventions, but it'd be really nice to be able to set patterns for the actual meta tagging as well.  For instance, right now, all tracks are using the book title even when there is a specific chapter name available.  Also, the track numbers aren't set at all.  

At the moment I can sort of use Mp3Tag to bulk update the files and tags, but even then, books that are in a series and include book numbers, or that have a secondary title are different than books without so I can't do a full bulk change to all at once, I have to update each book individually.

Lastly, it would be great if we could set the Audible ID to be stored in a tag, i.e. "Comment", rather than the file name.

Thanks!

cadric:
I was just about to open a feature request on something like this, but instead I will comment here :)

Tags I would like to see used are:
Title, Artist, Album Artist, Album, Discnumber, Year, Genre, Publisher, Comment, and Track

You are allready doing many of the things I list below, but some tags are still missing. 
Also it seems the tags allready have the correct chapter "title" and track numbers are set correctly in tags.
The tags I see missing or wrong is: 

- Album Artist - Should be blank as the writers name is allready in Artist
- Narrator - (See *1) Should be the name of the narrator
- Album - Should only be the title of the audiobook and not include series title  or book number
- Series - A tag for the series. (See *2) below
- Discnumber - Book number
- Publisher - Which recording studio made the audiobook


Using tags I would like to tag a book like [Apocalypse: Generic System](https://www.audible.com/pd/Apocalypse-Generic-System-Audiobook/1039404472) as this.

```
M4B Tag                         Audible "book info"                 Audible "tag"
Title                           Chapter 1: May Cause Drowsiness   - Chapter name or number
Artist                          Macronomicon                      - By:
Narrator                        Steve Campbell                    - Narrated by:
Album                           Apocalypse: Generic System        - Book title
*2                              Systems of the Apocalypse         - Series:   
Discnumber                      1                                 - Series:, book number
Year                            2021                              - Release date:
Genre                           Science Fiction                   - Audible category
Publisher                       Podium Audio                      - Publisher
Comment                         Asin: 1039404472                  - Product_ID/ASIN
Track                           Chaper Number                     - Chapter numbers should be set in track
```
*1 There is support for a [narrator](https://help.mp3tag.de/main_tags.html) tag in mp4
[...]

Issue #183, opened 2022-01-10
Chapter titles chopped after 50 characters


Kazaa3:
I have some audiobooks that are 20- to 40-hour collections by one author that are impractical to use, but cheap to buy. Which is why I split them up into individual books. The first step I take is to save the audiobook in separated chapters. This is possible with Libation (set Split my books into multiple files by chapter checkbox). 

Unfortunately all chapter titles are chopped off after 50 characters. I have shortened the templet to <ch# 0> - <ch title>, but this does not change anything (and choosing a folder that has a short save path). There seems to be a fixed string limit of 50 characters.

Would it be possible to at least store the full unabridged chapter title in <title> tag, or to use an additional parameter to specify to how many digits the tag is shortened (e.g.: <ch title, 256>) to shorten to 256 characters.

Best regards and thanks for this great piece of software.

rmcrackan:
@Mbucari : When splitting by chapter, does AAXClean store full chapter names in file metatags?

Mbucari:
Yes. ChapterInfo.Title is written into the chapter tags.

Mbucari:
@rmcrackan I actually got around to testing this, and I can indeed writer custom chapter titles longer than 50 characters.  I just tested with a 150 character string and it writes no problem.

rmcrackan:
@Mbucari this one's on me. The original poster means 50 char limit for file name. This is true. Especially with custom file names, the logic for handing file and folder names is really complex. To have a prayer of getting it right, I limit the file name to 50.

rmcrackan:
@Mbucari I thought I saw something in this weekend's big batches about 50 character stuff. Was that about #183 ? Or am I just remembering wrong?

Mbucari:
I totally missed this, but no, I haven't touched anything relating to file naming or trimming chapter titles.

Mbucari:
@rmcrackan
What exactly is the issue here? Is it that we are running up against the maximum file name length in windows, in which case nothing can be done and this issue should be closed? Or is it that these file names could be longer, but something within libation is preventing it?

rmcrackan:
[...]

PR #277, opened 2022-06-20
Issues #183 and #186, and a lot of other little things

Mbucari:
- #186
  - Added `FileManager.MetadataNamingTemplate` and `LibationFileManager.ChapterTitleTemplate` to handle metadata template naming.  They are not bound by the same filename rules as `FileNamingTemplate` and `FileTemplate`
  - Created a second `MultiConvertFileProperties` delegate for getting chapter titles in `AaxcDownloadMultiConverter`
  - Added a user setting in "Audio Settings" to change the template.
- #183
  - As I discussed in that issue, I made a `LongPath` class that takes care of adding the long path prefix. It works with file and UNC paths, and playes nicely with the rest of Libation.
  - Reworked `FileNamingTemplate.GetFilePath`. I now builds a list of filename parts, replacing replacing template parameters along the way. After the list is made, it removes 1 char from the end of the longest part until the filename == 255 chars.
- Added unicode analog repklacement chars for illegal path chars.
  - : = ꞉
  - * = ✱
  - ? = ？
  - < = ＜
  - > = ＞
-Housekeeping
  - Merged `Streamable` and `Processable` classes. `Streamable` was only used by ther file downloader, and after removing it, all `Streamable` objects are also `Processable`.
  - Split `SettingsDialog` into partial classes like you did with `Form1`
- Bug Fixes
  - `ConvertToMp3` now works properly with multipart mp4 files
  - `ConvertToMp3` now uses the mp3 encoder settings from the settings instead of AAXClean's defaults.
- Updated all Libation tests


Mbucari:
One more:

Created `AaxDecrypter.IDownloadOptions` interface and moved `AaxDecrypter.DownloadOptions` to `FileLiberator.DownloadOptions`, implementing `IDownloadOptions`.  This allowed me to add `GetMultipartFileName()` and `GetMultipartTitleName()` methods to `IDownloadOptions` so  we no longer have to pass delegates to AaxDecrypter.

Source/AaxDecrypter/AaxDecrypter.csproj
```diff
@@ -5,8 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="AAXClean" Version="0.4.7" />
-    <PackageReference Include="AAXClean.Codecs" Version="0.2.7" />
+    <PackageReference Include="AAXClean.Codecs" Version="0.2.9" />
```
rmcrackan, line 8: Cool. How did you manage to deploy these together?

Mbucari: I didn't, but I never needed to include AAXClean independently anyway since it's already referenced by AAXClean.codecs.

rmcrackan:
This looks cool. I know I sound like a nagging broken record (and maybe I am one) -- but messing with the paths inside on config can break the world. Test lots, incl. fresh install with no prev config files. A recent version (maybe 8.0.1) slipped past us where it wouldn't let a brand new user install the app. That one was completely my fault btw, just raising the awareness that it went out uncaught so now it's fresh on my mind. Ironically, that one was when I added universal error handling ... sigh

Source/AaxDecrypter/AaxDecrypter.csproj
rmcrackan: Forehead smack. Of course. I'm so not on top of my game today

rmcrackan:
Once you verify that you've tested new, no prev existing config, install -- I'm ready to pull the trigger. This is a nice upgrade.

rmcrackan:
> I added some more comments

Gracias. I read them now and I'll leave them there for later also.

Mbucari:
> leave them there for later also.

All of those comments are for both of us, because none of this stuff is intuitive.

PR #280, opened 2022-06-20
Add support for Audible's new  hierarchical chapters

Mbucari:
What convenient timing! I'm not actually mad though.  Having a hierarchy will make for better chapter title names.


Mbucari:
I also fixed #281 
Previous version only affected the file name and ignored all directories :confounded:

PR #282, opened 2022-06-21
Fixed rare bug that would hang if an error occured while downloading

Mbucari:
This took HOURS to find.  Every now and then, a book in the queue would just stop downloading with the CPU core maxed.  Libation was still responsive until you tried to cancel the hung book, then it would lock up the whole process. This would happen if there was an error in the download loop and the file didn't finish downloading. AAXClean would keep calling Read until it got what it needed, but it never checked if Read returned 0 bytes to signal the end of the stream, so it would hang in an infinite loop.

One of the main causes for errors in the download loop turned out to be an IO while trying to write the JsonFilePersister file. I wrapped the Update event invocation in a try/catch block and just log the error.  There's no reason to stop downloading if one of the updates fails to write to disk.

I also discovered some bad logic in the routine that waits for data to be downloaded. ~God know what damage that caused, but it probably caused some silent errors in decrypted files. And it's been around for a long time...~

Never mind. I just realized that this would not have been a problem because the AAXClean bug that I just fixed would have actually counteracted the bad WaitToPosition logic. Phew!

### DIFF
--- a/Source/AaxDecrypter/AaxDecrypter.csproj
+++ b/Source/AaxDecrypter/AaxDecrypter.csproj
@@ -5,7 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="AAXClean.Codecs" Version="0.2.9" />
+    <PackageReference Include="AAXClean.Codecs" Version="0.2.10" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


Mbucari:
The tl;dr for your release is: Fixed occasional hang and rare file corruption bugs in process queue

Issue #294, opened 2022-06-23
Crashing when trying to download book


Ninjalord-25:
when I go into download a book, it starts up like normal, click the red traffic light to start the process, it shows up in the download bar and then briefly flashes to the time to wait for completion, then the program promptly closes without having done a thing
[Log202206_003.log](https://github.com/rmcrackan/Libation/files/8970134/Log202206_003.log)

Foardskin:
I am having the same problem, crashing as soon as the download starts in the process queue. But that folder you mentioned doesn't exist on my computer

Mbucari:
@Foardskin Please attach you log file

Foardskin:
I don't know how to find one of those sorry. 
the log in libation, on the the tab next to the process queue is blank.

Mbucari:
In Libation top tool menu, click Settings > Settings...

The Setings window will open up.  On that first page you'll see a button called "Open log Folder". Click it. You'll see at least one file in there whose name begins with "Log". Upload all such files here (just drag-drop the files into the github text editor.

Mbucari:
@Foardskin That log file is very old. Are there any other log files that don't begin with an underscore (_)?

Foardskin:
[Log202206.log](https://github.com/rmcrackan/Libation/files/8970674/Log202206.log)

here we go, thank you

Foardskin:
I found the downloadinprogress folder. AppData was a hidden folder in windows 10.

unfortunately I'm getting the same issue with libation closing on download start

[Log202206.log](https://github.com/rmcrackan/Libation/files/8970754/Log202206.log)
t

Mbucari:
@rmcrackan I think the only thing we can do for not is to add more logging and error handling.  There's just not enough info in these reports to identify a problem.

I'm working on that now and will submit a PR tonight for you to review tomorrow.

Ninjalord-25:
> So sorry about the crashing. Thanks for the log. Please delete the files in `C:\Users\chaot\AppData\Local\Temp\Libation\DownloadsInProgress` and try again. If you're still crashing, please attach the new log file.

did that and it happened the same as before, here's the new log file
[Log202206_003.log](https://github.com/rmcrackan/Libation/files/8971714/Log202206_003.log)

rmcrackan:
Thank you for your patience. See if [version 8.1.2](https://github.com/rmcrackan/Libation/releases/latest) solves your issue. If it doesn't, it should at least give us much better logs to figure this out.

[...]

Issue #292, opened 2022-06-22
Crash when attempting to download.


fantoams18:
[Log202206_001.log](https://github.com/rmcrackan/Libation/files/8961605/Log202206_001.log)

Scans fine. Sees new titles. Click to download and it appears on the right. Calculates time. Crashes out.

Tried removing and re-adding account info.

rmcrackan:
So sorry about the crashing. Thanks for the log. Please delete the files in `C:\Users\Phils\Documents\Libation\DownloadsInProgress` and try again. If you're still crashing, please attach the new log file.

rmcrackan:
Thank you for your patience. See if [version 8.1.2](https://github.com/rmcrackan/Libation/releases/latest) solves your issue. If it doesn't, it should at least give us much better logs to figure this out.

rmcrackan:
Duplicate of #294

rmcrackan:
Since this issue #294 are the same bug, I'm going to consolidate these so I only have to track them in one place. If you reply in either place though, I'll see you. Sorry for this bug. I'm not sure what I did wrong but we're on the case until this is resolved.

Issue #285, opened 2022-06-21
Feature request: pseudo-characters which preserve appearance as an option


anthonynogales:
I noticed last night after updating to version 8.1.0.2 that downloading an audiobook with colons in the name results in the folder and file name showing colons as well. I reported this on a Reddit [thread](https://www.reddit.com/r/audible/comments/vbi0f5/comment/id5f2zz/?utm_source=reddit&utm_medium=web2x&context=3) by darchangel. 

`..\Drew Karpyshyn - Star Wars꞉ The Old Republic꞉ Revan\Star Wars꞉ The Old Republic꞉ Revan [B005ZT0X1C].m4b
`

Mbucari:
It's a feature, not a bug. 

Illegal path characters are replaced with Unicode analogs. Replacements are as follows:

- : → [꞉](https://unicodes.jessetane.com/%EA%9E%89)
- \* → [✱](https://unicodes.jessetane.com/%E2%9C%B1)
- ? → [？](https://unicodes.jessetane.com/%EF%BC%9F)
- \< →[＜](https://unicodes.jessetane.com/%EF%BC%9C)
- \> → [＞](https://unicodes.jessetane.com/%EF%BC%9E)
- " → [“](https://unicodes.jessetane.com/%E2%80%9C), [”](https://unicodes.jessetane.com/%E2%80%9D), or [＂](https://unicodes.jessetane.com/%EF%BC%82)using some logic to detect open and closed quotes.

Mbucari:
It's on forever. I thought everyone would appreciate it so I didn't make it optional.  @rmcrackan @anthonynogales Should this be optional?

anthonynogales:
I'm at the very beginning of organizing my audiobook library so it makes no difference to me. The only reason I'd rename these would be if these characters didn't play well with Plex or Audiobookshelf BUT I don't expect there will be any issues. In fact, I think it's pretty awesome! Once I finish my testing I'll probably utilize this across the board since (novice that I am) I didn't realize I could replace the illegal characters with Unicode analogs. My default has been to replace ":" with " - ". 

Playing devil's advocate, the only people I see being peeved are the OCD folks who have already organized thousands of audiobooks and would have to rename the folders\files after they're downloaded or else update their existing folders to use the analogs. They're probably a minority though and if they're that OCD they could probably script a solution.

Mbucari:
> if these characters didn't play well with Plex or Audiobookshelf 

[...]

PR #295, opened 2022-06-23
Optional illegal character replacement and more error handling/logging

Mbucari:
The biggie in here is optional illegal character replacements as discussed in #285 

- `FileManager.ReplacementCharacters` has 3 public methods:
  - `ContainsInvalid` accepts a string and returns true if any chars in that string are invalid path characters.
  - `ReplaceInvalidPathChars` accepts a path string and will replace all illegal characters with the user-defined replacements.
  - `ReplaceInvalidFilenameChars` accepts a filename string and will replace all illegal characters with the user-defined replacements. This works identically to `ReplaceInvalidPathChars`, with the addition of slashes as invalid characters.
- Defaults can be changed in Settings > Download/Decrypt > Edit how illegal filename characters are replaced. there are 3 defaults:
  - Default: Existing unicode replacements
  - LoFi Default: only ANSI chars are used.
  - Barebones: every invalid char is replaced with an underscore.
-  Updated tests

Add more logging and error handling to `AaxDecrypter` and `ProcessQueueControl`

This won't fix the problems in #292 and #294, but at least we'll have a better idea of what's broken and the app won't crash. We've got lots of log files that are crashing on download. Some of them are crashing in Step 2: Download and Decrypt, which means that the process is at least making it to the download stage.  Others are crashing before execution even reaches inside the Processable, so that probably means the exception is happening in the Process Queue.  I'm really curious to see what's causing these.


rmcrackan:
I'd love to comb through this in detail but I'm just going to have to trust you. Too big of code changes and way too much coming through today to allow for micromanaging. And clearly you know your stuff and have earned the leeway

Issue #201, opened 2022-02-27
Error Importing Library


rossdargan:
[Log202202.log](https://github.com/rmcrackan/Libation/files/8148703/Log202202.log)

I get this error while trying to do an import on my account:

Bad Data.

   at Internal.NativeCrypto.CapiHelper.ToKeyBlob(RSAParameters rsaParameters)
   at System.Security.Cryptography.RSACryptoServiceProvider.ImportParameters(RSAParameters parameters)
   at AudibleApi.Cryptography.CreateRsaProviderFromPrivateKey(String privateKey)
   at AudibleApi.Cryptography.signSha256(String private_key, String dataString)
   at AudibleApi.Cryptography.CalculateSignature(HttpRequestMessage request, DateTime dateTime, AdpToken adpToken, PrivateKey privateKey)
   at AudibleApi.Cryptography.SignRequest(HttpRequestMessage request, DateTime dateTime, AdpToken adpToken, PrivateKey privateKey)
   at AudibleApi.Api.signRequestAsync(HttpRequestMessage request)
   at AudibleApi.Api.AdHocAuthenticatedGetAsync(HttpRequestMessage request, IHttpClientActions client)
   at AudibleApi.Api.AdHocAuthenticatedGetAsync(String requestUri, IHttpClientActions client)
   at AudibleApi.Api.getLibraryAsync(String parameters)
   at AudibleApi.Api.GetLibraryAsync(LibraryOptions libraryOptions)
   at AudibleApi.Api.GetAllLibraryItemsAsync(ResponseGroupOptions responseGroups)
   at AudibleUtilities.ApiExtended.getItemsAsync(ResponseGroupOptions responseGroups, Boolean importEpisodes) in C:\Dropbox\DinahsFolder\coding\_NET\Visual Studio 2022\Libation\AudibleUtilities\ApiExtended.cs:line 131
   at Polly.Retry.AsyncRetryEngine.ImplementationAsync[TResult](Func`3 action, Context context, CancellationToken cancellationToken, ExceptionPredicates shouldRetryExceptionPredicates, ResultPredicates`1 shouldRetryResultPredicates, Func`5 onRetryAsync, Int32 permittedRetryCount, IEnumerable`1 sleepDurationsEnumerable, Func`4 sleepDurationProvider, Boolean continueOnCapturedContext)
[...]

PR #300, opened 2022-06-25
Option to combine Opening/End Credits chapters and other changes

Mbucari:
- Added a new user audio setting to merge Opening and End Credits chapters with the following/preceding chapter. I also added more chapter tests.
- Added audio bitrate to the database.  I added it as an enum type, `AudioFormatEnum`, that stores bitrate, sample rate, and channel configuration.  `AudioFormatEnum` is internal to DataLayer, and is exposed to referencing assemblies inside `AudioFormat `, which displays the info in a much more friendly way.  `AudioFormatEnum` enum value names match the `AvailableCodec.EnhancedCodec` names, allowing them to be parsed from strings.  The bitrate info is displayed in the `BookDetailsDialog`, but nowhere else yet.
- Configured libation to build via publish.
  -  To do this, I had to get rid of MSBUMP.  It would work, but after cleaning the solution `dotnet publish` would fail.  I replaced it with a dotnet tool: [frogvall.dotnetbumpversion](https://www.nuget.org/packages/Frogvall.DotnetBumpVersion/).  A pre-bild command will increment the build version, but only for debug builds. That way you can publish all 3 Libation exe files without AppScaffolding's build being incremented each time.
  - Added FolderPublish config files for all 3 libation exe's.
  - Added publish.ps1to the Source dir.  To publish, all you have to do is open the dev console and type `.\publish`.  The script will build all files and zip them to \\Source\\bin\\Libation.[ver].zip
    - NOTE: to run powershell scripts you need to enable script execution.  Run this command to enable: 
    - `Set-ExecutionPolicy -Scope CurrentUser Unrestricted`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -184,7 +184,7 @@ publish/
 *.azurePubxml
 # Note: Comment the next line if you want to checkin your web deploy settings,
 # but database connection strings (with potential passwords) will be unencrypted
-*.pubxml
+#*.pubxml
 *.publishproj
 
 # Microsoft Azure Web App publish settings. Comment the next line if you want to
--- a/Source/.config/dotnet-tools.json
+++ b/Source/.config/dotnet-tools.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "frogvall.dotnetbumpversion": {
+      "version": "3.0.1",
+      "commands": [
+        "bump-version"
+      ]
+    }
+  }
+}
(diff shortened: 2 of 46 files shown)


Mbucari:
I'm working on #301

Mbucari:
I've fixed #201, but it does require that change to AudibleApi.Common in my email.

rmcrackan:
the emailed change to audibleapi.common/audibleapi is now live. AudibleUtiliies is updated

Mbucari:
Updated package reference

PR #312, opened 2022-07-17
Avalonia Beta release

Mbucari:
Here's what I have so far.  You already saw the main form, which includes the display grid, menu items, status strip, and proces queue control.

I also converted the following dialogs:

- AccountsDialog
- BookDetailsDialog
- EditQuickFilters
- LiberatedStatusBatchDialog
- MessageBoxAlertAdminDialog
- ScanAccountsDialog
- SearchSyntaxDialog
- TagsBatchDialog
- DescriptionDisplay
- ImageDisplay

The dialogs that still need to be converted are:

- _2faCodeDialog
- ApprovalNeededDialog
- CaptchaDialog
- LoginCallbackDialog
- LoginChoiceEagerDialog
- LoginExternalDialog
- MfaDialog
- EditReplacementChars
- EditTemplateDialog
- LibationFilesDialog
- SettingsDialog
- SetupDialog

I changed how BookDetailsDialog works.  Once open, you can continue to click "Edit Tags" button for other books and they will be shown in the BookDetailsDialog without needing to reload the window.

I also added an option to opt-in to the beta in the settings dialog.

Other small changes that toughed the non-beta codebase:

- Reworked Program.cs to split up some loading tasks so I could load Avalonia in parallel.
- Fixed the flicker on the cover image display (both winforms and avalonia)
- Fixed a potential crash condition on product grid column reordering RemovableGridEntry.datasource
- Fixed a crossthread error in Form1.BackupCounts (I dont understand why UI invokations are fine 99% of the time, but the other 1% they're somehow on a different thread.)


rmcrackan:
> I changed how BookDetailsDialog works. Once open, you can continue to click "Edit Tags" button for other books and they will be shown in the BookDetailsDialog without needing to reload the window.

How does that affect saving -- 1st opened book vs last opened book?

> I also added an option to opt-in to the beta in the settings dialog.

Thanks, that will be crucial

> I dont understand why UI invokations are fine 99% of the time, but the other 1% they're somehow on a different thread.

Dude, you're telling me! As if there's some scenario where we'd actually want that f\*\*king cross thread exception. The framework should resolve this by now.

rmcrackan:
Minor update issue

Add a book to your audible library. Import > scan. New book is added to grid

* winforms: 'visible' counts (top and bottom) do change
* avalonia: 'visible' counts (top and bottom) do not change

Same issue when a book no longer matched the filter after a scan. It will get automatically removed from the grid but the visible counts refresh in winforms and do not refresh in avalonia.

Mbucari:
> How does that affect saving

Saving is done by the BookDetailsDialog, either when you click "Save" or when you press enter. Clicking "Save" just saves changes to the db, whereas pressing enter saves and closes the form.  See BookDetailsDialog.axaml.cs::SaveButton_Clicked()

> counts (top and bottom) do not change

Fixed

I also made a slight change to AccountsDialog wrt adding a new row, and I've pushed my work-in-progress SettingsDialog.  It's 1/4 done, but its not wired up anywhere.

Re x-thread. I thought about this a little more, and I *think* this is only a problem in async event handlers. The task scheduler tries to run on the main thread, but sometimes runs on a different thread (depending on current load and the phase of the moon). I don't think there's any guarantee that any code within an async event handler will execute on the main thread, so any interaction with the UI within those handlers should be invoked on the UI thread as a matter of course.

rmcrackan:
Looks good. Thanks!

The top menu feels unnecessarily tall but I'm not going to hold up the PR for minutia like that.

rmcrackan:
Fwew, glad I saw this before completing the deploy. I'll hold off until you're ready. No rush. No one else even knows this is coming

PR #321, opened 2022-07-24
Linux Beta

Mbucari:
Notable changes in this PR are:
- As I mentioned in my email, I moved all Avalonia stuff into LibationAvalonia project. I disabled the "Use Beta" checkbox in the winforms settings menu and removed it from the Avalonia settings menu.
- All dialogs have been converted except for EditReplacementChars, so that button in the settongs dialog is not enabled.
- I had to remove the update function because AutoUpdater.NET only works for WinForms or WPF application projects.
- Publish profiles for both win64 and linux64 are added for both LibationAvalonia and LibationCli.
- I added instructions [here](https://github.com/Mbucari/Libation/blob/master/Source/LibationAvalonia/README.md) for users to install and run Libation on Ubuntu

I found 1 backslash and replaced it. The setup routines have been fixed and thoroughly tested. I've tested both Libation and LibationCli on a fresh Ubuntu install using the instructions I wrote. It all works. I think it's ready for beta release. 

EDIT* I changed the publish profile to standalone, and now users do not need to install the dotnet runtime on their machines.


Mbucari:
Converted this to a draft because I'm pretty sure we will not have a good time releasing builds.  We'll need to make a script to dl and build from source.

Source/LibationFileManager/Configuration.cs
```diff
@@ -345,7 +345,7 @@ private void setTemplate(string settingName, Templates templ, string newValue)
 		#endregion
 
 		#region known directories
-		public static string AppDir_Relative => $@".\{LIBATION_FILES_KEY}";
```
rmcrackan, line 348: Good find

rmcrackan:
> Converted this to a draft because I'm pretty sure we will not have a good time releasing builds. We'll need to make a script to dl and build from source.

I don't follow. What's stopping us from releasing the old way?

Mbucari:
> I don't follow. What's stopping us from releasing the old way?
I built a release using visual dotnet publish, targeting linux-x64.  I copied those bins to my Ubuntu 22 VM and everything worked great! This morning I copied that same bin package to a Ubuntu 20 VM, and it wouldn't run. I don't know what dotnet considers "linux-x64", but it's obviously not universal.  Unfortunately I've found next to no info on what distributionstthose bins are supposed to work on. It just says "Linux", which is terribly unhelpful.

rmcrackan:
Damn. Everything I've found today says that ubuntu 20+ has been supported for a while. I guess since our goal is x-platform, this is a show stopper

Mbucari:
OK, the problem was with standalone builds.  They must be for very specific linux kernels, but they don't say which ones.

So we'll have to redistribute it as a framework-dependent build and users will need to install the runtime.

I've updated the publish profiles and added more instructions for installing.

https://github.com/Mbucari/Libation/blob/master/Source/LibationAvalonia/README.md

atylmo:
Did someone say Linux build? :D

Just tried the Test 5 release, seems to be fine here on Ubuntu 21.10.

Thanks to both of you for your hard work!

Mbucari:
> Just tried the Test 5 release, seems to be fine here on Ubuntu 21.10.

Woohoo! I only made that release for my own testing purposes and didn't intend for anyone else to use it, but I'm thrilled it worked for you?  Did it add Libation to your applications list with the icon?

rmcrackan:
> Thanks to both of you for your hard work!

You're quite welcome. I'm not shy about taking credit for the vision, the years of work, and the literally thousands of hours I put into Libation before anyone took notice. A lot of though went into this and I'm very proud of this program.

In the last 1 to 1 1/2 years though, depending on the month in question, @Mbucari 's contributions have been somewhere on the spectrum of 'small meaningful improvements' to 'damn near every new change.' And this cross platform effort has been 1,000% Mbucari . If anything, I've had brief moments of input -- and even that's probably giving myself too much credit. My man's been a *beast*!

rmcrackan:
@Mbucari I see this is no longer in `draft`. I'm ready to merge this if you are. Those new linux instructions look really good (says the guy who doesn't use linux and has no idea what he's talking about). They appear to not be linked to from other READMEs though. Is that intentional for now?

Mbucari:
> I'm ready to merge this if you are. 
Go fo it. I just updated the install script to make LibationCli executable and to add a sym link for it.

> They appear to not be linked to from other READMEs though

I just added it.

> Those new linux instructions look really good (says the guy who doesn't use linux and has no idea what he's talking about). 

Pfft. I'm faking it over here. I just wrote my first bash anything today, and that is it lol.

atylmo:
> > Just tried the Test 5 release, seems to be fine here on Ubuntu 21.10.
> 
> Woohoo! I only made that release for my own testing purposes and didn't intend for anyone else to use it, but I'm thrilled it worked for you! Did it add Libation to your applications list with the icon?

Oops, sorry! I was being lazy and didn't want to build it and just happened to notice the package :laughing:.

It didn't but that's because I unzipped and ran it directly. My bad, again. I just installed it with your script and it worked perfectly, icon and all! Great job on the scripting!

rmcrackan:
Southern summer thunder storm keeps making my power blink. Since I still haven’t gotten around to putting my router on a UPS, it means my wifi keeps going away. Typing this on my phone. Tomorrow I have pointless work meetings all day so I should have plenty of time to do a final review while I pretend I’m listening to those.

rmcrackan:
I guess this is still the best place to keep discussing before release. I have several questions. Instead of vomiting them all at once like I usually do, let's go one at a time.

If we're going to maintain both UIs anyway, there no reason *not* to let windows people have access to avalonia if they want to. I agree that the beta flag isn't the right way to do it anymore. New setting? I'm especially imagining people who use windows and linux who would appreciate it looking the same on both.

Mbucari:
I don't think you need a new setting, you just need to make and post a Windows build of avalonia. Do it the same way you already do it for winforms. We'll end up with 3 bin zips

Libation.#.#.#-win.zip
Libation.#.#.#-win-beta.zip
Libation.#.#.#-linux-beta.zip

A better way would be to have different version numbers for the avalonia betas, but since the version number is housed in app scaffolding which is shared by all packages, that's not possible.

rmcrackan:
Interesting. Sure, I guess it could be separate builds instead of hot-swappable. This would make smaller downloads. Win = winform, hangover, CLI. Win-beta (needs a better name) = avalonia, hangover, CLI. Linux-beta = avalonia, CLI

Next item:

> I had to remove the update function because AutoUpdater.NET

I didn't want to hold up this enormous PR just for this. We can't release with this part as-is though. It's enormously popular. More importantly though is that people rely on it now. Minimally winforms needs a way to autoupdate. Ideally avalonia needs a way. Super ideal would also be to add a CLI action. I'm not married to autoupdater.net specifically; it's just easy to use.

Mbucari:
I've been thinking about the updater, and I've come to the conclusion that this is a hard problem lol.


I sketched out a potential solution.  Take a look and tell me what you think.

```C#
public static void Main2(string[] args)
{
	//Don't look at this if block until you've read and understand what comes after it.
	//Then come back and take a look. 

	if (args?.Length == 2 && int.TryParse(args[0], out var processId) && Directory.Exists(args[1]))
	{
		//This Libation instance is running in a temp location for the purpose of updating

		var programFilesDir = args[1];

		//Wait for the calling Libation process to exit
		var lingeringProc = Process.GetProcesses().SingleOrDefault(p => p.Id == processId && !p.HasExited);
		if (lingeringProc is not null)
			lingeringProc.WaitForExit();

		var thisUpdateDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);

		//Copy all files from this temp directory to the program files directory
		foreach (var file in Directory.EnumerateFiles(thisUpdateDir))
		{
			var relFile = Path.GetRelativePath(thisUpdateDir, file);

			var dest = Path.Combine(programFilesDir, relFile);

			if (!Directory.Exists(Path.GetDirectoryName(dest)))
				Directory.CreateDirectory(dest);

			File.Copy(file, dest);
		}

		//All update files have been coppied. Start Libation in the program files directory, then exit.
		//Let the system take care of cleaning up these temp files.

		var libationExe = Path.Combine(programFilesDir, "Libation") + (islinux ? "" : ".exe");
		StartProcess(libationExe);
		return;
	}

	//This is where the existing Libation program starts
	var config = AppScaffolding.LibationScaffolding.RunPreConfigMigrations();
	var proc = Process.GetCurrentProcess();

	if (!string.IsNullOrEmpty(config?.UpdateFile) && File.Exists(config.UpdateFile))
	{
		//A previous Libation instance downloaded an update zip file, so we run the installer now.

		var updateExtractionDir = Path.Combine(Path.GetTempPath(), "libationupdate");

		ExtractUpdateFilesToTempLocation(updateExtractionDir);

		var libationTempExe = Path.Combine(updateExtractionDir, "Libation") + (islinux ? "" : ".exe");
		var thisProcID = Process.GetCurrentProcess().Id.ToString();
		var libationProgramFilesDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);

		//Run the Libation executable in the temp location with this instance's
		//process ID and directory as arguments, then exit.
		StartProcessWithArgs(libationTempExe, thisProcID, libationProgramFilesDir);

		return;
	}
}
```

Mbucari:
This would be great:
https://github.com/NetSparkleUpdater/NetSparkle#installing-netsparkle

Except it appears it requires a setup exe (or msi) on windows systems.

rmcrackan:
Sorry for being so absent today; today's meetings weren't as hands-off as I'd hoped.

> I've come to the conclusion that this is a hard problem lol

Yes. Yes sir it is.

Although I couldn't respond today, I did lots of thinking. In order of preference:

* What's the blocker on using autoupdater.net? The site doesn't specifically list avalonia but it just looks like just .net code
* I opened a [ticket](https://github.com/ravibpatel/AutoUpdater.NET/issues/556) with autoupdater . Last time he responded within a few days. Maybe we'll get lucky and he can address whatever issue you found
* Even if we can't use the nuget, we can fork and tweak it to our needs. Again: almost all of it just looks like UI-agnostic .net
* If autoupdater is a no-go, RE NetSparkle "Except it appears it requires a setup exe (or msi) on windows systems." That's a deal breaker but I don't see that anywhere on that link. Unless I'm missing something, those links to InnoSetup, WiX, etc. look like guidance, not requirements.
* If none of this works, then we take it on the chin and go homegrown. I think there are other promising options though.

Mbucari:
> What's the blocker on using autoupdater.net?

It's a GUI built on winforms (or WPF).

> I don't see that anywhere on that link.

Right near the top:

Built-in supported update download types:

    Windows -- .exe, .msi, .msp
    macOS -- .tar, .tar.gz, .zip, .pkg, .dmg
    Linux -- .tar.gz, .deb, .rpm

rmcrackan:
> It's a GUI built on winforms (or WPF).

I'll look into it further. Surely this is just a tiny bit of UI around core components. We could probably even use whatever his wpf is for avalonia. This feels like less work than homegrown. I assume you discovered the incompatibility when you tried to use it with avalonia. Did it not compile?

> Built-in supported update download types:

I saw that many times and thought: yeah, we do have exe. It's only now noticing macos that they don't include zip as a windows option.

Mbucari:
> Did it not compile?

Nope and here's why (lines 6-8)

 https://github.com/ravibpatel/AutoUpdater.NET/blob/be241529c00b7f4b6f11bd298a9ecebf6bf0e063/AutoUpdater.NET/AutoUpdater.NET.csproj#L8

Converting that to Avalonia is non-trivial.  It has two pieces:  the update form and downloader, and the zip extractor.  They are both GUIs and would need to be ported to Avalonia.

IMHO, forking, replacing winforms with Avalonia, and repackaging for our use is more work and is more error-prone than using my upgrader program (rember that from way back?  It looks like I deleted the repo, but I'm sure I still have the source files somewhere). upgrader was basically the AutoUpdater.NET's ZipExtractor utility, except that you would pass the download URL and it would do the downloading.

Mbucari:
My opinion is we need to roll our own.  It does not need to be complicated, and most (if not all) of the work can be done within Libation. At a minimum, an audo-updater need to be able to:

- Be copied to a location outside of the program files directory
- Be executed by Libation
- Write to the program files directory.

The way I see this working on the GUIs is:

1. Libation starts up a normal and checks for an update in the background. If it finds one, it quietly downloads it to the temp folder and adds an entry in its settings that an update zip file exists on the system.
2. On the next launch, Libation:
2.1  sees it has a pending update file
2.2 Copies the updater exe to a temp location
2.3 launches the update program as admin/root with whatever arguments it needs to do its work
2.4 and exits.
3. The update program will wait for Libation to exit, then extract the zip contents into the program files folder.
4. Updater will then start Libation as a standard user then exit.

For CLI:

1. User runs `libationcli upgrade`
2. CLI downloads the zip
3. Then continue with steps 2.2-4 above

rmcrackan:
Works for me.

You won't begin to believe how much time I've wasted today. I thought all this time you'd removed it entirely, including from winforms and frankly I was pretty wtf. I saw your PR comment and searched the new code to see if you'd just commented it out. And it was entirely gone. Ok, he must have fully removed it -- that sucks. So I've been testing old versions of code via detached git branches and digging through which points we touch in autoupdater etc. All this because I'd previously done a directory search from within VS when I working on a different project. And when I searched Libation code, I'd forgotten to change search back to 'current solution'. So I was really searching for autoupdater within the directory of an unrelated project. \@\#\$\%\^\! **!!**

rmcrackan:
I'd still like updating to be an available feature for linux and CLI but it's not a priority like winform is -- since all of those users have come to rely on it. For now, do you want to push this into an issue/ticket and one of us can tackle it later?

I like your ideas above. I'm still on the fence about extending autoupdater. I see what you mean about non-trivial, but they also have really battle-tested code.

Mbucari:
> You won't begin to believe how much time I've wasted today

Whoops! 😂

> For now, do you want to push this into an issue/ticket and one of us can tackle it later?

Yes please. I'm anxious to start getting feedback from Linux users.

I had one more thought for adding updating to avalonia. What if we make a new libation dll devoted to updating.

rmcrackan:
> What if we make 2 new libation dlls devoted to updating...

That could work. Can you make a new issue-ticket and dump and/or link you ideas there? We can hash that out on another day. Like you said: we're anxious to push this out so users can quickly ignore us :)

How should I be doing publishes now so we can get our 3 builds?

* Libation.#.#.#-win.zip
* Libation.#.#.#-win-beta.zip (still needs a better name)
* Libation.#.#.#-linux-beta.zip

My usual way is: `devenv.exe "{Solution}" /build "Release|AnyCPU"`

My last attempt at the other way was: `dotnet publish -c Release {project}\{project}.csproj -p:PublishProfile={project}\Properties\PublishProfiles\FolderProfile.pubxml`

Mbucari:
## Libation.#.#.#-win
Do whatever you normally do.

## Libation.#.#.#-linux
```console
dotnet publish -c Release LibationAvalonia\LibationAvalonia.csproj -p:PublishProfile=LibationAvalonia\Properties\PublishProfiles\LinuxProfile.pubxml
dotnet publish -c Release LibationCli\LibationCli.csproj -p:PublishProfile=LibationCli\Properties\PublishProfiles\LinuxProfile.pubxml
```
Files are in bin-Avalonia\publish\linux-x64

## Libation.#.#.#-win-PrettyNewUIbutBuggyAndUnstable
```console
dotnet build -c Release -o bin-Avalonia\Release-win LibationAvalonia\LibationAvalonia.csproj
dotnet build -c Release -o bin-Avalonia\Release-win LibationCli\LibationCli.csproj
dotnet build -c Release -o bin-Avalonia\Release-win Hangover\Hangover.csproj
```
TYou must build Hangover last else it won't run for some reason.

Files are in bin-Avalonia\Release-win

Issue #324, opened 2022-07-27
Add App Updater to Avalonia Builds


Mbucari:
Continuing our discussion in https://github.com/rmcrackan/Libation/pull/321

Problem: Add automatic app updating to Libation and LibationCli without using winforms.

I mentioned the possibility of making 2 new libation dlls devoted to updating, one for Windows and one for linux, with the windows one using Autoupdater.net. I tried this out, but it's going to be a real problem because we would have to find and load all the windows forms assemblies manually. So that's a no-go in my book.

My next thought is to just copy Autoupdater.Net's ZipExtractor.exe and ship it with Libation. We would have to do the zip file downloading, but after that we just call zipextractor like so:
```console
zipextracter.exe --input %temp%\libationupdate.zip --output C:\path\to\Libation\ --executable C:\path\to\Libation\Libation.exe
```

Then we just make the same kind of thing fox Linux using a bash script, which can double as the auto updater for libationcli on linux.

All that's left is the libationcli updater for windows. Maybe just reuse zipextractor?

rmcrackan:
Another issue with autoupdater (I haven't tested yet so I don't know the answer) -- which zip will it grab? We're about to have 3.

Mbucari:
Octokit.Release.Assets contains all files published as part of the release.  As you said we'll have three, so you'll need to be very strict with your file naming and hard-code which suffix each applications looks for.

rmcrackan:
That's right. I remembered that being part of the autoupdater magic but we actually do it ourselves in appscaffolding. Once we decide on a naming convention for all 3 versions, GetLatestRelease can find the best match inside those Assets for the current version.

You were right about the current version of autoupdater. He responded. As it is now, sadly, winforms are too tightly coupled -- https://github.com/ravibpatel/AutoUpdater.NET/issues/556

rmcrackan:
@Mbucari Do I remember right that you completed this?

Mbucari:
> @Mbucari Do I remember right that you completed this?

Only for windows. Once we get this Deb package build Incorporated into the release workflows, I think we should just make a message box pop up for Linux and mac users that an update is available.

rmcrackan:
Sounds good. Do you want me to keep this open for that? Or close?

Mbucari:
Open. I'll add it after the deb build happens. What I'd like to do for linux is provide the shell commands to execute the update in a textbox the the user can just copy/paste.

rmcrackan:
[...]

PR #325, opened 2022-07-27
Add app update to Avalonia Build

Mbucari:
As I said in https://github.com/rmcrackan/Libation/issues/324, I added Autoupdater.Net's ZipExtractor.exe to the Avalonia project and added a download and launch routine to the windows Avalonia build.

After the main form is shown, Libation will check for any updates and, if one is found, silently download it in the background.  Once the download is completed, it asks the user if they want to update.  If yes, Libation launches ZipExtractor and the update happens just like it already does with the winfoms version.

I also added an overload to LibationScaffolding.GetLatestRelease() that accepts a regex pattern used to match Octokit.Release.Assets names.  This will be used in future updates now that you'll be releasing multiple binaries at once.  The regex is case insensitive.

The winforms build uses the following search pattern:

`Libation\.\d+\.\d+\.\d+-win.zip"`  LibationWinForms.Program.cs#173

...so you'll need to name the zip file so it'll match.

The windows Avalonia build uses the following search pattern:

`Libation\.\d+\.\d+\.\d+-win-avalonia.zip`   LibationAvalonia.Views.MainWindow.axaml.cs#90

...so you'll need to name the zip file so it'll match or change the pattern.

These changes will prevent Libation from updating to the wrong package in the future.

**But for these next few release with multiple files** where people's existing installations aren't expecting more than 1 asset, make sure you upload the Libation.#.#.#-win.zip **FIRST**.  The Octokit.Release.Assets list is in the same order as files appear in github.  So the first asset is first in the list and so on.  Existing Libation builds will return the first asset in the release, so as long as winforms zip file is first, users will be able to migrate with no problem.


rmcrackan:
**Very** nice output for one evening. Using his windows-only updater just for avalonia is a great quick solution for now but if we get around to writing a Linux one, then we'll be supporting 3 different upgrade paths. Although presumably we'd make a cross platform solution for linux which could then become our only version.

> The Octokit.Release.Assets list is in the same order as files appear in github

I wasn't sure if it was upload order or alphabetical. It was in my queue to test that this morning. If you're positive it's not alpha, then I'll skip my testing.

> for these next few release ... upload the Libation.#.#.#-win.zip FIRST

It's worse than that -- potentially indefinitely. (Realistically, for many months.) Once I build it into an automated process though we won't have to remember anyway.

RE regex: even though your regex is perfect for the current convention, it also locks us into this convention basically forever. I've gone through some alternatives in my head this morning but they each have their own flaws, so let's stick with this. (I guess these sentences were pointless...)

`-win-avalonia`

I'd like to not name this after the technology used. When we release this into the wild, non-tech people are (hopefully) going to discuss it with each other too. I want to brand it out of the gate so it's easy for anyone to talk about. Since the new version is white, sticking with the wine theme, my idea is "chardonnay". (We don't have to keep it white themed in the future; such is the glory of ambiguous branding.) The codebase, namespaces, etc can stay as is; no one cares but us. For anything user-facing, including the zip downloads, I'd like to use user-facing branding. So the zips would be this, or whatever branding we go with

* Libation\.\d+\.\d+\.\d+-win.zip (Libation\.\d+\.\d+\.\d+-win-classic.zip ?)
* Libation\.\d+\.\d+\.\d+-win-chardonnay.zip
* Libation\.\d+\.\d+\.\d+-linux-chardonnay.zip

Thoughts?

Mbucari:
>  Although presumably we'd make a cross platform solution for linux which could then become our only version.

That probably won't happen.  The ZipExtractor waits for Libation to exit and then as it extracts, it checks if any files being overwritten are locked.  Those 2 checks are platform-specific, so I don't think there is a single solution that can work for both win and linux. Hopefully soon we'll only need 2 upgrade tools instead of 3 when avalonia becomes the only windows release , but we will need one per platform.

> If you're positive it's not alpha, then I'll skip my testing.

I tested this on two random repos I found that had more than 1 release asset not in alpha order.  It was true for them, but you should probably test it too.  I'd hate top be wrong on this.

> RE regex: even though your regex is perfect for the current convention, it also locks us into this convention basically forever. 

You're right, and I'm not thrilled about that.

What we could do instead is add a release index file to your repo, something like:
```Json
{
   "Linux":"Libation.\\d+.\\d+.\\d+-linux-chardonnay",
   "Windows-Classic":"Libation.\\d+.\\d+.\\d+-win-classic.zip",
   "Windows-Avalonia":"Libation.\\d+.\\d+.\\d+-win-chardonnay"
}
```
The upgrader will first download this file (whose path in your repo must never change), and then use the correct pattern to search for releases. If you ever change the naming scheme, just update the json and all programs will still be able to auto update. It's pretty future proof.

> I'd like to use user-facing branding. Thoughts?

If we go with the json filename matching index, we could change the branding.  How about doing what Linux distros do and use a different name for each major version, with each succeeding name starting with the next letter of the alphabet. Something like:

So the first one might be Asti, followed by Barolo, Chardonnay, Drambuie? ...

rmcrackan:
Damn!! Glad I checked. Note the CreatedAt times.

alphabetized

Mbucari:
> Unless you feel strongly otherwise

The more I think about it, the more strongly I feel we should do it.  Regex solves the problem today, but today were are dealing with a problem created by a solution that was "good enough" yesterday.  I don't know what you or we may do in the future to break the regex, but I'd rather not have to be certain about tomorrow, today.

> Damn!! Glad I checked. Note the CreatedAt times.

Well, I guess a preceding underscore for the win classic release it is!

Mbucari:
OK, I added a release regex index to the repo root, named .releaseindex.json.

getLatestRelease() accepts a `ReleaseIdentifier` enum. It downloads .releaseindex.json from the repo, parses it, and selects the regex pattern for the specified `ReleaseIdentifier`. It then returns that release, if found.

After you merge, be sure to change LibationScaffolding.cs#368 from "Mbucari" to "rmcrackan"

Also, I don't know if you do this, but I had to use a github auth token to debug this so I wouldn't keep getting locked out by the api.

`gitHubClient.Credentials = new Octokit.Credentials("ghp_TOKEN");'

Mbucari:
> what are the odds that we can just create another publish file and also support mac?

Very good odds, but I have no way of testing it. I don't own a mac or know anyone who does.  Years ago I tried to install a bootleg macos image on a VM, but found I couldn't do unless I was running the VM on a mac.  Unless someone buys me a mac or can supply me with a mac VM, I can't help here.

rmcrackan:
> I guess a preceding underscore for the win classic release it is!

If only. Again: glad I checked. I added files in this order, saving in between each time: ZZZ, AAA, _ZZZ, _AAA . Final order: AAA, AAA, _AAA, _ZZZ

rmcrackan:
> gitHubClient.Credentials = new Octokit.Credentials("ghp_TOKEN");

Yup. I do this for tools I don't release. I have a home spun way to manage secrets locally way outside of git's reach.

rmcrackan:
Stupid alphabetizing. It's going to look dumb but I think I'm going to do this for a while `Classic-Libation.#.#.#-win-classic.zip`. And since we're not using begin of string regex anchors, it'll still match.

fyi, For consistency I'm also changing Linux => LinuxAvalonia in the regex file and in the ReleaseIdentifier enum

rmcrackan:
And we're live!! I thought I found one bug and it was user error :) You did a helluva job on this. Especially in such a short time. I'll publicize it very soon.

rmcrackan:
This was the first release in nearly a year that I did completely manually. It felt weird. I didn't want to hold us up with tweaking my release app.

After people have had a chance to pound on this version, my thoughts are: the next version can be v9.0 and we can bring linux out of beta. Maybe even have created a macos build. Neither of us can support mac, so that will be in beta. Hell, it might indefinitely remain an 'unofficial' build.

Mbucari:
> And we're live!! I thought I found one bug and it was user error :) You did a helluva job on this. 

Woohoo, and thanks! The more forms I converted, the more I came to appreciate mvvm. You just don't have to worry about events and triggering nearly as much as with winforms.

> the next version can be v9.0 and we can bring linux out of beta

Agreed. The sooner we can reunify the better. I need to figure out that damn character replacement dialog, then I want to focus on aesthetics. Right now it seems a bit too bright to me. Just way too much white. I'm using the light avalonia theme and there is a dark one, but that's a true dark theme. I want to start by toning down the whites in the light theme, and then maybe figure out how to do theme swapping so users can pick a light or dark theme. Then there's the matter of some weird proportions, like the menu being too tall as you noted before.

> Neither of us can support mac

I'm going to look more into getting a VM up and running on Windows host. I found a tutorial for it, and I'm pretty sure I can rustle up a Big Sur image from some dark corner of the internet 😉

But it would be great if some mac developer would come on board and help.

rmcrackan:
> it would be great if some mac developer would come on board and help.

Don't I know it. Think of any of the number of things which might have kept you from originally joining this project. Since you've been the only other constant member, that's exactly how close I was to still doing this solo.

PR #356, opened 2022-08-13
Add useCoverAsFolderIconCb setting to avalonia

Mbucari:
I thought this would make a great teaching opportunity, so I added the setting myself and explained the process below.
---
We need to add two properties to the SettingsDialog view model:
- `UseCoverAsFolderIcon`: holds the setting's boolean value.
- `UseCoverAsFolderIconText`: contains the control's display string ("Set cover art as the folder's icon. (Windows only)")

The `CheckBox` will reside in the Download/Decrypt tab, so we need to add those two properties to `DownloadDecryptSettings`

Inside SettingsDialog.axaml, I added the `CheckBox` control to the bottom of the Download/Decrypt `TabItem` . Note that the `TabItem` is organized inside a `Grid`. That `Grid` had 3 rows, but now we're adding a 4th row so we need to change its `RowDefinitions`.

I added the `CheckBox` control to the end of the `Grid` xml, and set it to display inside the new 4th row (`Grid.Row="3"`).
I bound `CheckBox.IsChecked` to the newly-added `UseCoverAsFolderIcon` with a TwoWay binding. TwoWay means that the `CheckBox` will read the value from the bound property, and it will set the bound property's value when the user changes it.

The `TextBlock` nested inside `CheckBox` is the CheckBoxs's label, and I bound its `Text` property to `UseCoverAsFolderIconText`

Lastly, since this is a Windows-Only setting, I added a boolean `IsWindows` property to `SettingsPages` and bound `CheckBox.IsVisible` to it. Now this `CheckBox` (and its child `TextBox` control) will only be visible on Windows versions.

### DIFF
--- a/Source/LibationAvalonia/Dialogs/SettingsDialog.axaml
+++ b/Source/LibationAvalonia/Dialogs/SettingsDialog.axaml
@@ -2,8 +2,8 @@
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-        mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="600"
-		MinWidth="800" MinHeight="600"
+        mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="620"
+		MinWidth="800" MinHeight="620"
         x:Class="LibationAvalonia.Dialogs.SettingsDialog"
         xmlns:controls="clr-namespace:LibationAvalonia.Controls"
         Title="Edit Settings"
@@ -206,7 +206,7 @@
 					BorderThickness="2"
 					BorderBrush="{DynamicResource DataGridGridLinesBrush}">
 
-					<Grid RowDefinitions="Auto,Auto,*">
+					<Grid RowDefinitions="Auto,Auto,Auto,*">
 						<controls:GroupBox
 							Grid.Row="0"
 							Margin="5"
@@ -367,6 +367,20 @@
 
 
 						</StackPanel>
+
+						<CheckBox
+							Grid.Row="3"
+							Margin="5"
+							VerticalAlignment="Top"
+							IsVisible="{Binding IsWindows}"
+							IsChecked="{Binding DownloadDecryptSettings.UseCoverAsFolderIcon, Mode=TwoWay}">
+
+							<TextBlock
+								TextWrapping="Wrap"
+								Text="{Binding DownloadDecryptSettings.UseCoverAsFolderIconText}" />
+
+						</CheckBox>
+
 					</Grid>
 				</Border>
 			</TabItem>
--- a/Source/LibationAvalonia/Dialogs/SettingsDialog.axaml.cs
+++ b/Source/LibationAvalonia/Dialogs/SettingsDialog.axaml.cs
@@ -107,6 +107,7 @@ namespace LibationAvalonia.Dialogs
 			LoadSettings(config);
 		}
 
+		public bool IsWindows => AppScaffolding.LibationScaffolding.ReleaseIdentifier is AppScaffolding.ReleaseIdentifier.WindowsAvalonia;
 		public ImportantSettings ImportantSettings { get; private set; }
 		public ImportSettings ImportSettings { get; private set; }
 		public DownloadDecryptSettings DownloadDecryptSettings { get; private set; }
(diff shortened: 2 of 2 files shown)


Mbucari:
Note: the whole settings dialog could be greatly simplified if the controls were bound directly to the configuration file. That way you wouldn't need intermediate storage in the view model. It would have the side effect of committing each setting immediately as it is changed, but that seems to be the approach used by most software these days.

rmcrackan:
> wouldn't need intermediate storage in the view model. It would have the side effect of committing each setting immediately as it is changed, but that seems to be the approach used by most software these days.

That's always the debate isn't it? Live or transactional. With live you don't need to remember to hit save. With transactional you can hit cancel when you don't like how things are going. This is what makes me nervous with the modern paradigm -- I feel less confident in poking around a lot of settings because there's no easy way to cancel. This came up recently in Zoom which has a bazillion settings -- I was less willing to make changes. If we do switch to the live paradigm I'd want this addressed. Especially because we also have a lot of settings and some, like templates, are particularly esoteric.

A hybrid might be to do live and also keep an undo history so users can revert. When you're updating live though, your stack for string fields gets flooded with transitional steps: C , C: , C:\\ , C:\\f , C:\\fo , C:\\foo , C:\\foo\\ , C:\\foo\\b , ... You could persist only when the form is closed, but that's really just the transactional model again.

Mbucari:
All string settings are edited in a different dialog. Users edit this naming templates in the EditTemplateDialog, then when they save the complete string is sent back to the SettingsDialog in one step. Same goes for the custom file path, which is only settable using the OpenFolderDialog. But if you did need to buffer any other properties (like the MP3 quality sliders), you could just keep the existing paradigm and only write on save.

It's up to you, but I'm not too bothered to change it at this point. All of the hard work of creating the setting dialog has already been done, so changing it to be live would only serve to make editing it easier in the future.

rmcrackan:
I'm going line by line through your explanation -- which is very good. Do you know why the checkbox and text do show up correctly on the compiled form but not in the designer? In fact, this is a lot of why I had a hard time -- the tooling is really not helpful which makes it too hard to find things. I was hoping that if I clicked on the xaml or the graphical element, that the corresponding piece would highlight. I don't see how we're going to maintain complex UIs with avalonia. The tooling is really subpar.

Mbucari:
> Do you know why the checkbox and text do show up correctly on the compiled form but not in the designer?

There could be a couple of things happening here. First of all, pure XML changes will show up immediately in the designer, but any changes in the code behind (or changes to the XML that reference changes in the code behind) will not show up until you manually build the assembly. For example, if you add a new property to the view model and then bind an XML element to that property without building the assembly, the designer won't know that property exists and will not update because it's in an error state. Get used to Ctrl + B.

It could also be because the design dimensions are different from the display dimensions. See d:DesignWidth="800" d:DesignHeight="600" at the top.

> I don't see how we're going to maintain complex UIs with avalonia. The tooling is really subpar.

I agree it's not great, but I have found it manageable. SettingsDialog is the most complicated UI element, but it's broken down into tab items which should make browsing XML easier. Just collapse every TabItem that you aren't currently working on.

I think all other controls are much more manageable because they are much better encapsulated, but I'll let you be the judge of that.

rmcrackan:
> checkbox and text do show up correctly on the compiled form but not in the designer?

It just occurred to me that this is because we set the visibility to something which it presumably can't check at design time.

> Ctrl + B

When I tried Ctrl + B, VS prompted for the next part of the chord. I assume this is something to do with building. Fortunately I already compulsively hit F6 to build solution

> SettingsDialog is the most complicated UI element

This is obvious in hindsight. At the time I was expecting workflow like winforms: drag-drop and element (before I realized height needed more), and add 3 lines in code behind for description, load value, and save value. Plus I assumed another line or 3 of whatever boilerplate mvvm might need.

> 
Do you know why the checkbox and text do show up correctly on the compiled form but not in the designer?

There could be a couple of things happening here. First of all, pure XML changes will show up immediately in the designer, but any changes in the code behind (or changes to the XML that reference changes in the code behind) will not show up until you manually build the assembly. For example, if you add a new property to the view model and then bind an XML element to that property without building the assembly, the designer won't know that property exists and will not update because it's in an error state. Get used to Ctrl + B.

It could also be because the design dimensions are different from the display dimensions. See d:DesignWidth="800" d:DesignHeight="600" at the top.

I don't see how we're going to maintain complex UIs with avalonia. The tooling is really subpar.

I agree it's not great, but I have found it manageable. SettingsDialog is the most complicated UI element, but it's broken down into tab items which should make browsing XML easier. Just collapse every TabItem that you aren't currently working on.

I think all other controls are much more manageable because they are much better encapsulated, but I'll let you be the judge of that.

I agree. This one's gotten out of control; it's the nature of the beast. I think it's reached the tipping point to justify decomposition.

Thanks again for these notes. I'll link to this PR in the avalonia primer.

PR #363, opened 2022-08-18
Change assembly loadig

### LINKED ISSUE #361: Follow up from reddit user v8.4.1 pre-release bug
Following up from [here](https://www.reddit.com/r/audible/comments/wotihu/windows_10_support_has_been_nixed_how_am_i/ikgxbob/)

If you have no tags and if all books have been downloaded, the steps are easy, but this is a pretty specific case. Let me know if this is you.

Otherwise, let's see if your logs can help. Your log file is in `C:\Users\<user name>\Libation`. Can you attach it to this issue/ticket?

Mbucari:
Hopefully this fixes #361 

Changes:

- Only execute the *ConfigApp.exe if debugging. Published standalone builds have all dependencies inside the Libation folder, so load by filename in release builds.
- Search for *ConfigApp.dll instead of *ConfigApp.exe because Linux and Mac don't use the exe extension, but dotnet still uses the dll extension on those platforms.
- When resolving assemblies, look for the exact name. If it's not found, log that it's not found and return null instead of throwing an exception inside the AssemblyResolve event handler. This will let the runtime handle the assembly not found error. This event was firing for "Libation.resources, Version=1.0.0.0, Culture=en-US, PublicKeyToken=null", and since our handler couldn't find a file starting with "Libation.resources" it threw an exception. I don't know why that wasn't causing problems for everyone, but it could be the culprit we're looking for.

### DIFF
--- a/Source/LibationFileManager/InteropFactory.cs
+++ b/Source/LibationFileManager/InteropFactory.cs
@@ -121,7 +121,6 @@ namespace LibationFileManager
 
 #if DEBUG
 
-            // `First` instead of `FirstOrDefault`. If it's not present we're going to fail anyway. May as well be here
             var modulePath = ModuleList.SingleOrDefault(m => m.ModuleName.EqualsInsensitive(asmName))?.FileName;
 #else
             var here = Path.GetDirectoryName(Environment.ProcessPath);
@@ -134,6 +133,7 @@ namespace LibationFileManager
 #endif
             if (modulePath is null)
             {
+                //Let the runtime handle any dll not found exceptions.
                 Serilog.Log.Logger.Error($"Unable to load module {args.Name}");
                 return null;
             }
--- a/Source/LibationFileManager/InteropFactory.cs
+++ b/Source/LibationFileManager/InteropFactory.cs
@@ -66,7 +66,7 @@ namespace LibationFileManager
         {
             var here = Path.GetDirectoryName(Environment.ProcessPath);
 
-            // find '*ConfigApp.exe' files
+            // find '*ConfigApp.dll' files
             var appName =
                 Directory.EnumerateFiles(here, $"*{CONFIG_APP_ENDING}", SearchOption.TopDirectoryOnly)
                 // sanity check. shouldn't ever be true
--- a/Source/LibationFileManager/InteropFactory.cs
+++ b/Source/LibationFileManager/InteropFactory.cs
@@ -46,13 +46,17 @@ namespace LibationFileManager
                 return;
             }
 
+            /*
+            * Commented code used to locate assemblies from the *ConfigApp.exe's module list.
+            * Use this method to locate dependencies when they are not in Libation's program files directory.
 #if DEBUG
 
-            // runs the exe and gets the exe's loaded modules
-            ModuleList = LoadModuleList(Path.GetFileNameWithoutExtension(configApp))
-                .OrderBy(x => x.ModuleName)
-                .ToList();
+           // runs the exe and gets the exe's loaded modules
+           ModuleList = LoadModuleList(Path.GetFileNameWithoutExtension(configApp))
+               .OrderBy(x => x.ModuleName)
+               .ToList();
 #endif
+            */
 
             AppDomain.CurrentDomain.AssemblyResolve += CurrentDomain_AssemblyResolve;
 
@@ -76,6 +80,9 @@ namespace LibationFileManager
             return appName;
         }
 
+       /*
+        * Use this method to locate dependencies when they are not in Libation's program files directory.
+        * 
         private static List<ProcessModule> LoadModuleList(string exeName)
         {
             var proc = new Process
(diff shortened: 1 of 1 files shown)


Source/LibationFileManager/InteropFactory.cs
```diff
@@ -61,18 +68,13 @@ private static string getOSConfigApp()
             var here = Path.GetDirectoryName(Environment.ProcessPath);
 
             // find '*ConfigApp.exe' files
-            var exes =
-                Directory.EnumerateFiles(here, $"*{CONFIG_APP_ENDING}", SearchOption.TopDirectoryOnly)
+            var appName =
+                Directory.EnumerateFiles(here, $"*{CONFIG_APP_ENDING}*", SearchOption.TopDirectoryOnly)
```
rmcrackan, line 72: How will the final star help? Is it ever .dll.other ?

Mbucari: Oops. That was a mistake. Fixed.

Source/LibationFileManager/InteropFactory.cs
```diff
@@ -115,12 +117,28 @@ private static List<ProcessModule> LoadModuleList(string exeName)
...
         {
             // e.g. "System.Windows.Forms, Version=6.0.2.0, Culture=neutral, PublicKeyToken=b77a5c561934e089"
-            var asmName = args.Name.Split(',')[0];
+            var asmName = args.Name.Split(',')[0] + ".dll";
 
-            // `First` instead of `FirstOrDefault`. If it's not present we're going to fail anyway. May as well be here
-            var module = ModuleList.First(m => m.ModuleName.StartsWith(asmName));
+#if DEBUG
+
+            var modulePath = ModuleList.SingleOrDefault(m => m.ModuleName.EqualsInsensitive(asmName))?.FileName;
```
rmcrackan, line 123: If the `#else` code works in `DEBUG` also, it seems unnecessary to have special code for the `DEBUG` pragma

Source/LibationFileManager/InteropFactory.cs
```diff
@@ -115,12 +117,28 @@ private static List<ProcessModule> LoadModuleList(string exeName)
...
+            var modulePath =
+                Directory.EnumerateFiles(here, asmName, SearchOption.TopDirectoryOnly)
+                .SingleOrDefault();
+
+#endif
+            if (modulePath is null)
+            {
+                //Let the runtime handle any dll not found exceptions.
+                Serilog.Log.Logger.Error($"Unable to load module {args.Name}");
+                return null;
```
rmcrackan, line 137: I assume the program will crash when it gets this null, right? I like that we log it first.

rmcrackan:
Thank you! This looks very good. I have a few questions and nit-picks before I merge.

Source/LibationFileManager/InteropFactory.cs
Mbucari: It does if all dependencies copied to the same debug directory. That's a debug environment thing.  We can trim it if you like.

rmcrackan: If they aren't all copied the same dir then I don't think the debug can find it anyway. Or am I wrong?

Source/LibationFileManager/InteropFactory.cs
Mbucari: Not necessarily. The runtime will attempt to load assemblies for types when those types enter scope. If the code bringing a type into scope is non-critical and wrapped inside a try/catch block, the failure will be caught and execution will continue.

Source/LibationFileManager/InteropFactory.cs
Mbucari: I just went ahead and commented out the hot loading code.

rmcrackan: Last question: If they aren't all copied the same dir then I don't think the debug can find it anyway. Or am I wrong?

Mbucari: If you have the dotnet runtime installed on your machine (which of course you do if you're running VS), then the resolver will ask the runtime for the assembly name (which is in ProgramFiles). Otherwise, if it's not in the current directory or in %path%, it won't be found.

Mbucari: As an example, create a new .net6 winforms app and run it. Look inside the debug directory and you won't find any dependencies. By default, VS does not copy all dependencies to the debug directory.

rmcrackan: Oh, I get it. This is about the GAC et al. -- my mind was stuck thinking about just the AppConfig. Hmm, good point. My debug builds won't be the same as the all-inclusive publishes.

I think I like it this best in its current form: commented out with good comments

rmcrackan:
> Libation.resources

Damn you weren't kidding. The new build gives me over 50 of these in a row. That's going to make log walking fun in the future. Glad we're catching them though. Much thanks again!

rmcrackan:
How interesting. Those errors only happened in Classic so presumably Chardonnay should have worked for that user. Oh well. I'm going to publish the pre-release and have them take it for a spin.

Mbucari:
> How interesting. Those errors only happened in Classic so presumably Chardonnay should have worked for that user.

Yeah, but as that user said classic wouldn't load at all and chardonnay would at least run once.  If this does end up fixing his problem, then I suspect the error will have had something to do with partially matching assembly names. We were using First and StartsWith to find the assembly which had the possibility of returning an incorrect assembly. IDK why that would happen, but it seems just as plausible as anything else at this point.

PR #401, opened 2022-11-25
Add validate and release workflows

pixil98:
### Description
This pull requests adds two workflows to the project.

* Validate will run on any commit to master or on any pull request targeting master. It will build all 4 versions of the code and upload them as artifacts of the workflow. These can be later downloaded for troubleshooting/verifying they run as expected. Pull requests will also be blocked from merging until the workflow completes without any errors.

* Release will run on any tag that starts with a `v`. This workflow builds the projects using the same shared steps as validate then creates a draft release, renames the artifacts to be in the correct format, and uploads them to the release. All you should need to do to make the release official is to edit the body of the release and remove the draft status.

### Testing
Testing for this will need to be done on my fork which has the pipelines already. I've added you as a collaborator.
* Make sure that the build workflow for a pull request runs successfully. You can create a test pull request and optionally commit something to it.
  * Check that the artifacts generated under the actions tab run correctly on their respective OSes. 
* Try creating a test tag that starts with a `v` 
  * It should run the release workflow, you can verify this in actions tab for the repo.
  * Verify that the draft release is created and that you are happy with the naming of everything.

### Other Changes
I made a couple other changes to support the workflows.
* I updated all build targets to use folders that match the capitalization of the OSes used in the folder structure, `macos` -> `MacOS`. This was to simplify the shared build workflow so that I only needed one variable for the OS.
* The publish profiles were all still on net6.0 even though the projects were on 7. I updated them all to 7 so that it would build correctly.

### Not Done
There are a couple of things that I left undone that could be added.
* It's not running any unit tests, this should be done, minimally, as part of the validate step. It probably makes more sense to just add it to the shared build workflow. If you share the commands you're using to run the unit tests I'll add them as part of this request.
* It should be possible to add a body to the release automatically, probably based on the commit messages since the last release. I haven't looked into this and it didn't seem that important compared to the quality of life enhancement that getting the workflows in place will give.


pixil98:
It looks like it doesn't automatically run pipelines in this request. I suppose that's reasonable so that people can't maliciously run things on your repo. You can check out the latest pipelines on my fork to see what they look like. I've marked release 1.0.15 as released which is the latest release using net7.0.

pixil98:
@rmcrackan Will you have a chance to look at this PR soon? I'll have some time to work on adding a docker image build, but I'd like to get this merged in first if possible.

rmcrackan:
You're right. Apologies. It's been busy around here lately

pixil98:
> What's the benefit of fully building non-releases?

The main benefit is for pull requests. It can be configured (and I believe it is by default) to not allow a pull request to be merged if the build isn't successful. A normal dev workflow would be to do your dev in a branch and create a pull request when it's ready to be merged. At that point having it run an automated build acts as a sanity check to make sure the code merging to main isn't going to break the compile. Then when it merges to main it runs another build, once again to sanity check what is in main. It doesn't really need to save off the build artifacts in all of these cases like it is, but it doesn't cost anything and could be useful for troubleshooting. For instance if you're trying to fix an error someone is encountering you might link them to the build artifact to test it rather than having to do a point release just for troubleshooting.

> How should this change to accomidate your yaml?

I envision it working like this
* Create a tag `vX.Y.X`
  * This will run a build and create a draft release with the zips
* Edit the draft release notes
* Draft=false

> Unrelated: I got the email inviting me to "collaborate" in your fork. I'm unfamiliar with this github feature. What does this mean?

I've also never used this feature before. My intention was to let you use it for testing this since the pipelines aren't going to run here until they merge in. You can try creating tags there or creating a branch and making a pull request to see how it works.

rmcrackan:
I've still got questions about using tags vs creating a release. I'll figure it out next time I'm ready to release though.

rmcrackan:
I pushed my latest code and created a [tag](https://github.com/rmcrackan/Libation/releases/tag/v8.6.0). No new drafts are in [Releases](https://github.com/rmcrackan/Libation/releases). What step am I missing?

rmcrackan:
nvm -- I was just hasty. It works now

pixil98:
@rmcrackan, glad it worked, let me know what you think now that you've used it. It currently waits until the build is completed before creating the release, which usually takes 10 to 15 minutes. It's done this way to prevent creating empty releases if the build fails.

I'm working on adding a docker image to the release workflow and in the process I've made some under the hood tweaks. If there is anything you want added or changed let me know. Two things I'd like to add are unit tests and a linter. I have the linter added and working, I generally think they're a positive addition, but they can be annoying if you're not used to programming to the linter's standards. In this case I'm just using the github provided Super Linter. For unit tests, could you let me know what commands you're running to run the tests?

rmcrackan:
I use Visual Studio's built-in runner to run them manually. There aren't a ton. Currently just the small test solutions in [here](https://github.com/rmcrackan/Libation/tree/master/Source/_Tests). I do much more of my unit testing (also manually in VS) in my supporting libraries:

* https://github.com/rmcrackan/Dinah.Core/tree/master/_Tests
* https://github.com/rmcrackan/AudibleApi/tree/master/_Tests

rmcrackan:
FYI:

Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

pixil98:
It looks like those warnings are coming from the official create-release action (https://github.com/actions/create-release). I assume they'll come out with a v2 update before it breaks but it doesn't look like they have one at the moment.

rmcrackan:
So far I've been unable to get the builds from this PR change to work with the auto updater. To test this in your repo, publish a higher tag/version and change this variable to `pixil98`

https://github.com/pixil98/Libation/blob/11d59beeedd5b3940e8f262734fa6093f2533bd0/Source/AppScaffolding/LibationScaffolding.cs#L393

btw, I also updated the windows regex.s to accept either version of win/windows

https://github.com/rmcrackan/Libation/blob/ad8dac5fb0272f881004670d8534232ba4becb72/.releaseindex.json

pixil98:
I've found one problem and fixed it. The Windows Classic version should now update correctly. I was zipping the release up with the parent folder rather than just the contents of the folder. If I have time today I'll debug through the newer updating and see what's going wrong. Otherwise I'll set aside some time this weekend to look at it. Once I figure it out I'll get a PR to fix it.

rmcrackan:
Gracias. Fyi -- locally I also have to run a `dotnet clean -c Release` before each of the 4 builds

pixil98:
Okay, I have a feeling it is the same issue on both, but I can't test it at the moment because I'm on a mac and updating is disabled. If you're able to test it I can make the PR, otherwise I can test it sometime this weekend. To test it, grab this build https://github.com/pixil98/Libation/actions/runs/3654197613 and see if it properly updates.

The pipeline shouldn't need to perform a clean, it's using a separate build runner for each OS so they'll all start with a clean environment. As part of digging into this I did figure out how C# works a bit more and I'm going to make some tweaks to the builder and switch it over from a windows build runner to a linux one. That will allow it to properly mark the Linux and MacOS binaries as executable.

rmcrackan:
> see if it properly updates

No dice. I unzipped and used Settings \> About to check the version: 8.6.4.1 . When I said yes to the update, it looked like it went through the whole process, but on next launch it prompted for update and still said 8.6.4.1 . I tried it again just to be sure. Same result.

pixil98:
That was my fault, I tagged a release but didn't realize I needed to increase the version number in the source. I tagged a new release with the version set to 9.0.6 and it's working for both windows versions. I'll submit a PR to fix the pipeline.

rmcrackan:
Worked that time. Thanks!

PR #404, opened 2022-12-10
Fix Auto Updating

pixil98:
###  Changes
Release workflow no longer includes the parent folder in the zip.

### Testing Instructions
Download the windows builds at https://github.com/pixil98/Libation/actions/runs/3654197613 and make sure they both update automatically to version 9.0.6.


rmcrackan:
Thanks!

PR #405, opened 2022-12-12
Upgraded to Avalonia 11-Preview4

### LINKED ISSUE #398: Right-click copy in the UI
I am requesting the ability to right-click on a field in the user interface and have a copy option to be able to copy the content of that field to the clipboard. I am creating a formatted file for each audiobook. So I copy a field, eg. title, author, etc., and paste it into my file.  I appreciate that I can <ctrl>-c copy the info but it would be easier if I could copy, then paste, into my document without switching from mouse to keyboard.

Thank you for all of the effort you and the developers have put into this program.

Mbucari:
Upgrading to this preview build allows me to finish adding the illegal character replacement dialog. Now, as far as I'm aware, classic and chardonnay features now match features 1:1, except accessibility which is the whole reason classic is still around.

Avalonia 11 also adds accessibility. They say we shouldn't have to do anything and that it'll just work, but I don't know if it is working. I used Window's built-in narrator, but I couldn't get it to read any info from the data grid cells. I know nothing about accessibility, so I'd like someone to test it out and see if their screen reader will read the whole grid.

However, version 11 is still in preview. I found two bugs worked around them, but those are just the ones I found. For what it's worth, I tested all forms and workflows and everything worked.

I also:
- fixed  #398 - There's no right-click menu, but standard keyboard shortcuts cause the selected row(s) to be copied to the clipboard. 
- Refactored ProductsGrid and ProductsGridViewModel to make better use of built-in tools. The result should be MUCH more maintainable.


rmcrackan:
> except accessibility which is the whole reason classic is still around

That's a big reason. Winforms accessibility is free and amazing. I literally have done nothing for accessibility and yet disabled users will gush about how much better it is than alternatives. No one in my life is visually impiared; however I'm tapped into several disability networks and I'm very sympathetic to their needs not being met.

> version 11 is still in preview. I found and worked around two bugs, but those are just the ones I found. For what it's worth, I tested all forms and workflows and everything worked.

That's another reason classic is still around. That this is how much work it takes to use. And great job on this -- it  sounds like a *ton* of work.

The last reason classic is still around is me. I've yet to have any interaction with coding avalonia that wasn't painful. Winforms is an inferior programming model but the tooling is rock solid and the VS support is impeccable. I can slap a UI together then get back to the functional coding which is where I really want to spend my time. Something something cold dead hands.

Avalonia has been a unequivocal win for the community in bringing it to all desktops. It's slowed new features because I think of how much work it will take to fight avalonia into submission and I decide maybe the new feature isn't worth it. I highly value the work you've done in creating Chardonnay. I never could have.

> fixed #398 - There's no right-click menu, but standard keyboard shortcuts cause the selected row(s) to be copied to the clipboard.

Not sure what you mean here

> The result should be MUCH more maintainable

w00h00 -- always a fan of maintainability

> Added a placeholder context menu to the products grid

Awesome, thank you. I liked CharlieRussel's ideas of being able to do a la carte stuff from that context menu

Mbucari:
> I'm tapped into several disability networks

Mind asking them to take Chardonnay for a spin again after this update? They say it should just... work.

rmcrackan:
> Ctrl+C (or cmd+C on mac) will cause selected rows to be copied to the clipboard, but there's no right-click > context menu for copying.

Oh, I assumed that already worked. That one feature request is specifically for a rt-clk menu item with 'copy'. My free time is at a premium right now. If I can ever get around to it though, maybe I'll take a look at your new context menu.

> I'll at least help you create the UI element for you to modify (like I did with this context menu)

Thanks!

> Mind asking them to take Chardonnay for a spin again after this update? They say it should just... work. And if it still doesn't work for them, can you ask them what screen reader software they use so I can test it?

No one in my life uses a screen reader. Due to certain family members I'm close with people with various physical disabilities, esp muscular or limb loss. So I'm really tuned into what life looks like when you need assistance. I'm glad Libation is so accessible.

Mbucari:
> Oh, I assumed that already worked.

So had I. It's not something I used so I didn't ever think to test it. The problem was all the data grid columns are custom templates I made, and I needed to specify what data is copied to the clipboard.

> maybe I'll take a look at your new context menu

The one I made is static, so it won't support dynamically changing the menu items based on user settings. Let me know if you want a dynamic context menu and I can whip one up.

> My free time is at a premium right now.

Totally get it. I also have end-of-year work stuff to catch up with. Get to it when you get to it. But if you're planning to release this before implementing context menus, just comment out or delete these lines:

https://github.com/rmcrackan/Libation/blob/451af7bea9552884253130a8bf6bbcf18c03c6ca/Source/LibationAvalonia/Views/ProductsDisplay.axaml#L47-L53

rmcrackan:
I added that 'copy' context menu to classic [here](https://github.com/rmcrackan/Libation/commit/ce711a36bac0ec0d544c923990d5b63c00504ae5#diff-d8e7d7305ddc6f52a92c83d7be2cccbd1abedadcb774d23cd61fb4616cd4682d). Is this easy for you to add for avalonia? Right-click cover does nothing, rt-clk anything other than stoplight gives a copy which just calls this new shared bit of interop. Right-click stoplight coming soon

rmcrackan:
> Let me know if you want a dynamic context menu and I can whip one up.

Yes please. Looking back at CharlieRussel's suggestion, it will be dynamic

> just comment out or delete these lines

Thanks. I did just that and can uncomment later

PR #407, opened 2022-12-13
Add dynamic context menus to products grid

Mbucari:
I added grid cell dynamic context menus. Every time you right-click on a cell, `ProductsDisplay.ProductsGrid_CellContextMenuStripNeeded` will fire and you can create custom context menus. I added the "Copy Cell Value" menu item like you already did for winforms, and you can follow that pattern for adding other menu items.

I also implemented the "hidden" tag grey-out feature. I'd forgotten about that one until I saw you'd mentioned it elsewhere. While I was at it, I also cleaned up the ProductsDisplay xaml.

### DIFF
--- a/Source/LibationAvalonia/Assets/LibationStyles.xaml
+++ b/Source/LibationAvalonia/Assets/LibationStyles.xaml
@@ -1,8 +1,8 @@
 <Styles xmlns="https://github.com/avaloniaui" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
 	<Styles.Resources>
-		<Color x:Key="SeriesEntryGridBackgroundColor">#FFE6FFE6</Color>
+		<Color x:Key="SeriesEntryGridBackgroundColor">#cdffcd</Color>
 
-		<SolidColorBrush x:Key="SeriesEntryGridBackgroundBrush"  Color="{StaticResource SeriesEntryGridBackgroundColor}" />
+		<SolidColorBrush x:Key="SeriesEntryGridBackgroundBrush" Opacity="0.5" Color="{StaticResource SeriesEntryGridBackgroundColor}" />
 		<SolidColorBrush x:Key="ProcessQueueBookFailedBrush" Color="LightCoral" />
 		<SolidColorBrush x:Key="ProcessQueueBookCompletedBrush" Color="PaleGreen" />
 		<SolidColorBrush x:Key="ProcessQueueBookCancelledBrush" Color="Khaki" />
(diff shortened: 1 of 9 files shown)


Mbucari:
> Do you believe it will be pretty straight forward to adapt for the dynamic traffic light context menu?

Very. Just create a new Menu item and add it to the list like I did for copy.

Mbucari:
I made a little example for you to create nested context menus.

```C#
public void ProductsGrid_CellContextMenuStripNeeded(object sender, DataGridCellContextMenuStripNeededEventArgs args)
{
	if (args.Column.SortMemberPath == "Liberate")
	{
		var subItem1 = new MenuItem { Header = "Sub Item 1" };
		var subItem2 = new MenuItem { Header = "Sub Item 2" };

		var rootMenu = new MenuItem
		{
			Header = "Root Menu",
			Items = new AvaloniaList<MenuItem>()				
			{
				subItem1,
				subItem2
			}
		};

		var copy = new MenuItem { Header = "Copy" };

		args.ContextMenuItems.Add(copy);
		args.ContextMenuItems.Add(rootMenu);
	}
	else
	{
		// any non-stop light column
		// (except for the Cover column which does not have a context menu)
		var menuItem = new MenuItem { Header = "_Copy Cell Contents" };

		menuItem.Click += async (s, e)
			=> await Application.Current.Clipboard.SetTextAsync(args.CellClipboardContents);

		args.ContextMenuItems.Add(menuItem);
	}
}
```

rmcrackan:
Much thanks!

[Classic](https://github.com/rmcrackan/Libation/blob/master/Source/LibationWinForms/GridView/ProductsGrid.cs#L104)

[Chardonnay](https://github.com/rmcrackan/Libation/blob/master/Source/LibationAvalonia/Views/ProductsDisplay.axaml.cs#L77)

It feels weird that I had to [create this](https://github.com/rmcrackan/Libation/blob/master/Source/LibationAvalonia/AvaloniaUtils.cs#L21) to find the parent window. Does something like this not already exist?

Mbucari:
Nice!

> Does something like this not already exist?

You can use `((IClassicDesktopStyleApplicationLifetime)Application.Current.ApplicationLifetime).MainWindow` to get the main window handle, which is what I do in most places where I need a window handle inside a control. If you need the parent windows' handle and it's not the main window, then what you did is pretty much it I think (or you can pass the Window to Control so it has it).

Mbucari:
I was actually working on this, but you beat me to it!

BTW, an underscore in a `MenuItem` Header is for keyboard shortcuts keys (Winforms uses '&').

rmcrackan:
> underscore

Whoops. Thanks. I forgot I left those in there. Fixed

Mbucari:
One mroe thing, the file extension filter should just be "*"

rmcrackan:
Good catch

Issue #409, opened 2022-12-15
Hangover Chardonnay


rmcrackan:
Bring Hangover Chardonnay into feature parity with Classic. [This commit](https://github.com/rmcrackan/Libation/commit/b2cf837de7b099d41bbb98d1881491f5bb959bea) added a tab to restore deleted books

rmcrackan:
Fixed in @Mbucari 's latest PR -- #412 . Will go out with the next release

Issue #357, opened 2022-08-13
Character replacement failed on stylized single quote


pixil98:
**Describe the bug**
I tried liberating this [book](https://www.audible.com/pd/Mrs-Frisby-and-the-Rats-of-NIMH-Audiobook/B00852UXJ8), you'll notice the author's name is written with a stylized closing single quote. I have both stylized single quotes setup to be replaced with a standard single quote:

```
  "ReplacementCharacters": {
    "Replacement": [
      {
        "CharacterToReplace": "\u0000",
        "ReplacementString": "_",
        "Description": "All other invalid characters"
      },
      {
        "CharacterToReplace": "/",
        "ReplacementString": "_",
        "Description": "Forward Slash (Filename Only)"
      },
      {
        "CharacterToReplace": "\\",
        "ReplacementString": "_",
        "Description": "Back Slash (Filename Only)"
      },
      {
        "CharacterToReplace": "\"",
        "ReplacementString": "'",
        "Description": "Open Quote"
      },
      {
        "CharacterToReplace": "\"",
        "ReplacementString": "'",
        "Description": "Close Quote"
      },
      {
        "CharacterToReplace": "\"",
        "ReplacementString": "'",
        "Description": "Other Quote"
      },
      {
        "CharacterToReplace": "<",
        "ReplacementString": "{",
        "Description": "Open Angle Bracket"
      },
      {
        "CharacterToReplace": ">",
        "ReplacementString": "}",
        "Description": "Close Angle Bracket"
      },
      {
        "CharacterToReplace": ":",
        "ReplacementString": "-",
        "Description": "Colon"
      },
      {
        "CharacterToReplace": "‘",
        "ReplacementString": "'",
        "Description": "opening single quote"
      },
      {
        "CharacterToReplace": "’",
        "ReplacementString": "'",
        "Description": "closing single quote"
      }
    ]
  },
```

Here is the resulting folder, note the stylized quote.
```
drwxr-xr-x  3 k8s k8s  3 Aug 12 21:47 'Robert O’Brien'
```

**To Reproduce**
1. Setup your Settings.json with the above replacements
2. Try liberating the book

**Expected behavior**
I'd expect it to write the author's folder name using the substituted single quote

**Platform**
Linux -- Docker image https://hub.docker.com/r/pixil/libation
Sorry there isn't any documentation written for the Docker image yet. I'd be happy to help set it up if you want to try using it.

**Log Files**
[...]

Issue #352, opened 2022-08-11
Better error messages for these scenarios


rmcrackan:
**Problem**

These 2 bug reports keep coming to me in reddit PMs

1. they don't have permission -- especially if a book expired from the Plus catalog between the time when they scanned and tried to download
2. the rare user will start getting errors on all downloads. Libation won't work with audible for some people for about a day. It always comes back on its own in less than 2 days and doesn't seem to happen again. It doesn't seem to be users who hit the api a lot. I can find no pattern. Once they tell me it's *every* download, I know this is happening.

**Solution**

In the case of common misunderstandings, it's the system's fault for not being more clear.

1. Libation should be able to detect license issues and let the user know that the most likely cause is returned book or audible removed it from plus. Ideally we'd also know it's a plus title but we don't currently track that
2. The only one of these I can find logs for right now is #336 which was logged as a networking error -- I don't know if this is the same or not. Any time there's a batch of more than 3 books with 100% failures, we should suggest waiting a day and creating an issue ticket with log file if it doesn't resolve itself.

Mbucari:
Re 2, I suspect that inside the Audible app there is a built-in fallback to a secondary server when the primary is offline due to maintenance or an upgrade. The timing makes sense (being unreachable for ~1 day). Ideally we would debug the real audible app while the primary is down to find the secondary endpoint, but I haven't been able to do that yet.

rmcrackan:
I assumed something like that must be the case. I originally suspected that they were punishing heavy users but there's a lot of reasons that doesn't add up. Unless we can catch it in the act and intercept the new traffic, all we can really do is inform the user.

Mbucari:
It's funny you closed this now, because I was thinking about this problem and I think I've discovered a pattern. In both cases, `GetDownloadLicenseAsync` fails with a "Content License denied ..." message. All license denials come with license denial reasons like this:

```Json
{
   "license_denial_reasons":[
      {
         "message":"...",
         "rejectionReason":"...",
         "validationType":"Membership"
      },
      {
         "message":"...",
         "rejectionReason":"...",
         "validationType":"Ownership"
      },
      {
         "message":"...",
         "rejectionReason":"...",
         "validationType":"Client"
      },
      {
         "message":"...",
         "rejectionReason":"...",
         "validationType":"AYCL"
      }
   ]
}
```

[...]

PR #412, opened 2022-12-16
Fix Character Replacements and Add More Useful Error Messages

Mbucari:
- Filename / path character replacements now work as expected on both Windows and Unix. Additionally, it's now a general file character replacement tool as requested in #357.
- Updated all tests to work on both Windows and Unix
- Add new fail states and log messages for `ContentLicenseDeniedException`. If the temporary denial of service is detected, launch a message box informing the user. #352 
- Added deleted books tab to Hangover. #409 

And ignore the Settings.Designer file. All I did was change the size of the "Edit how filename characters are replaced" button.


rmcrackan:
Woah, this is a huge one. I bet those file path tests took a ton of work. I spent a few hours last night thinking I had an easy fix then woke up this morning to see why they wouldn't work so I reverted my local changes.

Can you look at the 'files changed' tab for this PR and ctrl+F `obsolete` ? The new build files have started to mark these. **If** they're low effort, can you fix them? If not -- if it ain't broke...

Mbucari:
> I bet those file path tests took a ton of work. 

And about 1/4 of that work was figuring out how to debug tests on my Ubuntu VM, and 1/2 of the work was making character replacement work for both Windows and Unix!

> **If** they're low effort, can you fix them

All of the file dialog ones are pretty easy. The reason I haven't changed `WebRequest` (which was deprecated in 6.0) is because it has the `AddRange()` method built-in. But I'll see how much bother it is to change to `HttpClient`

Mbucari:
Oh hey, did you see this?

https://github.com/rmcrackan/Libation/blob/19860e9f09e2416b303b6e86f386f549c8f947e4/Source/LibationFileManager/Configuration.Logging.cs#L18-L27

Now every file path will be logged once instead of 3 times!

rmcrackan:
NIIIIIICE -- no I did not see this. Actually 4 times: Path, ShortPathName, LongPathName, PathWithoutPrefix. tbh, I've gotten so used to viewing paths in logs that way that I'd stopped questioning it. I prefer the PathWithoutPrefix since that's how I'm used to seeing it, but no complaints about this way either.

Now that I'm scrutinizing my old vs new logs I noticed something else. A few paths aren't longpaths (LibationFiles, InProgress). Obviously that's not a problem though or it would have bitten us by now.

Issue #410, opened 2022-12-15
Pipeline: Use Linux runners


pixil98:
The build pipeline is currently using Windows runners to perform the build. If we switch to Linux we gain the benefit of setting Linux and MacOS binaries as executable. They are also significantly faster to setup and run. There are a couple of problems standing in the way of fully implementing them.

* Unit tests fail when run on Linux. The couple I looked at seem to be related to newline handling. If this is the extent of the problems it should be solvable by using `Environment.NewLine`.
* Building Windows versions on Linux runs into problems detailed in https://github.com/dotnet/runtime/issues/3828. For now we can split the build pipeline in half and move the Linux and MacOS builds over to Linux runners.

Part of this change is going to be to use tar.gz files for Linux and MacOS instead of zip files. This will preserve permissions on files.

@rmcrackan - you can assign this to me, I would appreciate some of your time to look at the unit tests.

rmcrackan:
As long as I can still use VS's "Test Explorer" to run tests locally, I have no preferences how they run in devops land. Can you tell me which ones are failing due to newline? I'd like to personally see what I did wrong

rmcrackan:
After you deduce the best way to run these here, I'll look at how easy this is to apply to the supporting repos: [Dinah.Core](https://github.com/rmcrackan/Dinah.Core) , [AudibleApi](https://github.com/rmcrackan/AudibleApi) , [LuceneNet303r2](https://github.com/rmcrackan/LuceneNet303r2) . That's where a lot more unit tests live.

pixil98:
You can view the output of the Unit Test step at https://github.com/pixil98/Libation/actions/runs/3701507387/jobs/6270874053. This shouldn't stop you from using the Test Explorer.

Getting the unit tests to run is pretty easy, I added the following step to the build workflows:

```
      - name: Unit test
        working-directory: ./Source
        run: dotnet test
```

Mbucari:
@pixil98 Thanks for adding the tests!

@rmcrackan The linux tests are failing because of the drive letter colon. I'm currently working on character replacements, so I'll try to fix the tests.

Mbucari:
@rmcrackan It looks like the removeInvalidWhitespace_regex isn't working on linux.

pixil98:
As part of the pipeline rewrite, I'm pulling the build version from AppScaffolding.csproj. Should I add a check when tagging a release to make sure that the tagged version matches the version in the project?

rmcrackan:
[...]

Issue #395, opened 2022-11-14
Linux Docker Container to automate library downloads


sarg3nt:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
I'd love a Linux Docker container version that we can set Environment variables to configure.  Many of us run a NAS at home and most NAS platforms now support Docker as standard for software installation.  

**Describe alternatives you've considered**
None that I can think of.

**Additional context**
I run UnRaid and use AudioBookshelf to host my books that Libation downloads for me, only at the moment it's a very manual process.  I'd like a Docker container we could run that we can configure with ENV vars or a config file and that has a method to initialize a library sync and download any new books on a schedule.  
I'm a DevOps engineer and have a decent amount of experience with .Net from back when I was a backend Software Engineer so I might be able to help with this.  I mainly focus on Linux, Docker and Kubernetes these days but could probably help with doing this if you'd like.  My schedule is a bit rough right now, so no promises on it being fast.

rmcrackan:
Sounds promising. I'm very much not a devops person so I'll need your help. (In fact we could really use another one at work -- looking for a job?) What's the difference between your goal and using the current CLI?

pixil98:
I made a docker image awhile ago using the Linux CLI. I run it in a Kubernetes CronJob once a week to download any new books directly into my Plex directory. I didn't document it since I was the only one using it, but I can put together some documentation and put the source up on github. It's pretty basic for now, but it wouldn't be too hard to wrap it in a script to periodically check for books and keep the image running. It should also be possible to have a default Settings.json file that is configured using environment variables.

For now if you want to try it, it's available at:
https://hub.docker.com/r/pixil/libation

It expects the following volumes to be mounted:
/config -- mount your configuration folder here
/db -- optionally mount your database here, if not provided it will assume it's in /config (I wanted to keep my config as configmaps rather than on persistent storage)

Additionally make sure you mount in the folder that Settings.json is configured to save to.

wtanksleyjr:
How does the configuration work? Do you shut down the container, run the
desktop version to set the config and exit it, and then once you start the
container it'll work? Or does one need to filter out path settings first
(obviously you won't see the same environment Docker sees)?

-Wm

pixil98:
Since mine is a one shot scan/download, you should be able to copy your existing Libation config directory, edit the Settings.json to use Linux style paths, then run the container mounting the new directory in as /config. 

Here is the Settings.json file I use in case it helps.
```
{
[...]

PR #416, opened 2022-12-19
Linux runners and Docker image

pixil98:
Fixes https://github.com/rmcrackan/Libation/issues/410, https://github.com/rmcrackan/Libation/issues/395

## Changes
* Refactored build workflows to use Linux runners for Linux and MacOS builds
  * Builds finish quicker
  * Executables are properly marked as executable
  * Linux and MacOS versions use gzip instead of zip in order to preserve permissions
* Build workflows get build version from AppScaffolding by default. Release workflow overrides version using the tag name.
* Pull requests and merges to main now run unit tests automatically on Linux and Windows.
* Docker image is now build and published automatically as part of release.
* Dependabot configuration to check github actions for updates.
* C# projects have been been updated to allow Linux to build Windows versions.
  * This is unused currently due to a bug in dotnet.

## Post merge
After merging this, you'll need to setup the following secrets for docker
* DOCKERHUB_USERNAME -- Your username, `rmcrackan1`
* DOCKERHUB_TOKEN -- A token you generate on docker's website. https://docs.docker.com/docker-hub/access-tokens/#create-an-access-token

## Example
Validate workflow: https://github.com/pixil98/Libation/actions/runs/3727173017
Release workflow: https://github.com/pixil98/Libation/actions/runs/3727147663
Release: https://github.com/pixil98/Libation/releases/tag/v9.0.16


rmcrackan:
I added the DOCKERHUB_TOKEN secret. DOCKERHUB_USERNAME isn't really a secret. Where do you suggest adding it to github?

rmcrackan:
Done

rmcrackan:
Not that it should matter for your work -- just in case though:

I got docker to give me back control of my old rmcrackan. Secrets are updated. This is what I'll start using again and I'll abandon rmcrackan1

pixil98:
That shouldn't matter, I set it up to base everything on the username secret so I can run the exact same workflow when testing and use my account. It would be good if you created the `libation` project in your account if you haven't already.

rmcrackan:
libation docker repository created

rmcrackan:
Gracias. What's the new release protocol? Do I still push a new tag to initiate publish?

pixil98:
Exact same as before, push a tag in the format of `vX.Y.Z` and it will build all four configurations, create a draft release, and publish a docker image with the tagged version as well as marking it as `latest`.

rmcrackan:
Thanks. I recall you'd said you were going to have your scripting compare it to app scaffolding. Is that just a sanity check?

pixil98:
No, I didn't do a sanity check like I wanted. It was more complicated because you're not tagging it with the same version. It probably wouldn't be too hard to add at some point, but I didn't have the time to right now. For now, versions are used in the following ways:
* Pull requests and merges to main will build with the version in AppScaffolding.csproj
* Releases will build with the version from the tag name
* The release itself will use the version from the tag name
* Docker images will be created using the version from the tag name as well as being set to latest

PR #417, opened 2022-12-19
Improve download cancellation

Mbucari:
I recently experienced a network timeout error while I was doing a batch podcast download, and I noticed I couldn't cancel a processable until after the network timed out. This PR improves cancellation at all stages of downloading by using `CancellationToken` for all network connection and stream operations.


rmcrackan:
Good call. That NetworkFileStream code looks a lot safer. Is it ready to be pulled in?

Mbucari:
> Is it ready to be pulled in?

Yessir

Issue #430, opened 2022-12-28
Not all PDFs are actually downloaded


wrampy:
**Describe the bug**
10 books in my library have accompanying PDFs. But when I search the Libation library folder there are only 2 PDFs.
Libation shows the PDF icon for all 10 audiobooks. The Liberate menu shows "All PDFs have been downloaded".
Clicking the PDF icon opens the proper directory in Nautilus, but these contain only CUE, JPG and M4B files, no PDF (except for the above mentioned 2 audiobooks).

**To Reproduce**
Opened Libation. Clicked the Liberate - "All PDFs have been downloaded" menu item. Nothing happened. Logs still show all 10 PDFs are downloaded.
I removed one of the audiobooks from Libation library. I also deleted the files. Rescanned my Audible library, which detected the now missing audiobook which the application successfully liberated, but there is still no PDF next to it. I do not see in the logs any attempt for downloading the PDF.
I can download the PDF from audible.com with a browser.

**Expected behavior**
All the accompanying PDFs properly downloaded.
A button or menu item to enforce PDF re-download.
Logs to show the path of the PDF where it was saved.

**Platform**
Linux (Ubuntu 22.04)
Version 8.5.1.1

**Log Files**
[Log202212.log](https://github.com/rmcrackan/Libation/files/10316562/Log202212.log)

rmcrackan:
Sorry about this. Do you still have the older log from when the pdfs were downloaded? If so, can you also attach them to this bug? I'm curious to know if they could shed some light.

wrampy:
No, I don't have the old logs, but I can reproduce the behavior.

When I first downloaded the whole library I run into issues due to filenames being too long. Therefore I changed the filename pattern in the settings and re-downloaded/liberated all the audiobooks. And it looks like the PDFs were not downloaded after the first time. The two audiobooks that have the PDFs were added later, so there was no re-downloading.

Steps to reproduce:
1. search for the book
2. click the "Set the Downloaded status manually..." menu item under "Visible books"
3. change it to "Not Downloaded" and save it
4. the semaphore lights next to the book will change to red
5. the menu items under "Liberate" show that 1 audiobook needs liberation, but all PDFs were downloaded ![libation](https://user-images.githubusercontent.com/71339764/209918274-939239f7-6b57-4114-83ee-414914bc3fed.png)

You can see this behavior in the logs I already attached.

[...]

PR #431, opened 2022-12-29
Fix file naming template on unix systems (#430)

Mbucari:
This was a fun one!

There were a few bugs with how FileManager dealt with filenames which led to #430 

- `GetFilePath` was calling `replaceFileName` on the full filename including the extension. Depending on the user's template, it was possible for the file extension to be truncated.
- ext* filesystems limit the number of bytes in a filename, whereas NTFS and FAT limit the number of characters. If a filename contained any non-ASCII characters, the filename's UTF-8 byte length would be longer than the character length, but Libation wouldn't truncate the filename.
- While ext* filesystems don't have a path length limit, Linux has a 4096 byte limit and OSX has a 1024 byte limit.


rmcrackan:
Well done my dude!

ffs -- there's got to be a library which already handles these low level headaches so we don't have do reinvent every wheel.

Issue #346, opened 2022-08-07
Writing narrator to TCOM (COMPOSER)	tag


TheMetalCenter:
**Is your feature request related to a problem? Please describe.**
No. 

**Describe the solution you'd like**
The composer tag is pretty widely used as a field for the narrator, which is currently written to the file as `nrt`. It would be useful if the narrator was also written to the `TCOM` composer field. 

**Describe alternatives you've considered**
Manually writing the tag

**Additional context**
![image](https://user-images.githubusercontent.com/49706654/183272424-5246b30e-01b7-4943-acad-983c45170f9f.png)

Thank you for developing such a useful program!

rmcrackan:
A similar bit of conversation happened [last week](https://www.reddit.com/r/audible/comments/wcjrpe/audible_backup_one_big_file_or_a_file_per_chapter/iidlbi2/) on reddit. As noted in that thread -- my recommendation for now is to use [MP3TAG](https://www.mp3tag.de/en/), especially [this feature](https://docs.mp3tag.de/getting-started/first-steps/#copying-one-field-to-another-field).

For the time being, I'll leave this enhancement request open in case we decide to incorporate it into Libation also.

TheMetalCenter:
Thank you for the response! That feature is indeed powerful, I've found "Filename to Tag" very helpful in the past. In this case, however, I'm not sure if it works; though I may have the tag code incorrect. I tested the "Tag to Tag" option with `NRT`, `NARRATOR`, and `©NRT` (which is what MusicBee's tag inspector says the tag code is) but MP3TAG did not find a tag with those codes. The tag is also not observed by MP3TAG's extended tag feature. 

Anyway, it's not a huge deal, I just thought I'd suggest it as it would simply workflows for some whenever a new audiobook is added to collection.

Mbucari:
I just like to pipe in and say that the tags are copied verbatim from the audible files. ©nrt is the field code used by apples QuickTime tags.

See https://exiftool.org/TagNames/QuickTime.html

(Unfortunately I couldn't find an official Apple reference)

RandmTask:
I would love to see the Narrator in the Composer tag as well. PLEX and Prologue both use this as the Narrator field. Would be great to get it to work without having to post process the files. 

At the moment I have to use a program called TagEditor by Amvidia (not free) to update the Narrator field

FYI MP3Tag on a Mac is not a free download, it costs $21

Mbucari:
[...]

PR #432, opened 2022-12-31
Add ability for users to edit book ratings from the main grid

Mbucari:
I'm really happy with how this is turning out. I'm leaving this as draft because I'll probably be doing other stuff over the weekend and I expect you may not get around to reviewing it before the Tuesday, but I think the rating feature is ready. Please take if for a spin and give me some feedback.

## Things in this PR
- The biggie: user ratings are editable. This feature touched a lot of the code base, bust mostly in minor ways.
  - For classic, I added `MyRatingGridViewColumn` which uses `MyRatingCellEditor` as its editing control. When you click into a My Rating cell, the editor displays and allows you to change all three rating types.
  - Chardonnay works substantially similarly, using the new `DataGridMyRatingColumn` custom column with `MyRatingCellEditor` as the editing control.
  - To save the edited ratings to the DB, I had to tweak `LibraryCommands.UpdateUserDefinedItem()` and `LibraryCommands.UpdateBookStatus`.
  - To re-index the search engine, I had to tweak `SearchEngine` and `SearchEngineCommands`.
    - In `SearchEngineCommands` I removed the separate methods for tags and liberated status and combined all attributes into a single `UpdateUserDefinedItems` method. I didn't see the point in using three different methods, but please correct me if I missed something.
    - In `SearchEngine`, I added `UpdateUserRatings`. While doing this I noticed [this bug](https://github.com/rmcrackan/Libation/pull/432#issuecomment-1368317527), so I changed all field names to lowercase when calling `RemoveField()`.
- Implemented automatic re-filtering on property change in Chardonnay. I thought this was already done, but I guess I missed it.
- Fixed NREs on file dialogs.
- Deleted Form1 from WindowsConfigApp
- Remove `migrate_from_7_10_1`. That was to fix my screwup from way back. The bad release that caused it was only in the wild briefly, and this migration code has been running for months.
- Remove `findAndAddMissingParents()` and `removedOrphanedEpisodes()` from `LibraryCommands`. These helped migrate users' DBs to the new model where we store series parents in addition to individual episodes. Again, this was added MONTHS ago and every regular user should be upgraded now. Additionally, the library display code is more solid so orphaned episodes will no longer cause a crash.
- Updated AAXClean to add #346


Mbucari:
Hey @rmcrackan, I need some help with the search architecture. I'm trying to make use of the `SearchEngineCommands.SearchEngineUpdated` event to re-filter the grid after reindexing, but I think there's a bug. When `SearchEngineCommands.Search()` is called immediately after the search engine updates, I'm given old search results. However, if I manually type the same query a second later then the search completes with the new results.  There's some race condition here, but I'll be damned if I can find it. The search update and re-filter code is executing synchronously. The writer seems to flush to disk inside `updateDocument`, so the new results should be available when a new `SearchEngine` instance is created. I'm truly stumped.

Also, I noticed a bug in the search update. All your search field names are pascal case, but Lucene converts them to lowercase. `Document.RemoveField(string name)` is case sensitive, so those fields haven't been removed. I'm not surer why this wasn't a problem with the bool fields, but it is a problem with number fields.

Source/ApplicationServices/SearchEngineCommands.cs
```diff
@@ -43,28 +43,18 @@ public static void UpdateBooks(IEnumerable<Book> books)
...
-		);
-
-		internal static void UpdateLiberatedStatus(Book book) => performSafeCommand(e =>
-			e.UpdateLiberatedStatus(book)
-		);
-
-		internal static void UpdateBookTags(Book book) => performSafeCommand(e =>
-			e.UpdateTags(book.AudibleProductId, book.UserDefinedItem.Tags)
-		);
+		public static void FullReIndex() => performSafeCommand(fullReIndex);
```
rmcrackan, line 50: This is the kind of refactor where, after you see it, you say -- why the hell did I ever do it my way in the first place? ::forehead-slap::

rmcrackan:
> I noticed a bug in the search update

I never found this to be a bug. Glad you found it though if that's the case (no pun intended). I think there's still a refresh bug. But it's minor and I haven't been able to track it down. Maybe your case fix will solve it: if I'm filtering by '-israted' (which I almost always do), then I rate a book on audible, then I do a new scan: the book's rating will refresh in the grid but it would still be visible. If I search for something else then try my -israted search again, it disappears.

rmcrackan:
I notice this is no longer in draft. Ready for prime time?

I intended to test it this weekend. Then I had one of those culmination of wtfs that make you feel compelled to enumerate them, but when someone else does it to you you're like -- and you thought I needed this level of detail why? ...So I'll spare you that part.

Mbucari:
> Ready for prime time?

Yup! Like I said, I'm really happy with how it turned out. The UX feels so natural and seamless.

Issue #435, opened 2023-01-01
Feature Request -- Bandwidth Caps


bigtrucker26:
First, Let me THANK YOU for working on this program, and continually trying to improve it.

I am running into a slight issue, and I hope you would be willing to implement a solution.

The Libation program is trying to download at the maximum internet speed, and is basically blocking everything else on my network trying to sate it's thirst for my audiobooks.  While I appreciate this thirst, it is causing me a issue as it works.
Many download and streaming programs include the ability to limit how much bandwidth it may consume, allowing those of us with limited internet connections to maintain other data flows at the same time (both on the same machine, or on multiple machines).

So, I hereby request the ability to set limits (or cap the data rate) of the downloads (and uploads), to ensure it doesn't interfere with other internet traffic.  I leave it to you, to decide if you want to implement a solution with time based limits, or just a single limit (either is fine with me, but matters to some).

Again, Thank You for creating and maintaining this program, it is allowing me to manage and properly utilize a Large library of audiobooks (I'm a trucker, and I chew through them pretty quick), especially since audible won't make the necessary tweaks to their own app to allow proper categorization and organization..., let alone, integrate other sourced books into that system....

rmcrackan:
Flattery will get you everywhere :) I hadn't considered that people might want to throttle their downloads.

@Mbucari this sounds more like the kind of details you delve into. How big of an ask is this?

Mbucari:
@rmcrackan Just so happens I was just looking into this. I don't think it will be too bad to just add a timer into the download loop, but I'm not sure how to wire it to the UI. The downloader is pretty far down the stack., so that's a lot of layers to pass controls through. Settings are one thing because you can just pass them down on init, but speed limits need to be updatable at any time. Part of me just wants to use a static variable that's checked on each download loop iteration and be done with it. You cool with that?

rmcrackan:
Tell me more -- where would the static var live?

Mbucari:
> where would the static var live?

As a member of `NetworkFileStream`. I'd set it directly from the UI code when a download is started and again any time the user changes the setting.

rmcrackan:
Sure why not.

I once had ambitions of giving each setting a 'has changed' event. Well, wouldn't we want the before and after value? What about a 'before changed' event. I guess they each need a new shared 'setting' object and a way to compose with them and refactor all of Libation to use then and UGGG! who's got that kind of time? Screw it -- the current way has worked for over 5 years so if it ain't broke... 😄

[...]

PR #436, opened 2023-01-02
Add download speed limit

Mbucari:
As discussed in #435


Source/LibationAvalonia/ViewModels/ProcessQueueViewModel.cs
```diff
@@ -46,6 +53,37 @@ public ProcessQueueViewModel()
 		public bool AnyErrors => ErrorCount > 0;
 		public double Progress => 100d * Queue.Completed.Count / Queue.Count;
 
+		public decimal SpeedLimit
```
rmcrackan, line 56: just curious -- why decimal type?

Issue #421, opened 2022-12-21
Check for Updates with Download Link


lycannon:
Please add a version check on startup with a menu option (under Settings) to view the download page of the new version? Perhaps add a manual Check for Updates menu option under settings.

rmcrackan:
Interesting, please elaborate. In what way would you like this feature to differ from the current auto-update option which runs on startup?

lycannon:
It's easy to forget that there is an auto-update as there isn't any kind of notification / status message letting the user know they are on the latest version. Now that you mention it, I do remember getting the notification on a new release; but I imagine that people would want to wait until any current sync / conversions complete before updating.

Ha- Just started the app to look at the UI and got the notification for an update.

What I envision is updating the About dialog with additional information, see attached image.

![QR7vYyaWh4](https://user-images.githubusercontent.com/2049556/208820770-5b7a80a0-cbe0-4053-abb9-d681103f939b.png)

lycannon:
The update link would re-run the auto-updater.
The GitHub link would open the default system browser to the repo

I also just shared the app with a friend- took me a bit to track down the repo for the download link.

rmcrackan:
> people would want to wait until any current sync / conversions complete before updating

The updater doesn't run until you close the app

> attached image

Nice mock-up. Thanks!

> I also just shared the app with a friend- took me a bit to track down the repo for the download link

I'm open to suggestions. On the main page it has:

```
Libation: Liberate your Library
Download Libation
```

I also created [getlibation.com](https://getlibation.com). I need to do more to advertise it though. At the top, it has a tab for Download

rmcrackan:
Included in latest: [v8.8.1](https://github.com/rmcrackan/Libation/releases/latest)

Issue #345, opened 2022-08-06
Bookmarks/Clips


gabriel-vanca:
Would it be possible to make bookmarks/clips available? Perhaps they can be downloaded too when the book is downloaded.

Mbucari:
I think it's possible to download bookmarks and clips through the API, but the question is what file format would they be saved in, and how would you use them? Bookmarks and clips are designed to work with the Audible audiobook player. I suppose clips could be just that, pieces of the audiobooks stored as individual mp4 or mp3. But I don't know about bookmarks. How would you like to see bookmarks saved to a file?

Mbucari:
Bookmark timestamps and annotation text are easy: just put them in a CSV which you can then import into pretty much any software ever created. For the accompanying audio file, maybe:
1. Save the 3-ish second clip into same folders as a CSV
2. At a third column in the CSV with the audio files name

Sound good?

Mbucari:
Going back to the clips, I'm not crazy about having to split chunks of the audio out into their own files. Starting with what we have right now, that would just be a lot of changes to the audio processing routines. It would be a lot of work and I don't have confidence I'd get it right on the first try.

So instead of that, what about .cue files? One .cue file per clip, or all clips in a single Bookmarks.cue file?

In fact, bookmarks could be done in pretty much the same way. One .cue file for all of your bookmarks.

rmcrackan:
I don't use the bookmark features myself and I don't know how audible stores them. It sounds like you want small portions of the main audio file. Depending on how they're stored in your account, sounds to me like the output of a book + some time markers (which might be manageable by a cue file). Am I close? If not, where am I off?

rmcrackan:
@mkb79 What do you know about bookmarks? I see [this](https://audible.readthedocs.io/en/latest/misc/external_api.html#get-https---cde-ta-g7g.amazon.com-FionaCDEServiceEngine-sidecar). However, that doesn't look like bookmarks from an account.

mkb79:
@rmcrackan This will give you the bookmarks, clips and annotations for the given asin and the authenticated user. [Here](https://github.com/mkb79/audible-cli/blob/94e2d9a713973393c2c13d3802e7c82b92e6e752/src/audible_cli/models.py#L334) is the implementation for audible-cli.

Mbucari:
[...]

PR #440, opened 2023-01-07
Configuration Change Tracking and Bookk Records

Mbucari:
# Configuration now tracks properties changing and changed
All Property setters go through SetString() or SetNonString(), and those two methods get the existing value, compare it to the new value with Equals(), and then only fire the change events if the old and new values are not equal.
## There's a new base class to help subscribing to property change events: `PropertyChangeFilter`
This class allows users two ways to subscribe to change events; event subscription or observable.
### Event Subscription

You can either subscribe to all property changes like this:

```C#
Configuration.Instance.PropertyChanged += Configuration_PropertyChanged;

void Configuration_PropertyChanged(object sender, PropertyChangedEventArgsEx e)
{
  // All property changes will fire this handler.
}
```

Or you can filter which properties fire your handler with the new `PropertyChangeFilterAttribute`
```C#

Configuration.Instance.PropertyChanged += Configuration_PropertyChanged;

[PropertyChangeFilter(nameof(Configuration.AutoScan))]
[PropertyChangeFilter(nameof(Configuration.AutoDownloadEpisodes))]
void Configuration_PropertyChanged(object sender, PropertyChangedEventArgsEx e)
{
  // Only the AutoScan and AutoDownloadEpisodes properties will fire this handler
}
```

Both of these support multiple subscriptions and unsubscribes from one handler.
### Observable

You can create a disposable observable that watches for changes until the observer is disposed. For example:

```C#
 using var observer = Configuration.Instance.ObservePropertyChanging<long>(
                 nameof(Configuration.DownloadSpeedLimit),
                 MyPropertyChanged);

private void MyPropertyChanged(long oldValue, long newValue)
{
    // Fires whenever DownloadSpeedLimit changes until observer is disposed
}
```

I used an observable set changes to download speed. In fact this whole Configuration change tracking started because of[ your comment](https://github.com/rmcrackan/Libation/issues/435#issuecomment-1368618556).
# Added option to download Clips, Notes and Bookmarks

Per #345 

Both classic and Chardonnay have a context menu to view clips and bookmarks which opens up a data grid window. From there, users can delete the records or export them to xlsx, csv or json.

Additionally, there is a new setting to automatically download these records when liberating books.

# Added a full New Release window
Addresses #421
## Classic
Update_Classic

## Chardonnay Windows
Update_Chardonnay_win

## Chardonnay Non-Windows
Update_Chardonnay_nonwin
### Double clicking tags in template editor
You can now double click on tags and the template Editor to add them to the edit text box.

### various bug fixes
Fixing little bugs here and there as I found them.


Source/LibationFileManager/PropertyChangeFilter.cs
```diff
@@ -0,0 +1,331 @@
...
+				// "MyProperty" will fire this action.
+			});
+		
+		//The observer is delisted when cancellation is disposed
+
+	 */
+
+	#endregion
+
+	public abstract class PropertyChangeFilter
```
Mbucari, line 77: This class could/should probably go in Dinah. It's highly reusable.

Source/LibationFileManager/PropertyChangeFilter.cs
```diff
@@ -0,0 +1,331 @@
...
+
+					void filterer(object s, PropertyChangedEventArgsEx e)
+					{
+						if (e.PropertyName.In(matches)) value(s, e);
+					}
+
+					changedFilters.Add(new(value, filterer));
+
+					_propertyChanged += filterer;
+				}
```
Mbucari, line 119: If the handler has `PropertyChangeFilterAttribute`s, pass them through the `filterer` delegate and subscribe `filterer` to the event instead of `value`.

Source/LibationFileManager/Configuration.PersistentSettings.cs
```diff
@@ -3,336 +3,250 @@
...
+
+		public void SetString(string newValue, [CallerMemberName] string propertyName = "")
+		{
+			var existing = getExistingValue(propertyName);
+			if (existing?.Equals(newValue) is true) return;
+
+			OnPropertyChanging(propertyName, existing, newValue);
+			persistentDictionary.SetString(propertyName, newValue);
+			OnPropertyChanged(propertyName, newValue);
+		}
```
Mbucari, line 44: All setters go through these two methods which take care of checking for changes and then firing the events. 

`[CallerMemberName] string propertyName = ""` allows the caller to not pass its own name, simplifying all the setting's property definitions.

Source/LibationFileManager/Configuration.PersistentSettings.cs
```diff
@@ -3,336 +3,250 @@
...
+		public bool DownloadCoverArt { get => GetNonString<bool>(); set => SetNonString(value); }
+
+		[Description("Download clips and bookmarks?")]
+		public bool DownloadClipsBookmarks { get => GetNonString<bool>(); set => SetNonString(value); }
+		
+		[Description("File format to save clips and bookmarks")]
+		public ClipBookmarkFormat ClipsBookmarksFileFormat { get => GetNonString<ClipBookmarkFormat>(); set => SetNonString(value); }
+
+		[JsonConverter(typeof(StringEnumConverter))]
+		public enum ClipBookmarkFormat
```
Mbucari, line 150: With this property, enums will be serialized as strings instead of as numbers.

Source/FileManager/PersistentDictionary.cs
```diff
@@ -46,16 +46,28 @@ public string GetString(string propertyName)
...
+			{
+				if (jValue.Type == JTokenType.String && typeof(T).IsAssignableTo(typeof(Enum)))
+				{
+					return
+						Enum.TryParse(typeof(T), jValue.Value<string>(), out var enumVal)
+						? (T)enumVal
+						: Enum.GetValues(typeof(T)).Cast<T>().First();
+				}
+				return jValue.Value<T>();
+			}
```
Mbucari, line 66: This takes care of de-serializing enums from their stored string values.

Source/FileLiberator/DownloadOptions.cs
```diff
@@ -32,15 +40,51 @@ public string GetMultipartFileName(MultiConvertFileProperties props)
...
             // no null/empty check for key/iv. unencrypted files do not have them
-        }
-    }
+
+			LibraryBookDto = LibraryBook.ToDto();
+
+          cancellation = Configuration.Instance
+                        .ObservePropertyChanged<long>(
+                            nameof(Configuration.DownloadSpeedLimit),
+                            (_, s) => DownloadSpeedChanged?.Invoke(this, s));
```
Mbucari, line 87: Notify AaxDecryptor that the download speed has changed.

Source/LibationWinForms/Form1.ScanAuto.cs
```diff
@@ -56,12 +56,19 @@ private void Configure_ScanAuto()
...
+
+		[PropertyChangeFilter(nameof(Configuration.AutoScan))]
+		private void Configuration_PropertyChanged(object sender, PropertyChangedEventArgsEx e)
+		{
 			// when autoscan setting is changed, update menu checkbox and run autoscan
-			Configuration.Instance.AutoScanChanged += updateAutoScanLibraryToolStripMenuItem;
-			Configuration.Instance.AutoScanChanged += startAutoScan;
+			updateAutoScanLibraryToolStripMenuItem(sender, e);
+			startAutoScan(sender, e);
 		}
```
Mbucari, line 69: Detect changes to AutoScan property

Source/LibationAvalonia/Views/MainWindow.ScanAuto.cs
```diff
@@ -77,9 +79,11 @@ private void accountsPostSave(object sender = null, EventArgs e = null)
 				startAutoScan();
 		}
 
+		[PropertyChangeFilter(nameof(Configuration.AutoScan))]
 		private void startAutoScan(object sender = null, EventArgs e = null)
 		{
-			if (Configuration.Instance.AutoScan)
+			_viewModel.AutoScanChecked = Configuration.Instance.AutoScan;
+			if (_viewModel.AutoScanChecked)
 				autoScanTimer.PerformNow();
```
Mbucari, line 87: Detect changes to the AutoScan property

Source/LibationFileManager/PropertyChangeFilter.cs
```diff
@@ -0,0 +1,335 @@
...
+
+					void filterer(object s, PropertyChangedEventArgsEx e)
+					{
+						if (e.PropertyName.In(matches)) value(s, e);
+					}
+
+					changedFilters.Add((value, filterer));
+
+					_propertyChanged += filterer;
+				}
```
Mbucari, line 120: If the handler has `PropertyChangeFilterAttribute`s, pass them through the `filterer` delegate and subscribe `filterer` to the event instead of `value`.

Source/FileLiberator/DownloadOptions.cs
```diff
@@ -4,43 +4,88 @@
...
+			UserAgent = ArgumentValidator.EnsureNotNullOrEmpty(userAgent, nameof(userAgent));
+			// no null/empty check for key/iv. unencrypted files do not have them
+
+			LibraryBookDto = LibraryBook.ToDto();
+
+			cancellation =
+				Configuration.Instance
+				.ObservePropertyChanged<long>(
+					nameof(Configuration.DownloadSpeedLimit),
+					newVal => DownloadSpeedChanged?.Invoke(this, newVal));
```
Mbucari, line 88: Notify AaxDecryptor that the download speed has changed.

Source/LibationWinForms/Updater.cs
```diff
@@ -1,50 +1,57 @@
...
+
+				if (config.GetString(ignoreUpdate) == args.CurrentVersion)
+					return;
+
+				var notificationResult = new UpgradeNotificationDialog(upgradeProperties).ShowDialog();
+
+				if (notificationResult == DialogResult.Ignore)
+					config.SetString(upgradeProperties.LatestRelease.ToString(), ignoreUpdate);
+
+				if (notificationResult != DialogResult.Yes) return;
```
Mbucari, line 40: Use the new UpgradeNotificationDialog instead of a MessageBox in the updater.

Source/LibationFileManager/Configuration.PersistentSettings.cs
rmcrackan: Finally! I've been meaning to do something like `[CallerMemberName]` for ages. So much cleaner

While we're mucking in all this deep settings stuff, should default values go here then we can rip out all of that ad hoc add-setting-if-not-exists migration code?

Source/LibationFileManager/Configuration.PersistentSettings.cs
rmcrackan: Good idea. How will this work retro actively? ie -- with existing enum settings (assuming they exist)

Source/LibationWinForms/Form1.ScanAuto.cs
rmcrackan: Noice

Source/LibationFileManager/PropertyChangeFilter.cs
rmcrackan: lol I spent a long time combing through the code to figure out how you did your attribute magic. I should have looked at these comments first. This is literally what I was looking for :foreheadsmack:

Source/LibationFileManager/Configuration.PersistentSettings.cs
Mbucari: Works seamlessly with existing settings files. The comment right below this one shows how enums are deserialized.

Source/LibationFileManager/PropertyChangeFilter.cs
Mbucari: Yeah, This was originally the first comment to appear. Then I made a minor change and it became outdated, so I re-posted the comment and then it went to the bottom lol.

Source/LibationFileManager/Configuration.PersistentSettings.cs
Mbucari: Good idea. Let me think if there's an elegant way to do that. If it's inelegant, I'd just assume leave it the way it is and keep the Configuration file more readable.

rmcrackan: Agreed. That can be added anytime so I won't let that hold up this PR

Mbucari: @rmcrackan I put together what I think is an elegant solution for default values, but it results in a change that I'm not sure you'd be happy with.  Basically, settings won't be written to the Settings.json file unless they are different from the default. Personally I like it because it means that anything in the Settings.json file is not default (excepting basic stuff like Serilog and Books directory).

Thoughts?

rmcrackan: Hmm, interesting. Feels uncomfortable but not wrong per se. I could live with that

PR #442, opened 2023-01-09
Add a version parameter to DEB creation, do sanity checks, use it.

wtanksleyjr:
To create a DEB, we need to use the correct version. This PR offers a simple change to the shell script that adds a version parameter and checks it. I've confirmed that the resulting DEB file correctly operates as an "upgrade" of an existing install.

### DIFF
--- a/Source/targz2deb.sh
+++ b/Source/targz2deb.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-FILE=$1
+FILE=$1; shift
+VERSION=$1; shift
 
 if [ -z "$FILE" ]
 then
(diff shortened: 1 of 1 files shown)


rmcrackan:
Gracias!!

wtanksleyjr:
Who if anyone knows how to add this script as a step in the build process?
We should still provide the .tar.gz for Linux, but hosting a .DEB would
also be nice.

rmcrackan:
@pixil98 Does this sound like your wheelhouse?

rmcrackan:
> Who if anyone knows how to add this script as a step in the build process? We should still provide the .tar.gz for Linux, but hosting a .DEB would also be nice.

@pixil98 Does this sound like your wheelhouse?

*edit to include orig quote*

PR #443, opened 2023-01-10
Chardonnay Bugfix + Moved Default Settings into Configuration

Mbucari:
Hey, sorry for the rocky rollout of 8.8.

This PR is really to fix a couple of chardonnay bugs that I just noticed, but it also contains the migration of default settings into Configuration because I had already committed those.

I know 17 files looks like a lot, but 10 of them are <= 3 lines of changes.

I tested all settings with an existing settings file and a fresh installation, and every setting works as expected.

### DIFF
--- a/Source/AppScaffolding/LibationScaffolding.cs
+++ b/Source/AppScaffolding/LibationScaffolding.cs
@@ -77,117 +77,12 @@ namespace AppScaffolding
 		public static void RunPostConfigMigrations(Configuration config)
 		{
 			AudibleApiStorage.EnsureAccountsSettingsFileExists();
-			PopulateMissingConfigValues(config);
 
 			//
 			// migrations go below here
 			//
 
 			Migrations.migrate_to_v6_6_9(config);
-		}
-
-		public static void PopulateMissingConfigValues(Configuration config)
-		{
-			config.InProgress ??= Configuration.WinTemp;
-
-            if (!config.Exists(nameof(config.UseCoverAsFolderIcon)))
-                config.UseCoverAsFolderIcon = false;
-
-            if (!config.Exists(nameof(config.BetaOptIn)))
-                config.BetaOptIn = false;
-
-            if (!config.Exists(nameof(config.AllowLibationFixup)))
-				config.AllowLibationFixup = true;
-
-			if (!config.Exists(nameof(config.CreateCueSheet)))
-				config.CreateCueSheet = true;
-
-			if (!config.Exists(nameof(config.RetainAaxFile)))
-				config.RetainAaxFile = false;
-
-			if (!config.Exists(nameof(config.SplitFilesByChapter)))
-				config.SplitFilesByChapter = false;
-
-			if (!config.Exists(nameof(config.StripUnabridged)))
-				config.StripUnabridged = false;
-
-			if (!config.Exists(nameof(config.StripAudibleBrandAudio)))
-				config.StripAudibleBrandAudio = false;
-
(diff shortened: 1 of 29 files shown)


Source/LibationFileManager/Configuration.cs
```diff
@@ -24,9 +24,6 @@ public static bool SettingsFileIsValid(string settingsFile)
 			if (booksDir is null || !Directory.Exists(booksDir))
 				return false;
 
-			if (string.IsNullOrWhiteSpace(pDic.GetString(nameof(InProgress))))
-				return false;
-
```
Mbucari, line 29: IsNullOrWhiteSpace check moved into InProgress getter. Returns WinTemp if null or white space.

Mbucari:
I've tested the crap out of both classic and chardonnay, and everything I've tried works as expected.

Source/AppScaffolding/LibationScaffolding.cs
```diff
@@ -86,110 +85,6 @@ public static void RunPostConfigMigrations(Configuration config)
...
 		}
 
-		public static void PopulateMissingConfigValues(Configuration config)
-		{
-			config.InProgress ??= Configuration.WinTemp;
-
-            if (!config.Exists(nameof(config.UseCoverAsFolderIcon)))
-                config.UseCoverAsFolderIcon = false;
-
-            if (!config.Exists(nameof(config.BetaOptIn)))
```
rmcrackan, line 96: It feels like the right move to get these values out of scaffolding and into their own domain areas. I didn't check each of these default throughout your PR but I trust that you did

Source/LibationFileManager/Configuration.PersistentSettings.cs
```diff
@@ -20,9 +20,9 @@ public partial class Configuration
 
 		private PersistentDictionary persistentDictionary;
 
-		public T GetNonString<T>([CallerMemberName] string propertyName = "") => persistentDictionary.GetNonString<T>(propertyName);
+		public T GetNonString<T>(T defaultValue, [CallerMemberName] string propertyName = "") => persistentDictionary.GetNonString(propertyName, defaultValue);
 		public object GetObject([CallerMemberName] string propertyName = "") => persistentDictionary.GetObject(propertyName);
-		public string GetString([CallerMemberName] string propertyName = "") => persistentDictionary.GetString(propertyName);
+		public string GetString(string defaultValue = null, [CallerMemberName] string propertyName = "") => persistentDictionary.GetString(propertyName, defaultValue);
```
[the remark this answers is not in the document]
rmcrackan: What happens when the value is changed from default, to non default, then back to default? Is the entry removed from persistence? Or is the default value written? Either is fine. The only wrong answer is if the new value is detected to be default and persistence thinks it has no action -- thus leaving the non-default value in place

rmcrackan:
Reviewed. It feels weird but right.

Just that one question about default then I'm ready to merge.

Source/AppScaffolding/LibationScaffolding.cs
Mbucari: I did. One by one I deleted them from scaffolding and added the default value into configuration.

Mbucari:
If you decide you want default settings to be written to the Jason file, you can just edit PersistentDictionary to write the default value when the property is not found.

Source/LibationFileManager/Configuration.PersistentSettings.cs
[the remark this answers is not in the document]
rmcrackan: I just tried this. It persists the default, which is fine.

PR #444, opened 2023-01-11
Build and attach deb package

Mbucari:
This workflow will now build and publish a deb package along with the other binaries.

Before you release version 9, I need to fix AAXClean's dll management. It still tries to extract dlls from resources to the program files directory, and that doesn't work with anything other than windows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,3 +28,10 @@ jobs:
     with:
       version_override: ${{ inputs.version_override }}
       run_unit_tests: ${{ inputs.run_unit_tests }}
+
+  linux_deb:
+    needs: [linux]
+    uses: ./.github/workflows/build-deb.yml
+    with:
+      version: ${{ needs.linux.outputs.version }}
+      
--- a/.github/workflows/build-deb.yml
+++ b/.github/workflows/build-deb.yml
@@ -1,0 +1,39 @@
+# build-deb.yml
+# Reusable workflow that builds the Linux Debian package.
+---
+name: build_deb
+
+on:
+  workflow_call:
+    inputs:
+      version:
+        type: string
+        description: 'Version number'
+        required: true
+
+env:
+  FILE_NAME: "Libation.${{ inputs.version }}-linux-chardonnay"
+  
+jobs:
+  build_deb:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      
+      - name: Download Artifact
+        uses: actions/download-artifact@master
+        with:
+          name: "${{ env.FILE_NAME }}.tar.gz"
+
+      - name: Build .deb
+        id: deb
+        run: |
+          chmod +x ./Source/targz2deb.sh
+          ./Source/targz2deb.sh "${{ env.FILE_NAME }}.tar.gz" ${{ inputs.version }}
+      
+      - name: Publish .deb
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ env.FILE_NAME }}.deb
+          path: ${{ env.FILE_NAME }}.deb
+          if-no-files-found: error
--- a/.github/workflows/build-deb.yml
+++ b/.github/workflows/build-deb.yml
@@ -28,8 +28,8 @@ jobs:
       - name: Build .deb
         id: deb
         run: |
-          chmod +x ./Source/targz2deb.sh
-          ./Source/targz2deb.sh "${{ env.FILE_NAME }}.tar.gz" ${{ inputs.version }}
+          chmod +x ./.github/workflows/scripts/targz2deb.sh
+          ./.github/workflows/scripts/targz2deb.sh "${{ env.FILE_NAME }}.tar.gz" ${{ inputs.version }}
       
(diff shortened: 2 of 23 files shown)


Mbucari:
@pixil98 I have no idea what I'm doing, but I somehow managed to string together enough valid lines of yaml to get this done. Mined taking a look to see if I should change anything?

pixil98:
Yeah, sorry I’ve been mia lately. Life got really busy all of a sudden. I’ll take a pass over it.

pixil98:
The way I would implement this is to create a new yaml file similar to the docker one which is only called on the release workflow. It can depend on the build step being done already, pull the linux artifact that was built and then run the script on it. That way we have a reusable workflow module that is only run on release and doesn't need to do any matrix state checking. If you want to try tackling this go for it. Otherwise I probably won't have time for a week and a half to work on it and we could just release it like this for now and write up an issue to assign to me to come back to it.

Mbucari:
> create a new yaml file similar to the docker

That's what I tried originally, [but I ran into this issue](https://github.com/community/community/discussions/17245).  I tried to use [github-action-matrix-outputs-read](https://github.com/cloudposse/github-action-matrix-outputs-read), but I just could not get it to work.

Mbucari:
Oh, I see what you mean. I don't need the build job to output the tarball file name because I can just recreate it from the prerelease version. Thanks!

rmcrackan:
> @rmcrackan I fixed the AAXClean.Codecs nuget package so it now adds its native libs to the project and they're copied to the build. Now the only reason Libation has to write to its program files directory is for appsettings.json.

Thanks!

I get that a lot of people don't like that it's written to that directory. I do like it though. The real solution will eventually have to be a flow like

* if exists in curr dir: use it
* if exists in user dir: use it
* if can create in curr dir: do this
* else create in user dir

The basic logic isn't hard. The hard part is that this must be 100% rock solid code or *everything* gets bricked. Also the `UNSAFE` startup methods need the same logic.

Right now, I have a few users that find the current way to be subpar. When I finally try to update this, if I mess it up, we'll have a whole lot more upset users (...all of them)

Mbucari:
All right, it's finished.

Mbucari:
@rmcrackan I also tweaked a few other things Libation. Took care of some compiler warnings, reused some code, took figured out the correct way to bind Avaloniaproperties from code.

Probably the most significant change is for BookImporter to import average rating instead of star rating. Star rating is rounded to the nearest 0.5, but average rating is a true weighted average.

.github/workflows/build.yml
```diff
@@ -28,3 +28,10 @@ jobs:
     with:
       version_override: ${{ inputs.version_override }}
       run_unit_tests: ${{ inputs.run_unit_tests }}
+
+  linux_deb:
```
pixil98, line 32: Something we should figure out is whether we want to build a deb file for every PR/merge to main or only on release of a new tagged version. I liken this to the docker image where it's a specialized packaging of the build and we shouldn't need it for testing, outside of specialized cases. With docker I made the call to only do it on release, but I can see an argument for either way.

The advantage of doing it only on release is you could call it from the release yaml which already calculates the official release version from the tag name and passes that through to the current build scripts. As is the build scripts aren't a great source of truth for the version as their either grabbing the AppScaffolding version which is not ment to be end user facing or getting the version from the release workflow.

Mbucari: Makes sense, but the version number I need for build-deb is the one that was used to create the tar filename. I need to know which artifact to use with the deb-pkg script, so it has to match.

BTW, where does this value come from? There's no step names 'version' in this job.
https://github.com/rmcrackan/Libation/blob/09118b1ddf634adec35d213bb735604e80773a13/.github/workflows/release.yml#L53

pixil98: In a release context we generate the release version in the release.yml and pass it into the build scripts rather than parsing the value out of the csproj file. If you pass that same version into the new deb packager it would be able to construct the name of the artifact to pull down as long as it's marked as needing the build step first. There's probably something to be said for not defining the naming structure of the artifacts in multiple locations, but I'm not overly concerned about it.

That version is a bug and it should probably be `${{ needs.prerelease.outputs.version }}`. I had to move the version step into it's own prerelease job so that the version could be passed into both the build and docker jobs. I must have missed that one. If you wouldn't mind fixing it since you'll need to test making a release anyway.

Mbucari:
@rmcrackan I fixed #447

Mbucari:
> Lookin' good. Ready for merge?

In about an hour I will be. I friggin did it. I managed to compile a~1.6MB shared library on linux for decoding aac audio WITHOUT the need for installing ffmpeg on the system! I am STOKED! I've been trying to do this off and on for a year.

I'm going to upgrade AAXClean.Codecs and edit the deb script, then I'm done. WOOT!

Mbucari:
Done! I like @pixil98's suggestions, but I need to not look at yaml for a while.

rmcrackan:
> compile a ~1.3MB shared library on linux for decoding aac audio WITHOUT the need for installing ffmpeg on the system!

That's incredible!

PR #454, opened 2023-01-18
Fixed #453

### LINKED ISSUE #453: One of the last two updates has disabled custom file naming
**Describe the bug**
I updated Libation yesterday, after restarting it then told me to update again immediately, after doing both and restarting again I downloaded another dozen audio books and they all ended up in the same subfolder called 

(0000.00) ＜series＞ ＜series#＞꞉ ＜title short＞ [Audio] {＜year＞ ＜narrator＞} - ＜author＞

which is the folder format I defined in settings, however they aren't extracting the info from the file and inserting it into the folder name

**Screenshots**

**Platform**

Windows 10

**Log Files**
Attach your Libation log file here.

Capture
Capture2
Capture3
[Log202301.log](https://github.com/rmcrackan/Libation/files/10448298/Log202301.log)

Excellent work on the program in general folks, it's greatly appreciated.

Mbucari:
Recently I added a check to FileNamingTemplate.GetFilePath() to try to preserve the file extension from being obliterated by replaceFileName(). If GetPortionFilename() was called on a template that contains a period, GetFilePath() assumed everything that came after the dot was a file extension and would preserve it from the templating procedure.

My fix is to require callers to specify the file extension being used. Also, refrain from using Path.GetFileNameWithoutExtension() unless you know that the filename 1) is a file not a directory, and 2) knows the filename already has an extension. GetFileNameWithoutExtension is just too dumb.

### DIFF
--- a/Source/AaxDecrypter/AaxcDownloadMultiConverter.cs
+++ b/Source/AaxDecrypter/AaxcDownloadMultiConverter.cs
@@ -203,7 +203,8 @@ That naming may not be desirable for everyone, but it's an easy change to instea
 		private FileStream createOutputFileStream(MultiConvertFileProperties multiConvertFileProperties)
 		{
 			var fileName = DownloadOptions.GetMultipartFileName(multiConvertFileProperties);
-			fileName = FileUtility.GetValidFilename(fileName, DownloadOptions.ReplacementCharacters);
+			var extension = Path.GetExtension(fileName);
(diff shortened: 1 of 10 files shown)


rmcrackan:
With the file extension stuff, plz ensure the param is valid with or without the leading dot

rmcrackan:
nvm, forgot about `GetStandardizedExtension` -- that's already doing the lifting

PR #456, opened 2023-01-20
Add file creation DateTime to naming templates

Mbucari:
This was unbelievable easy to do thanks to your Template arch and Microsoft's powerful format providers!

### DIFF
--- a/Source/AaxDecrypter/MultiConvertFileProperties.cs
+++ b/Source/AaxDecrypter/MultiConvertFileProperties.cs
@@ -10,6 +10,6 @@ namespace AaxDecrypter
         public int PartsPosition { get; set; }
         public int PartsTotal { get; set; }
         public string Title { get; set; }
-
+        public DateTime FileDate { get; } = DateTime.Now;
     }
 }
--- a/Source/FileLiberator/UtilityExtensions.cs
+++ b/Source/FileLiberator/UtilityExtensions.cs
@@ -27,11 +27,13 @@ namespace FileLiberator
 		public static LibraryBookDto ToDto(this LibraryBook libraryBook) => new()
 		{
 			Account = libraryBook.Account,
+			DateAdded = libraryBook.DateAdded,
 
 			AudibleProductId = libraryBook.Book.AudibleProductId,
 			Title = libraryBook.Book.Title ?? "",
 			Locale = libraryBook.Book.Locale,
 			YearPublished = libraryBook.Book.DatePublished?.Year,
+			DatePublished = libraryBook.Book.DatePublished,
 
 			Authors = libraryBook.Book.Authors.Select(c => c.Name).ToList(),
 
--- a/Source/FileManager/FileNamingTemplate.cs
+++ b/Source/FileManager/FileNamingTemplate.cs
@@ -9,11 +9,15 @@ namespace FileManager
 	/// <summary>Get valid filename. Advanced features incl. parameterized template</summary>
 	public class FileNamingTemplate : NamingTemplate
 	{
+		public ReplacementCharacters ReplacementCharacters { get; }
 		/// <param name="template">Proposed file name with optional html-styled template tags.</param>
-		public FileNamingTemplate(string template) : base(template) { }
+		public FileNamingTemplate(string template, ReplacementCharacters replacement) : base(template)
+		{
+			ReplacementCharacters = replacement ?? ReplacementCharacters.Default;
+		}
 
 		/// <summary>Generate a valid path for this file or directory</summary>
-		public LongPath GetFilePath(ReplacementCharacters replacements, string fileExtension, bool returnFirstExisting = false)
+		public LongPath GetFilePath(string fileExtension, bool returnFirstExisting = false)
 		{
 			string fileName = 
 				Template.EndsWith(Path.DirectorySeparatorChar) || Template.EndsWith(Path.AltDirectorySeparatorChar) ?
@@ -22,7 +26,7 @@ namespace FileManager
 
 			List<string> pathParts = new();
 
-			var paramReplacements = ParameterReplacements.ToDictionary(r => $"<{formatKey(r.Key)}>", r => formatValue(r.Value, replacements));
+			var paramReplacements = ParameterReplacements.ToDictionary(r => $"<{formatKey(r.Key)}>", r => formatValue(r.Value, ReplacementCharacters));
 
 			while (!string.IsNullOrEmpty(fileName))
 			{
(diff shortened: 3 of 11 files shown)


Mbucari:
I'm converting this to draft because I have some uncertainty about how I implemented this. The dynamic nature of the daytime tag adds a new dimension to templates. Is there anywhere in libation where GetFileName will be called more than once trying to find the same file? If so, it's possible the time stamp will change and the file won't be found even though it exists.

To guard against that, maybe add a DateTime field to LibraryBookDto? The dto is created as the first file is about to be written, so it will be accurate down to the second for single file audiobooks. But, it'll lose accuracy for multi-files split audiobooks. So do we add a date time field to the chapter dto? If so, does it need a different tag name or should chapter dates override file dates?

rmcrackan:
What's the impetus for this addition?

Mbucari:
> What's the impetus for this addition?

[This guy](https://github.com/rmcrackan/Libation/issues/453) seemed to like having date info in his filenames. Other than that, nothing more or less than finding something to do with myself in the evening. Did you mean to be so existential? XD

rmcrackan:
I like using regex and give the user the power of format strings. Some thoughts:

* The dto feels like the right place for the reasons you mentioned
* Many date time formats result in slashes and invalid path characters like colons
* Please add unit tests, especially for the slashes and invalid path char cases
* We'll want to give the user some guidance. DateTime format strings can be tricky. Examples and/or documentation links. [standard](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings), [custom](https://learn.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings)
* I recommend optional space after date: `@"<date\s*\[(.*?)\]>"`
* What's the intended result if someone uses empty inside brackets? `<date[]>`
* If empty brackets are allowed as some default, then without brackets should probably be the same thing: `<date>`
* "date" feels too generic. Books have lots of relevant dates: date of publication, date added to audible account, date liberated/downloaded, ... This tag should use a better word, esp since any or all of these could be added in the future
* If you implement everything above, the regex is going to get hairy. I'm happy to help

Mbucari:
> The dto feels like the right place for the reasons you mentioned

Done.

> Many date time formats result in slashes and invalid path characters like colons

Handled. This took a lot of though, but I like my solution. Basically, date formatters are exempt from the initial round of verifications, and they are sanitized in `getFileNamingTemplate()` before `FileNamingTemplate` is instantiated.
> Please add unit tests, especially for the slashes and invalid path char cases

I made a ton. Please look them over and see if you can think of any edge cases I missed.
> We'll want to give the user some guidance. DateTime format strings can be tricky. Examples and/or documentation links. [standard](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings), [custom](https://learn.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings)

Some pointers in the wiki is a good idea, but we don't have to show them the entire range of capabilities. A lot of them are obvious (mm, ss, hh, HH, MM, yy, yyyy), and with the real-time display in `EditTemplateDialog` I figure users can play with it until they get a result they like.

> I recommend optional space after date: `@"<date\s*\[(.*?)\]>"`

Done.
> What's the intended result if someone uses empty inside brackets? `<date[]>`

I added a default value for empty format strings: yyyy-MM-dd

> If empty brackets are allowed as some default, then without brackets should probably be the same thing: `<date>`

Done.
> "date" feels too generic. Books have lots of relevant dates: date of publication, date added to audible account, date liberated/downloaded, ... This tag should use a better word, esp since any or all of these could be added in the future

Agreed. File creation date is now <filedate> ( or <file date>).  Also, since all the heavy lifting for custom date formatters was already done, I also added <pub date> and <date added> tags for the publication date and the date added to the library, respectively.

> If you implement everything above, the regex is going to get hairy. I'm happy to help

I think it turned out to be pretty simple because I didn't rely on regex to handle illegal path characters. The new regex patterns are:

```Regex
<file\s*?date\s*?(?:\s*?|\[(.*?)\])\s*?>
```
```Regex
<date\s*?added\s*?(?:\s*?|\[(.*?)\])\s*?>
```
```Regex
<pub\s*?date\s*?(?:\s*?|\[(.*?)\])\s*?>
```

rmcrackan:
This looks like what you intend (untested):

```
<file\s*?date     # important bit #1: type of date tag
\s*?              # optional space
(?:               # begin optional format block. don't capture space or [ ] literals
  \[              # start format
    (.*?)         # important bit #2: format capture
  \]              # end format
  \s*?            # optional space
)?                # end optional format block
>
```

Mbucari:
> This looks like what you intend

Almost, but this produces the intended results:

```regex
<file\s*?date\s*?(?:\[(.*?)\]){0,1}\s*?>
```

[All of these should be valid date tags.](https://regex101.com/r/9nbWz5/2)

Mbucari:
I added a couple more unit tests for these cases:

```
<filedate[yyyy][]>
<filedate[yyyy[]]>
```
and then decided on this regex pattern:

```regex
<file\s*?date\s*?(?:\[([^\[\]]*?)\]){0,1}\s*?>
```

https://regex101.com/r/9nbWz5/3

rmcrackan:
Looks good. btw

* `{0,1}` and `?` are equivalent. 
* putting the final `\s*?` directly after the last `]` is slightly more efficient when no brackets are present

Mbucari:
> Looks good. btw
> 
>     * `{0,1}` and `?` are equivalent.
> 
>     * putting the final `\s*?` directly after the last `]` is slightly more efficient when no brackets are present

Good to know! My general approach to writing regex expressions is to pull up [https://regex101.com](https://regex101.com/r/9nbWz5/5) and https://www.rexegg.com/regex-quickstart.html and then start trying stuff until I get a result that works for me lol. I'm a regex script kiddie!

PaulBrad25:
> > What's the impetus for this addition?
> 
> ~[This guy](https://github.com/rmcrackan/Libation/issues/453) seemed to like having date info in his filenames.~ Other than that, nothing more or less than finding something to do with myself in the evening. Did you mean to be so existential? XD
> 
> EDIT* Never mind. I misremembered what he said he used those placeholders for. So all I'm left with is just something to do lol.

My placeholder can't be filled in with audiobook data, sadly. You'd need a much more complicated script to go to goodreads, identify the book and scrape the 'first published' date. I'm happy with the insertion of 'audio recorded date' for now (I recognise that's only accurate for audible exclusives and audible upload dates)

For example this is my end goal for each author folder, so better to just leave it as a placeholder AFAIC;

Capture
Capture1

Mbucari:
I briefly looked into doing that. GoodReads has (or had ) an API, but they're in the process of shuddering it and not allowing new devs to sign up.

PR #462, opened 2023-01-25
Refactor AaxDecrypter

Mbucari:
This is an unusual PR for me. It's almost exclusively a refactor of AaxDecrypter for clarity and maintainability.

Also, I noticed the `FilePathCache` was broken since I added a second constructor to `LongPath`. I fixed that.


rmcrackan:
It builds again and unit tests are all passing. Gracias.

I like do refactors when they make the code cleaner. There's something satisfying about cleaning up a mess and reducing lines, even when functionality is unchanged. It's an indulgence you rarely get in corporate life so it's all the more fun on passion projects.

Mbucari:
Can confirm, was super satisfying.

Mbucari:
Although it is really annoying that visual studio's code metrics got ever so slightly worse after the refactor.

PR #471, opened 2023-02-02
Add a more general NamingTemplate

Mbucari:
This PR attempts to simultaneously simplify and increase the power of naming template. The core of the new template module is `FileManager.NamingTemplate.NamingTemplate`

A template is created using `NamingTemplate.Parse(string template, IEnumerable<TagClass> tagClasses)`.

`TagClass`s are collections of named tags linked to properties. There are 2 types of `TagClass`:
- `PropertyTagClass<TClass>` - Gets properties from `TClass` and converts them to string.
- `ConditionalTagClass<TClass>` - Supports conditional tags in the form of \<tagname-\>\<-tagname\>

You create these `TagClass`s and then register properties like this:
```C#
PropertyTagClass<LibraryBookDto> lbProperties = new();
lbProperties.RegisterProperty(TemplateTags.Title, lb => lb.Title ?? "");
lbProperties.RegisterProperty(TemplateTags.Series, lb => lb.SeriesName ?? "");

ConditionalTagClass<LibraryBookDto> lbConditions = new();
lbConditions.RegisterCondition(TemplateTags.IfSeries, lb => !string.IsNullOrWhiteSpace(lb.SeriesName));

var namingTemplate = NamingTemplate.Parse("<if series-><series><-if series> <title>", new TagClass[] { lbProperties, lbConditions });
```

And then you get the templated text like this:
```C#
TemplatePart templateParts = namingTemplate.Evaluate(dto);
```

The returned `TemplatePart` is an ordered linked list of all template parts, and each part is identifiable as either a property tag (along with the `TemplateTags` of that property) or a string literal (non-templated parts).

When registering properties, you have the option to provide a string formatting delegate with this signature:
```C#
delegate string PropertyFormatter<T>(ITemplateTag templateTag, T value, string formatString);
```

If you provide a formatter when registering a property, it will be called for that property when the template is evaluated. The `formatString` parameter is the contents of square brackets after the tag name (e.g. \<pub date\[\]\>). You can create any custom formatters you want. I added an \[l\] and \[u\] formatters for all strings to convert to lowercase or uppercase, respectively. I also added a \[#\] formatter to integer types (where # is an integer literal) that zero-pads the integer. And I kept the same date formatting as already exists.

I also added \<if podcast-\>\<-if podcast\> conditional tag for podcasts, and \<if bookseries-\>\<-if bookseries\> for series that are books and not podcasts.

I've been working on this for a while, and I'm really happy with the results. Not only is it more powerful in general, but templates will be a lot easier to add and modify later.


Mbucari:
Some examples:
|Pattern|Rendering|
|-|-|
|\<title short[u]\>|A STUDY IN SCARLET|
|<title[l]\>|a study in scarlet꞉ a sherlock holmes novel|
|\<series#[3]\>|001|
|\<bitrate[4]\>|0128|
|\<samplerate[6]\>|044100|

rmcrackan:
Wow, this is pretty slick. Can't say I'm surprised that my first draft at tags was replaced with something better. It wasn't bad but it always felt a bit half-baked. There's obviously too much churn here to micro-analyze. What I see looks good. I like that there are tests. The more the merrier.

* I assume this didn't remove/break any previous tag functionality, correct?
* RegisterProperty: Where is this called?
* What do you think about creating documentation? Nothing super in-depth. Users should be able to find your examples like those above. And that will give us a starting point if it needs to be expanded later.

Source/FileManager/NamingTemplate/PropertyTagClass[TClass].cs
```diff
@@ -0,0 +1,88 @@
...
+
+	/// <summary>
+	/// Register a string type property.
+	/// </summary>
+	/// <param name="propertyGetter">A Func to get the string property from <see cref="TClass"/></param>
+	/// <param name="formatter">Optional formatting function that accepts the string property and a formatting string and returnes the value formatted to string</param>
+	public void RegisterProperty(ITemplateTag templateTag, Func<TClass, string> propertyGetter, PropertyFormatter<string> formatter = null)
+		=> RegisterPropertyInternal(templateTag, propertyGetter, formatter);
+
+	private void RegisterPropertyInternal(ITemplateTag templateTag, Delegate propertyGetter, Delegate formatter)
```
Comment (rmcrackan) [inline Source/FileManager/NamingTemplate/PropertyTagClass[TClass].cs:42]: Good call. The other name rubbed me the wrong way but I didn't want to nit pick.

Mbucari:
>     * I assume this didn't remove/break any previous tag functionality, correct?

The only thing I removed is warnings for using chapter tags in file and folder templates. It didn't really make sense with the new parser because if a tag isn't registered it's treated as a string literal. I also removed warnings for illegal filename characters because the character replacer takes care of that, so no need to warn the user.

>     * What do you think about creating documentation? 

Absolutely. It's gotten to be just a bit too sophisticated to expect users to muddle their way through blind.

Source/LibationFileManager/Templates.cs
```diff
@@ -2,451 +2,347 @@
...
+			lbProperties.RegisterProperty(TemplateTags.SampleRate, lb => lb.SampleRate, IntegerFormatter);
+			lbProperties.RegisterProperty(TemplateTags.Channels, lb => lb.Channels, IntegerFormatter);
+			lbProperties.RegisterProperty(TemplateTags.Account, lb => lb.Account, StringFormatter);
+			lbProperties.RegisterProperty(TemplateTags.Locale, lb => lb.Locale, StringFormatter);
+			lbProperties.RegisterProperty(TemplateTags.YearPublished, lb => lb.YearPublished, IntegerFormatter);
+			lbProperties.RegisterProperty(TemplateTags.DatePublished, lb => lb.DatePublished, DateTimeFormatter);
+			lbProperties.RegisterProperty(TemplateTags.DateAdded, lb => lb.DateAdded, DateTimeFormatter);
+			lbProperties.RegisterProperty(TemplateTags.FileDate, lb => lb.FileDate, DateTimeFormatter);
+			return lbProperties;
+		}
```
Mbucari, line 232: > RegisterProperty: Where is this called?

Here.

rmcrackan: > Large diffs are not rendered by default.

Thanks, github. This is why I couldn't find it. I'd thought: he either did something very clever, or **too** clever

Mbucari: > or **too** clever

I try not to do too much of that. Doing something really clever can very fun/satisfying, but then I remember this quote:

> Debugging is twice as hard as writing the code in the first place. Therefore, if you write the code as cleverly as possible, you are, by definition, not smart enough to debug it.

rmcrackan:
There's nothing wrong with PropertyTags being a list. Since you're keying on TagName though, you might find value in having it be a dictionary with a TagName key.

Superfluous future enhancement:

At it's heart `TagClass` is a management module for the PropertyTags. We could probably make it IEnumerable then PropertyTagClass could have an Add which delegates to RegisterProperty. This would let you streamline GetFilePropertyTags

```
PropertyTagClass<LibraryBookDto> lbProperties = new() {
    new(TemplateTags.Id, lb => lb.AudibleProductId))
};
```

Alternatively, you might be able to introduce some cleverness into RegisterProperty as RegisterProperty<T> and determine your formatter based on `T`

> Absolutely. It's gotten to be just a bit too sophisticated to expect users to muddle their way through blind.

Thanks!

> Add conditional negation

Another operator or 2 and we can make this Turing complete 🤣

Source/LibationFileManager/Templates.cs
rmcrackan: > Debugging is twice as hard as writing the code in the first place. Therefore, if you write the code as cleverly as possible, you are, by definition, not smart enough to debug it.

That's a good thing to remember. Another favorite:

> Always code as if the guy who ends up maintaining your code will be a violent psychopath who knows where you live.

The problem here is: that coder was me 4 months ago. He lives 4 months ago where I can't reach him. Fortunately, I know how to ruin that asshole's future.

Mbucari:
> Since you're keying on TagName though, you might find value in having it be a dictionary with a TagName key.

The reason it's not a dictionary is because when parsing the template string, we don't know if the string starts with a tag or not, and if it does, we don't know how long the tag string is. The RegEx is responsible for answering those questions. Unless you tokenize the whole template string first (which is basically a rewrite of the whole parser), the only way to match tags is to just loop through all `PropertyTags` and see if there's a match.

> . We could probably make it IEnumerable then PropertyTagClass could have an Add which delegates to RegisterProperty.

Interesting thought. When I began this project I was starting with an expression parser that I'd written a couple of years ago. In that version, the TagClass equivalent was an ICollection<T>. Maybe we could do that here too. I'll have a think on it.

rmcrackan:
> we don't know if the string starts with a tag or not

Good point.

Is this PR ready for merge?

Mbucari:
> Is this PR ready for merge?

Yes, it's ready. I could probably keep tinkering with it forever, but right now it is functional and all tests are passing.

rmcrackan:
Oh yeah -- documentation. Then that should be it.

Mbucari:
> documentation.

[Done](https://github.com/Mbucari/Libation/blob/master/Documentation/NamingTemplates.md)

Mbucari:
> . We could probably make it IEnumerable then PropertyTagClass could have an Add which delegates to RegisterProperty. This would let you streamline GetFilePropertyTags

You're right, [this is so much sexier!](https://github.com/Mbucari/Libation/blob/c06c1447b3f5a68972db7a4b04ae7b0d7068a924/Source/LibationFileManager/Templates.cs#L192-L239)

rmcrackan:
Noice -- that is sexy. I don't know if that syntax is flexible enough to support this to eliminate the formatter field. It'd be even cleaner though.

```
Add<T>(Tag tag, Func<T> func)
{
    Formatter formatter;
    if (typeof(T) == typeof(string))
        formatter = StringFormatter;
}
```

rmcrackan:
Or different overloads

```
Add(Tag tag, Func<string> func) => addInternal(tag, func, StringFormatter);
Add(Tag tag, Func<int> func) => addInternal(tag, func, IntegerFormatter);
```

Mbucari:
> I don't know if that syntax is flexible enough to support this to eliminate the formatter field

That would only work if the string formatter lived inside of FileManager. My idea for how this whole thing works is that NamingTemplate is format-agnostic. Only the caller knows how values get formatted.

But I'm also a wee bit drunk and may not be understanding your suggestion completely LOL

Mbucari:
> Noice -- that is sexy.

I'm convinced that ~90% of my code quality can be attributed to my sense of aesthetics. If it doesn't look pretty to me then it needs to be refactored.

Issue #485, opened 2023-02-10
Rename and save FileLocations.json instead of deleting it.


Andytr1:
See issue #483 for background. It appears that the error handling deletes FileLocations.json even if it may not be corrupt. Since this file "links" the physical location of audiobooks with the UI, it would be helpful to rename it and keep it in file system, while alerting the user.

rmcrackan:
The fix is [here](https://github.com/rmcrackan/Libation/pull/487#issuecomment-1426613423). Available in [latest](https://github.com/rmcrackan/Libation/releases/latest)

Issue #483, opened 2023-02-09
after installing latest update, unable to open folder with existing books [from email]


rmcrackan:
This was sent to my personal email. Put here for visibility, triage, and tracking:

I noticed the following issue two updates ago.  
After installing the update, I click on the green button in the Liberate column.  I get the "File not found" error message instead of having the explorer open to the location of the book.  
No paths were changed and the install directory remained the same. The update was performed from the automatic prompt.

I would appreciate any insight you can provide.

Mbucari:
The first obvious question is if this is an ongoing problem them, or if something changed once and made his whole FileLocations.json useless. I noticed a couple of releases ago that FileLocations was broken because I had added a second constructor to `LongPath and didn't mark one of them as a `[JsonConstructor]`. This caused the json parser to throw an exception which caused Libation to delete the existing FileLocations.json and make a new one.

Andytr1:
Original reporter - Just got source code and stepping through. It appears that the regex generated with productId is not matching anything. 
It's happening in BackgroundFileSystem.cs_FindFiles()::44. This only happens for books that were liberated before the patch. Any books that were imported after update are working fine.
I wonder if the fsCache is not being populated correctly.

Mbucari:
@Andytr1 Can you please email your `FileLocations.json` file to me mbucari1@gmail.com. When you do, can you also give me one example of a book that is found and one example of a book that isn't found?

Andytr1:
You are right. Looks like FileLocations.json gets deleted/re-created. I am looking at it and it only has the books liberated after upgrade and a lot of other books are missing from it.
Is there a way to regenerate it?

Mbucari:
@Andytr1 If your audio files of the folder contain those files has the audible product ID in the name (e.g. \[B00DJ8JAY2\]), Libation should find them. If you didn't use the \<id\> naming template, I believe you're out of luck.

@rmcrackan please correct me if I'm wrong.

Andytr1:
I didn't use the option of having product ID in the name. I guess I will manually have to find a way to crawl the database/audio book files to recreate the FileLocations.json.

Mbucari:
> I didn't use the option of having product ID in the name. I guess I will manually have to find a way to crawl the database/audio book files to recreate the FileLocations.json.

If your bandwidth is uncapped, the path of least resistance might be to just re-download them. I'm sorry for the error :/

Andytr1:
[...]

Issue #481, opened 2023-02-09
Custom Naming Fields for Author First or Last Name Only


schmads42:
I prefer to organize my books by the author's last name, followed by the author's first name, i.e.:  "Doyle, Arthur Conan".

I'd like two new tags, `<author first name>` and `<author last name>` which would return only the last name, and any names that are not the last name, respectively.

A single tag that returns the name in Last, First format would also work, but wouldn't be as flexible for others who may want only the last or first names.

I have already made the changes on a clone of the repo, locally.  I tried to push a branch to do a PR, but I am not permitted.  I will be happy to submit my changes for a pull-request if that's easy for you, but I'll also attach a diff.  Any whitespace changes were unintentional.

I have tested the changes on Windows 10 x64 only, but they shouldn't be platform-specific.

[SepNameFieldsDiff.txt](https://github.com/rmcrackan/Libation/files/10691576/SepNameFieldsDiff.txt)

Libation rocks!  Feel free to use or discard this as you like.

rmcrackan:
Thanks much for your idea and for your code. I've considered this before and unfortunately the [edge cases](https://github.com/rmcrackan/Libation/issues/420#issuecomment-1361311573) are just too numerous. (I wish github made searching easier; it took me a while to find this even knowing it existed)

This has been a popular request with at least 3 tickets so far. I'll keep a look out and hope that audible improves their metadata.

@Mbucari I'm interested in your opinion. Am I over-thinking this? Or are these good reasons to not create such a feature?

Mbucari:
> @Mbucari I'm interested in your opinion. Am I over-thinking this? Or are these good reasons to not create such a feature?

I think it's worth a try. I think the worst case scenario is they authors's name parts get jumbled and filenames will require manual correction, but the people who want this feature are probably already manually editing filenames. I recently took a look at all the author and narrator names in my library and have a couple of ideas for how to be a little clever about it and avoid some of the pitfalls you outlined. Would you mind exporting your library and sharing it with me so I could see some more examples?

And I just had a crazy idea. What if we used the OpenAI service to do the name parsing for us?

CharlieRussel:
If you need a library or 4, I have some large ones. Including pre-Amazon and a UK one. Ping me in email and I'll share.

rmcrackan:
> Oh! I’m sorry for the duplicate request

No worries. Github's defaults assume you're searching open tickets only. You have to trick it a bit to also search closed too.

[...]

PR #487, opened 2023-02-10
Custom author and narrator names formatting and batch locate books

Mbucari:
First, the big one: Custom name formatting for authors and narrators. This was actually pretty darned easy to implement with NameParserSharp. Names are parsed into \{Prefix, First, Middle, Last, Suffix\}. By default, the \<author\> tag behaves exactly as it already does: all authors concatenated with ", ". The options are format(), sort(), separator(), and max(). Users can use any combination of those virtual functions, but only the first occurrence of each one is used.

I also added an audiobook scanner to locate liberated m4b audiobooks and add them to FilePathCache and set the book's status to Liberated (#485).


Source/LibationFileManager/Templates.cs
```diff
@@ -247,6 +251,70 @@ private static List<List<string>> GetPathParts(IEnumerable<string> templateParts
 
 		#region Tag Formatters
 
+		private static readonly string[] suffixes = { "introductions", "introduction", "adaptation", "translator", "contributor", "illustrator", "director", "foreword", "editor", "preface", "adaptor", "afterword", "interviewer", "introductions", "essay", "editor/introduction" };
```
rmcrackan, line 254: If these are all 'space hyphen space annotation', we should split on the pattern instead of trying to enumerate. Audible will inevitably have more of these later.

rmcrackan:
This looks amazing. Love the updates to documentation too.

My only request is: can there by more unit tests for how we expect names to be parsed? It could just be some of the [examples from yesterday](https://github.com/rmcrackan/Libation/issues/481#issuecomment-1424627883) including with some of the [suffixes you identified](https://github.com/rmcrackan/Libation/issues/481#issuecomment-1424601816). I don't expect perfection. It'd be nice to see plainly what we do expect though. Also, how do you think we should handle 1 word names? It feels like they should have special handling to match user expectations:

* first name: Jewel
* last name: Jewel
* middle name: *null* or ""
* full name: Jewel
* last name comma first name: Jewel

Source/LibationFileManager/Templates.cs
Mbucari: Done.

Mbucari:
> Can there by more unit tests for how we expect names to be parsed?

I added a lot more.
> Also, how do you think we should handle 1 word names? It feels like they should have special handling to match user expectations:

1-word names are parsed as first names. For names without a last name, I set the last name equal to the first name.

Source/LibationFileManager/Templates.cs
```diff
@@ -68,6 +70,9 @@ static Templates()
 			Configuration.Instance.PropertyChanged +=
 				[PropertyChangeFilter(nameof(Configuration.ChapterTitleTemplate))]
 				(_, e) => _chapterTitle = GetTemplate<ChapterTitleTemplate>((string)e.NewValue);
+
+			HumanName.Suffixes.Add("ret");
+			HumanName.Titles.Add("professor");
```
Mbucari, line 75: Keep this in mind when confronted with edge cases in the future.

Mbucari:
I also replaced readonly Regex instances with the new `GeneratedRegexAttribute`. It's pretty schwifty!

Source/LibationFileManager/Templates.cs
```diff
@@ -247,6 +252,89 @@ private static List<List<string>> GetPathParts(IEnumerable<string> templateParts
...
+				nameFormatString
+					= nameFormatString
+					.Replace("{T}", "{0}")
+					.Replace("{F}", "{1}")
+					.Replace("{M}", "{2}")
+					.Replace("{L}", "{3}")
+					.Replace("{S}", "{4}");
+
+				return string.Format(nameFormatString, humanName.Title, humanName.First, humanName.Middle, lastName, humanName.Suffix).Trim();
+			}
```
Mbucari, line 335: I don't think we ever discussed local functions or how you feel about them. I like them because they make methods more readable while maintaining encapsulation, but I don't think I've ever seen you use them. What are your thoughts?

rmcrackan: I love them in moderation. They're something I've long wished for without knowing what to call them. If there are more than a few, it might be an indication that it's time for a new class. That's a case-by-case judgement call.

rmcrackan:
last question: does this close out #481 , #483 , #485 ?

Mbucari:
> #483, #485

Yes, this solves his problem. To re-add books to his library, he needs to go to Import > Locate Audiobooks and select a folder to scan.

For #481
- `<author[format({L}) max(1)]>` will retrieve the first author's last name.
- `<author[format({T} {F} {M} {S}) max(1)]>` will retrieve the first author's first name and middle name, prefixes, and suffixes (if any).

Issue #476, opened 2023-02-07
Issue installing Libation on a Mac


FTime8:
Hi,

I am trying to install the Mac version of Libation. I am hoping you can help here.

I followed the step by step instructions in the embedded video regarding installing on a Mac. I was able to get to the step where I select New User, but after selecting it the terminal pops up again with the below and nothing progresses.

![Libation progress halt](https://user-images.githubusercontent.com/124659843/217201924-7665c6b7-600f-4a41-8a60-91f01f48aaa6.png)

colossal-squid:
<img width="489" alt="image" src="https://user-images.githubusercontent.com/32293829/217513673-29085697-0eeb-4b22-863c-bb66a1b035f2.png">
same here

Mbucari:
Hi @FTime8 Can you please post the log file? It's located in [user]/Libation (~/Libation)

rmcrackan:
Fixed by @Mbucari in [version 9.3.1](https://github.com/rmcrackan/Libation/releases/latest)

FTime8:
Sorry @Mbucari, I gave up over the weekend before the update & deleted everything and backed up my audiobooks on my friends Windows based computer to an external drive.

PR #489, opened 2023-02-13
Mac App Bundle and added mp3 conversion support on mac

Mbucari:
Everything in this PR is for Mac users.

- The release workflow now builds the macos app bundle.
- Updated AAXClean so that mp3 conversion now works on mac.
- Changed how appsettings.json is searched for and created.
  - On load, Libation will search for appsettings.json in the following locations:
    1. App directory
    2. %localappdata%\Libation\
    3. %appdata%\Libation\
    4. %temp%\
  - If it's not found in any of those folders, Libation will try to create it starting at (a.) If it failes to create it there, it tries again in (b.) and so on. Eventually at (d.) it will be created in %temp%. That should never happen.


Mbucari:
Here's a new video

[](https://user-images.githubusercontent.com/37587114/218372842-e4f7bc5f-4d9d-463b-a59a-7482119bac8e.mp4)

rmcrackan:
I'm increasingly impressed by the maturity of the linux and mac versions. I spent years writing the backend in hopes that eventually this could happen but one part of my brain never actually thought the day would come. Wow!

1) RE appsettings.json: Crucial -- the same logic needs to make its way to `UNSAFE_MigrationHelper` or else future migration methods won't work.

I've kept the logic in UNSAFE_MigrationHelper separate from production code because UNSAFE_MigrationHelper by definition can be run on an out of date system for the purposes of bringing it up to date. This means it can't rely on the up to date properties, methods, or logic in `Configuration`. As the all-caps name implies, be careful playing around in there. Errors or untested mistakes could brick the user's local config.

2) Is this PR addressing #476 ?

3) Your new video: I assume this just needs to be updated in InstallOnMac.md . Minor request: **if** it's not too much trouble, can this be recorded showing the main github page instead of your? If it's a hassle, just leave it. I doubt many people will ever notice.

Mbucari:
> I'm increasingly impressed by the maturity of the linux and mac versions.

Same here. Just the other day I was reflecting on how much Libation has changed since I first saw it. It's really come a long way!

> RE appsettings.json: Crucial -- the same logic needs to make its way to `UNSAFE_MigrationHelper`

Good call. I think that class is one of the few places in Libation that I've never poked my nose into. 
I think I'm going to make `Configuration.APPSETTINGS_JSON` public.

> Is this PR addressing [Issue installing Libation on a Mac #476](https://github.com/rmcrackan/Libation/issues/476) ?

It was not designed to, and I have no idea if it will help since I have no idea what his problem is. He should probably post a log file.

> can this be recorded showing the main github page instead of your?

Sure thing. Don't want to confuse people more than they already always seem to be xD

<hr>

I'm converting this to draft because I want to investigate something else on mac. Right now I'm just not sure how users can make use of Hangover or cli. I think it will come down to bundle permissions, but I need to figure that out so we can add it to the wiki.

rmcrackan:
Tagging #490 here so I remember to close it after the PR

Mbucari:
Ready for review.

Changes since your last review:

- UNSAFE_MigrationHelper uses `Configuration.AppsettingsJsonFile`
- Updated Dinah.Core to v7.2.1.1
- Fixed a bug in HangoverAvalonia where query output wouldn't display
- HandoverWinForms now compiles to Hangover.exe
- Added a new menu item Settings > Launch Hangover
- Right click > remove book now executes `LibraryCommands.RemoveBook` asynchronously
- Moved NameListFormatter to its own class. Your comment from the last PR made me think it deserved it.
- Added better support for Hangover and LibationCli in MacOS


I'll update documentation after you merge and release so I can capture video showing your repo.

Source/LibationAvalonia/Program.cs
```diff
@@ -13,8 +14,25 @@ namespace LibationAvalonia
...
+				return;
+			}
+			if (Configuration.IsMacOs && args != null && args.Length != 0 && args[0] == "cli")
+			{
+				//Open a new Terminal in the sandbox
+				Assembly asm2 = Assembly.GetExecutingAssembly();
+				string libationProgramFiles = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+				Process.Start("/System/Applications/Utilities/Terminal.app/Contents/MacOS/Terminal", $"\"{libationProgramFiles}\"");
+				return;
+			}
```
Mbucari, line 34: We can do this because we're already executing inside the sandbox. Any process created in the sandbox executes in the same sandbox. Unfortunately, all sandbox files are read/execute, so no writing!

Mbucari:
Converting back to draft because I think I figured out how to do automatic updates!

rmcrackan:
> NameListFormatter

Good idea. We'll want tight control over this. I'm sure there are future tweaks and having them here will make them significantly easier.

It's bittersweet to see the changes you're making in Configuration which have been there nearly since day 1. My little baby's growing up.

Mbucari:
Holy crap, what a ride this has been. I learned a lot about mac that I wish I had not had to! But the payoff is great.

Added auto-updater for macOS and Linux!
Caveats:
- Mac updater only works if they user is running in /Applocations. This shouldn't be a problem.
- Linux updater only works if users installed Libation with the .deb package.

I also __probably_ fixed #492. It's hard to tell because it's so hard to reproduce.
<hr>
**Next steps**

Apple is moving towards arm, so we should deploy an arm build. Building it will be trivial, but we'll need to figure out the workflows. I really want to get that working soon as a thank you to @hutattedonmyarm, but I don't have it in me right now. Also, we probably shouldn't continue releasing zipped binaries for linux and mac anymore. Those OS's have bundles and that's what everyone should be using.

.releaseindex.json
```diff
@@ -1,6 +1,6 @@
 {
   "WindowsClassic":  "Libation\\.\\d+\\.\\d+\\.\\d+-win(dows)?-classic\\.zip",
   "WindowsAvalonia": "Libation\\.\\d+\\.\\d+\\.\\d+-win(dows)?-chardonnay\\.zip",
-  "LinuxAvalonia":   "Libation\\.\\d+\\.\\d+\\.\\d+-linux-chardonnay",
-  "MacOSAvalonia":   "Libation\\.\\d+\\.\\d+\\.\\d+-macos-chardonnay"
+  "LinuxAvalonia":   "Libation\\.\\d+\\.\\d+\\.\\d+-linux-chardonnay\\.deb",
+  "MacOSAvalonia":   "Libation\\.app-x64-\\d+\\.\\d+\\.\\d+\\.tgz"
```
rmcrackan, line 5: Can we still have 'macos' in here somewhere? I assume this is the package name that will be seen by downloaders and the app and tgz don't make it immediately obvious that this is for mac.

Source/HangoverWinForms/HangoverWinForms.csproj
```diff
@@ -4,6 +4,7 @@
     <OutputType>WinExe</OutputType>
     <TargetFramework>net7.0-windows</TargetFramework>
     <EnableWindowsTargeting>true</EnableWindowsTargeting>
+	<AssemblyName>Hangover</AssemblyName>
```
rmcrackan, line 7: This is a good idea

Source/LoadByOS/WindowsConfigApp/WinInterop.cs
```diff
@@ -35,8 +36,31 @@ public void SetFolderIcon(string image, string directory)
...
+			var zipExtractor = Path.Combine(Path.GetTempPath(), "ZipExtractor.exe");
 
-        public void CopyTextToClipboard(string text)
-            => Clipboard.SetText(text);
+			File.Copy("ZipExtractor.exe", zipExtractor, overwrite: true);
+
+			RunAsRoot(zipExtractor, $"--input \"{updateBundle}\" --output \"{thisDir}\" --executable \"{thisExe}\"");
+		}
+
+		public Process RunAsRoot(string exe, string args)
```
rmcrackan, line 51: I tried to do something like this with Dinah.Core.WindowsDesktop.Processes.Runner but I think I made mine too restrictive

Source/LoadByOS/LinuxConfigApp/LinuxInterop.cs
```diff
@@ -1,14 +1,73 @@
 ﻿using LibationFileManager;
+using System.Diagnostics;
 
 namespace LinuxConfigApp
 {
     internal class LinuxInterop : IInteropFunctions
     {
-        public LinuxInterop() { }
+		//Different terminal apps possibly installed on a linux system
```
rmcrackan, line 8: Yay! So validating to see 'just in case' infrastructure get proper use. When I added these interop files last year I didn't know if the mac or linux ones would actually get used.

Source/LibationWinForms/GridView/ProductsGrid.cs
```diff
@@ -122,7 +123,7 @@ private void gridEntryDataGridView_CellContextMenuStripNeeded(object sender, Dat
                     {
                         var dgv = (DataGridView)sender;
                         var text = dgv[e.ColumnIndex, e.RowIndex].FormattedValue.ToString();
-                        InteropFactory.Create().CopyTextToClipboard(text);
+						Clipboard.SetDataObject(text, false, 5, 150);
```
rmcrackan, line 126: Interesting. I jumped though these hoops because I didn't think Clipboard was xplat

rmcrackan:
This looks fantastic! Any final changes or words of wisdom before I hit 'merge'? Also confirming: am I correct that this PR closes these tickets:

* #476
* #490 
* #492

.releaseindex.json
Mbucari: Done!

Source/LoadByOS/LinuxConfigApp/LinuxInterop.cs
Mbucari: I felt the same way about my addition of `.releaseindex.json` lol

Source/LibationWinForms/GridView/ProductsGrid.cs
Mbucari: Yup! It's a gui-only feature though, so you need Avalonia to do it.

Documentation/InstallOnMac.md
```diff
@@ -4,40 +4,36 @@
...
-```console
-chmod +x ./Libation
-sudo spctl --add --label "Libation" ./Libation
-./Libation
-```
-
-## Trouble with Gatekeeper?
-
-If Gatekeeper is giving you trouble with Libation:
+- Download the `Libation.app-x64-x.x.x.tgz` file from the latest release and extract it.
```
rmcrackan, line 12: Thanks for re-adding 'macos'. Here also, plz

Mbucari:
> Any final changes or words of wisdom before I hit 'merge'?

Probably, but I'm too tired to think of any lol. As usual, the more features you have the more there is to go wrong. I'm curious/dreading to see how this will go wrong in the wild.

> Also confirming: am I correct that this PR closes these tickets:

Yup, I think that's all of them.

Documentation/InstallOnMac.md
Mbucari: Thanks. Done.

Issue #492, opened 2023-02-14
Unhandled Exception Crash


chgi:
Hello there, I just found this project and tried it out. Great stuff! Thank you!

**Describe the bug**
I got a crash that may be related to non-threadsafe behaviour in Avalonia.
This didn't leave an entry in the log file, but I got the output below on the console.
After a little searching, I think it might be related to to <https://github.com/AvaloniaUI/Avalonia/issues/8810>.
I am not familiar with Avalonia, but it seems to me this could be fixed by 
dispatching `RunningTime = timeToStr(DateTime.Now - StartingTime);` to the UI thread in `Source/LibationAvalonia/ViewModels/ProcessQueueViewModel.cs:256` instead of updating it directly from the timer thread.

```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Avalonia.Utilities.WeakHashList`1.GetAlive(Func`1 factory) in /_/src/Avalonia.Base/Utilities/WeakHashList.cs:line 239
   at Avalonia.Utilities.WeakEvent`2.Subscription.OnEvent(Object sender, TEventArgs eventArgs) in /_/src/Avalonia.Base/Utilities/WeakEvent.cs:line 157
   at System.Reactive.AnonymousSafeObserver`1.OnNext(T value) in /_/Rx.NET/Source/src/System.Reactive/AnonymousSafeObserver.cs:line 54
   at System.Reactive.Linq.ObservableImpl.SelectMany`2.EnumerableSelector._.OnNext(TSource value) in /_/Rx.NET/Source/src/System.Reactive/Linq/Observable/SelectMany.cs:line 1379
   at System.Reactive.Linq.ObservableImpl.Buffer`2.Boundaries._.BufferClosingObserver.OnNext(TBufferClosing value) in /_/Rx.NET/Source/src/System.Reactive/Linq/Observable/Buffer.cs:line 829
   at System.Reactive.Linq.ObservableImpl.Merge`1.Observables._.InnerObserver.OnNext(TSource value) in /_/Rx.NET/Source/src/System.Reactive/Linq/Observable/Merge.cs:line 240
   at System.Reactive.Linq.ObservableImpl.Select`2.Selector._.OnNext(TSource value) in /_/Rx.NET/Source/src/System.Reactive/Linq/Observable/Select.cs:line 48
[...]

PR #503, opened 2023-02-18
Mac and Linux Arm64 releases and Fixed #502

### LINKED ISSUE #502: Strip Audible + Split Chapters Still Truncating Chapters
**Describe the bug**
This is a followup to #494.

The beginning of the first file is no longer clipped but the chapter divisions are still off. 

I have not had time to investigate like I did with the previous but it _seems_ like the amount of time that the first file is extended, cutting into the second chapter is about the same length as the "This is Audible" intro (if I had to guess).

Example: https://www.audible.com/pd/Discussing-Design-Audiobook/B08WH2MMMF 

1. Settings > Audio File Serttings
2. Enable Split my books into multiple files by chapter.
3. Enable Strip audible branding from the start and end of audiobooks
4. Download a book or 200.
5. Play _second_ or subsequent track(s)

Mbucari:
There's a lot of little things in here, but the headlines are 

- Fixed #502 by upgrading AAXClean.
- Redesigned Linux/Mac workflows
  - Both x64 and arm64 are now built
  - Zipped bins are no longer uploaded, only Deb for Linux and app bundles for Mac
  - Added minimum version limits to the Mac app bundles so Mac will notify the user that their machine is not supported.
- Added ReleaseIdentifiers `LinuxAvalonia_Arm64` and `MacOSAvalonia_Arm64`
- Fixed the DirectorySelectControl on Avalonia (again. I feel like this keeps breaking)
- Added some more UI thread safety to address #492
- Added `Configuration.ProcessDirectory`. Mac and Linux can't rely on Environment.ProcessPath because unlike in Windows, dotnet is the process.
- Moved `runasroot.sh` to the project folder and copy it to the output (don't know why I added it as a resource to begin with).
- Removed unnecessary and unused code InteropFactory and the config apps.
- Fixed a couple of mac bugs (I'm pleased to learn that .net assemblies are truly portable. I can just compile a new debug change on my windows host, copy it to the mac VM, and run it.)
- Added some svgs. Only the main glass is used for now (by the Linux app), but the rest are there for whatever (I was having fun).

I tested this PR extensively, including the auto update, on all 3 platforms. It works well.


Source/AppScaffolding/LibationScaffolding.cs
```diff
@@ -296,21 +298,20 @@ private static (Octokit.Release, Octokit.ReleaseAsset) getLatestRelease(TimeSpan
...
 			//Download the release index
 			var bts = await gitHubClient.Repository.Content.GetRawContent(ownerAccount, repoName, ".releaseindex.json");
 			var releaseIndex = JObject.Parse(System.Text.Encoding.ASCII.GetString(bts));
 			var regexPattern = releaseIndex.Value<string>(ReleaseIdentifier.ToString());
+			var regex = new System.Text.RegularExpressions.Regex(regexPattern, System.Text.RegularExpressions.RegexOptions.IgnoreCase);
 
-			// https://octokitnet.readthedocs.io/en/latest/releases/
-			var releases = await gitHubClient.Repository.Release.GetAll(ownerAccount, repoName);
+			//https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release
+			var latestRelease = await gitHubClient.Repository.Release.GetLatest(ownerAccount, repoName);
```
Mbucari, line 313: Unless you make a new release with a lower version number than a previous release, this is equivalent (with >200 fewer API calls).

Source/LoadByOS/WindowsConfigApp/Program.cs
```diff
@@ -1,18 +1,7 @@
...
+    class Program
     {
-        public override Type InteropFunctionsType => typeof(WinInterop);
-        public override Type[] ReferencedTypes => new Type[]
-        {
-            typeof(Bitmap),
-            typeof(Dinah.Core.WindowsDesktop.GitClient),
-            typeof(Accessibility.IAccIdentity),
-            typeof(Microsoft.Win32.SystemEvents)
-        };
```
Mbucari, line 14: This was necessary when we were executing the config app to get a list of its process modules, but we're not doing that so we don't need the runtime to load the types. In fact, we don't need the launcher exes at all which is why I modified the workflows to delete them.

Mbucari: As a reminder, the only reason we do not have to execute the config app to load its dependencies is because we are publishing it as self-contained, so all dependencies are in the same folder and we can just load them by file name. If we ever published not self-contained (framework dependent), we would need the run time to locate the assemblies for us because they'll be somewhere in program files or system.

This is as much a reminder to me as it is to you.

Source/AppScaffolding/LibationScaffolding.cs
rmcrackan: Woah! That's a large difference -- noice

Source/LoadByOS/WindowsConfigApp/Program.cs
rmcrackan: > If we ever published not self-contained

Seems like this could be left in-code. Move to an un-called method and put these comments in it. Otherwise if we do ever need it, I'll be going: oh crap, I remember we removed in one of those PRs -- now how did we do this again...?

.releaseindex.json
```diff
@@ -1,6 +1,8 @@
 {
-  "WindowsClassic":  "Libation\\.\\d+\\.\\d+\\.\\d+-win(dows)?-classic\\.zip",
-  "WindowsAvalonia": "Libation\\.\\d+\\.\\d+\\.\\d+-win(dows)?-chardonnay\\.zip",
-  "LinuxAvalonia":   "Libation\\.\\d+\\.\\d+\\.\\d+-linux-chardonnay\\.deb",
-  "MacOSAvalonia":   "Libation\\.app-macOS-x64-\\d+\\.\\d+\\.\\d+\\.tgz"
+  "WindowsClassic":      "Libation\\.\\d+\\.\\d+\\.\\d+-win(dows)?-classic\\.zip",
+  "WindowsAvalonia":     "Libation\\.\\d+\\.\\d+\\.\\d+-win(dows)?-chardonnay\\.zip",
+  "LinuxAvalonia":       "Libation\\.\\d+\\.\\d+\\.\\d+-linux-amd64\\.deb",
```
rmcrackan, line 4: I liked having the Chardonnay branding evident in the download file names. I'm often asked the difference between Classic and Chardonnay and it's easy to reply that Chardonnay is available on all platforms. Without it in the download name, I think this explanation is confusing.

rmcrackan:
svg.s -- 'cheers' and 'slosh' are cute and very much on-brand. How do you see them being used?

rmcrackan:
I have to say, you've made amazing and massive changes over the past year. I'm also pretty proud of my software paradigm/architecture which has gone largely unchanged throughout the process even with such edits and additions. It's also mostly unseated what was the only game in town -- open audible. We have a lot to be proud of here.

Source/AppScaffolding/LibationScaffolding.cs
```diff
@@ -35,7 +37,7 @@ public static class LibationScaffolding
 		public static ReleaseIdentifier ReleaseIdentifier { get; private set; }
 		public static VarietyType Variety
 			=> ReleaseIdentifier == ReleaseIdentifier.WindowsClassic ? VarietyType.Classic
-			: ReleaseIdentifier.In(ReleaseIdentifier.WindowsAvalonia, ReleaseIdentifier.LinuxAvalonia, ReleaseIdentifier.MacOSAvalonia) ? VarietyType.Chardonnay
+			: Enum.IsDefined(ReleaseIdentifier) ? VarietyType.Chardonnay
```
rmcrackan, line 40: In theory, this will also catch `None`. I don't know if it's possible in practice though.

Source/LibationAvalonia/Program.cs
```diff
@@ -55,9 +53,25 @@ static void Main(string[] args)
 			if (Configuration.IsWindows)
 				AppScaffolding.LibationScaffolding.SetReleaseIdentifier(AppScaffolding.ReleaseIdentifier.WindowsAvalonia);
 			else if (Configuration.IsLinux)
-				AppScaffolding.LibationScaffolding.SetReleaseIdentifier(AppScaffolding.ReleaseIdentifier.LinuxAvalonia);
+			{
```
rmcrackan, line 56: This could be cleaned up by refactoring to the OS specific libraries. If you don't think it would divorce it too much from the logic flow here. To me, this is a lot cleaner and equally expressive. Just a matter of taste though; not a mandate.

    AppScaffolding.LibationScaffolding.SetReleaseIdentifier(osconfig.GetReleaseIdentifier());

Source/LibationFileManager/InteropFactory.cs
```diff
@@ -52,94 +58,36 @@ static InteropFactory()
...
-                Directory.EnumerateFiles(here, $"*{CONFIG_APP_ENDING}", SearchOption.TopDirectoryOnly)
-                // sanity check. shouldn't ever be true
-                .Except(new[] { Environment.ProcessPath })
+                Directory.EnumerateFiles(Configuration.ProcessDirectory, $"*{CONFIG_APP_ENDING}", enumerationOptions)
                 .FirstOrDefault(exe => MatchesOS(exe));
 
             return appName;
         }
 
-        /*
```
rmcrackan, line 89: Yeah, these were such hard-won knowledge that it feels weird to discard them wholesale. I'd like to keep the knowledge ... somewhere.

rmcrackan:
Overall this looks fantastic. I'd like to tug on those conversation threads a bit before a merge.

Source/LoadByOS/WindowsConfigApp/Program.cs
Mbucari: I didn't remove it from the _Demo projects. Is that in-code enough? I'll add my comments in this thread to the code too.

Mbucari:
> svg.s -- 'cheers' and 'slosh' are cute and very much on-brand. How do you see them being used?

Not one bit. I was just fooling around with the SVGs, having a bit of fun. I never realized what a versitile format it is. You can duplicate objects and trtansform them all within the markup. You can even add "glow", though I didn't.

Mbucari:
> I have to say, you've made amazing and massive changes over the past year.

Thank you. I'm really quite proud too. I never used OpenAudible before; I strictly used ffmpeg command line to remove drm. In some ways I'm actually kind of a luddite.

.releaseindex.json
Mbucari: I added chardonnay to the built filenames.

Source/AppScaffolding/LibationScaffolding.cs
Mbucari: OK, I reworked this a little. I added an OS enum to `Configuration`, and `Configuration.OS`.

The `ReleaseIdentifier` is now a the sum of the `OS`, `Variety` (renamed from `VarietyType`), and `Architecture` enum values.

`SetReleaseIdentifier` only takes `Variety` as an argument and figures out the release ID from that and the OS/arch.

Source/LibationFileManager/InteropFactory.cs
Mbucari: Is keeping it in the demos sufficient?

Mbucari:
Hold on, something I just changed caused a problem on linux.

Mbucari:
OK, I fixed it. Merge away!

Mbucari:
Fixed the releaseindex.json again. Fully tested and working.

Source/LoadByOS/WindowsConfigApp/Program.cs
rmcrackan: Oh yeah -- great point

rmcrackan:
> I never used OpenAudible before; I strictly used ffmpeg command line to remove drm

Back before they started charging, OpenAudible was a free app which was easy to do java reflection on. I learned a lot and changed my library import from web scraping to the same ajax calls that OA used. (before discovering mkb79's amazing api). Back then (and probably still now) all they did was marshal calls to ffmpeg. Since they are xplat though, they just bundled the right ffmpeg each installer.

Source/AppScaffolding/LibationScaffolding.cs
```diff
@@ -18,30 +18,45 @@ namespace AppScaffolding
 	public enum ReleaseIdentifier
 	{
 		None,
-		WindowsClassic,
-		WindowsAvalonia,
-		LinuxAvalonia,
-		MacOSAvalonia,
-		LinuxAvalonia_Arm64,
-		MacOSAvalonia_Arm64
+		WindowsClassic = OS.Windows | Variety.Classic | Architecture.X64,
```
rmcrackan, line 21: Clever!

.releaseindex.json
```diff
@@ -1,8 +1,8 @@
 {
   "WindowsClassic":      "Libation\\.\\d+\\.\\d+\\.\\d+-win(dows)?-classic\\.zip",
   "WindowsAvalonia":     "Libation\\.\\d+\\.\\d+\\.\\d+-win(dows)?-chardonnay\\.zip",
-  "LinuxAvalonia":       "Libation\\.\\d+\\.\\d+\\.\\d+-linux-amd64\\.deb",
-  "MacOSAvalonia":       "Libation\\.\\d+\\.\\d+\\.\\d+-macOS-x64\\.tgz",
-  "LinuxAvalonia_Arm64": "Libation\\.\\d+\\.\\d+\\.\\d+-linux-arm64\\.deb",
-  "MacOSAvalonia_Arm64": "Libation\\.\\d+\\.\\d+\\.\\d+-macOS-arm64\\.tgz"
+  "LinuxAvalonia":       "Libation\\.\\d+\\.\\d+\\.\\d+-linux-chardonnay-amd64\\.deb",
```
rmcrackan, line 4: gracias!

Source/LibationFileManager/InteropFactory.cs
rmcrackan: yup

PR #509, opened 2023-02-24
Add More MP3 Options and improved AAXClean

Mbucari:
- AAXClean now has a lower memory footprint
- M4b chapters are written at the front of the file instead of at the end, so streaming performance should improve.
- Added mp3 file chapter markers
- Added settings to control mp3 encoder quality and sample rate.
- Added context menu to convert individual books to mp3.

### DIFF
--- a/Source/AaxDecrypter/AaxDecrypter.csproj
+++ b/Source/AaxDecrypter/AaxDecrypter.csproj
@@ -13,7 +13,7 @@
 	</PropertyGroup>
 
 	<ItemGroup>
-	  <PackageReference Include="AAXClean.Codecs" Version="0.5.16" />
+	  <PackageReference Include="AAXClean.Codecs" Version="1.0.0" />
 	</ItemGroup>
 
 	<ItemGroup>
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -23,7 +23,7 @@ env:
 
 jobs:
   build:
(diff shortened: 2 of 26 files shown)


Source/AaxDecrypter/AaxDecrypter.csproj
```diff
@@ -13,7 +13,7 @@
 	</PropertyGroup>
 
 	<ItemGroup>
-	  <PackageReference Include="AAXClean.Codecs" Version="0.5.16" />
+	  <PackageReference Include="AAXClean.Codecs" Version="1.0.0" />
```
rmcrackan, line 16: w00h00 -- 1.0. Bet that feels good

wtanksleyjr:
Wow. Wow wow wow.

Wow.

This app keeps getting better!

-Wm

rmcrackan:
Looks great! Are there any tickets to close? Also, can I release a new version now? Or do you have other stuff coming soon?

Mbucari:
> Looks great! Are there any tickets to close?

Just the captcha one.

> can I release a new version now?

Go for it!

Source/AaxDecrypter/AaxDecrypter.csproj
Mbucari: It does. I've done like 3 separate overhauls since version 0.1. I finally removed what I consider the last "hack" bits, so I think 1.0 is justified (:

Issue #514, opened 2023-02-26
No way to actually remove titles no longer in library


rmcrackan:
When we 'remove' titles which are no longer in the user's audible library, they are now only soft deleted. Which means this sequence puts us in a weird state:

* add podcast on audible
* scan libation to import podcast
* remove podcast on audible
* 'remove books' on libation. These titles are soft deleted
* re-add podcast on audible
* scan libation
* episodes do not show up again

fractaldonut:
I think the error is actually more general than that and that it occurs even without steps 3 & 5

[@517 ](https://github.com/rmcrackan/Libation/issues/517)

Mbucari:
What if we change the "Save" button in Hangover's deleted books tab to "Restore" and add a "Remove from Database" button?

rmcrackan:
I like both ideas. If I can ever get any free time (and if you don't beat me to it) my other idea was to see if the removed book was still part of the active audible library. This is easy in the mass 'remove books' feature since we've already query that. The hang up is: it has more overhead when removing a book any other way and I'm less happy about that.

Actually, 'remove from database' might be better worded as 'fully delete from Libation' or similar. Dammit -- I knew this would happen: once remove was an option, we needed an un-remove and now we are exposing Hangover to everyone which means we have to care about crap like this. I really should never have released the soft delete feature into the wild until I had an in-app way to un-soft-delete. I knew better and I let user requests make me feel pressured to be a people pleaser instead of waiting until it was ready. ::sigh:: Oh well. It is what it is and we can only move forward.

Mbucari:
The Removal Books tab is pretty basic. If you don't like it and hangover, it wouldn't be difficult to implement the same dialogue window inside libation. You could even use the Trash metaphor. Items can either be restored or you can empty the trash.

rmcrackan:
Good point. There's certainly precedent in mac and windows for a trash bin limbo.

rmcrackan:
I answer this damn question almost daily on reddit. (Actually 3x yesterday.) I'm kicking myself for releasing this before having a real solution. I guess we made it too easy to accidentally remove books ☹️

rmcrackan:
@Mbucari : 

![Who's awesome?](http://www.quickmeme.com/img/e5/e56c5591a895603b5e41c61c0f930d19aa87cfc00f6c77761e8fcd23102d709c.jpg)

Issue #498, opened 2023-02-16
Track Libation Version and Downloaded Timestamp for Downloaded Files


manalive:
**Is your feature request related to a problem? Please describe.**

I would like to be able to see what version of Libation was used to download a file.

A recent bug fix was related to malformed downloads. Another recent (to me) enhancement was moving the moov atom to the beginning of the file.

It would be great to know which downloaded files need to be re-downloaded to take advantage of bug fixes or a new feature.

**Describe the solution you'd like**

Optimally, I would this info in Libation as columns AND as a custom tag (TXX:Downloader, TXX:Downloaded for example) but even just in Libation would be very useful.

Mbucari:
Date and time download tags [were recently added](https://github.com/rmcrackan/Libation/blob/master/Documentation/NamingTemplates.md). I need a libation version tag easy enough.

The grid is already pretty busy with lots of info, but I guess that doesn't matter so much since you can disable columns. what do you think @rmcrackan?

manalive:
Extra points:
- Value in Libation version column is link to release/notes on github
- Value in Download Date column is clickable link that auto-populates search box with string to find all from the same day (or if not possible, copies that string to the clipboard)

rmcrackan:
> a libation version tag easy enough

It's doable. This will be a break with current practices though. Other than the 'file locations' file we actually don't store anything about specific files. We'd need to create a new location for that info. After that it should be added to the search engine and as a hide-able grid column.

> Value in Libation version column is clickable link to release/notes on github

It's not hard to do this. It's hard to do without cluttering the user interface. It would be effortless to put it in the details box as normal text just like the other details so that's where I'd start.

> Value in Download Date column is clickable link that auto-populates search box with string to find all from the same day

I'm not sold on the specifics but I see where you're going with this. It's a feature idea with far reaching implications. When I filter on this date, do I add the date to my current filter? Or replace my current filter? In my case, I use quite advanced filters. How would it interact with those? Would I just lose my last search? Would we also need a back-button/undo feature?

If we do this, it will need to be well thought out.

> or if not possible, copies that string to the clipboard

Taking over the user's clipboard when you normal left click on something shouldn't happen. We do however have right-click \> copy and ctrl+c also works.

manalive:
> we actually don't store anything about specific files

Personally, I don't care about storing info about the files in libation. What I really want is to see in Libation when the whole book was last successfully downloaded. 

[...]

Issue #433, opened 2023-01-01
Undo Remove from library


Allylleal:
I looked through through the documentation and couldn't find the steps for this. I accidentally hit 'Remove from library' instead of Locate File while right clicking the Liberate icon on a book. 
I've already tried to rescan, remove and then add my account and then rescan. Is there a quick fix I'm missing?
It was too easy to remove the book from the library if it's permanent.

Mbucari:
Launch the `Hangover` app. It's a recovery program in Libation's folder.

Mbucari:
@Allylleal were you able to undelete your books?

imno007:
This does not work for me. The deleted titles show back up in Libation, but when I click to download them, I get a "File not found" error.

This has been an issue for a couple of versions now, including the latest. I probably encounter it more than most because I often convert in both lossless format and by chapter, but I can no longer do this in Libation because I can't import the books a second time. It's usually not a problem, because I'll just convert once as lossless with Libation and then chapterize in Inaudible, but occasionally Inaudible will for some reason hang on converting a Libation .m4b conversion.

imno007:
[_Log202301.log](https://github.com/rmcrackan/Libation/files/10385571/_Log202301.log)

Mbucari:
@imno007 I only see one book download attempt in your log file (Sherlock Holmes: The Thinking Engine), and it was successful.

Can you try downloading one of the problem books again then post that log file? (Also, I think the log files with a preceding underscore are backups. Pick the one without the underscore).

imno007:
I think you're misunderstanding, which might be down to me. Yes, I could download it, but once I remove it from Libation, I cannot download it a second time. 

Also, the reason you only see one download, and why the log being a backup probably does not matter in this case, is because I uninstalled Libation, all folders, and reinstalled it.

rmcrackan:
What happens if you RIGHT click on the stop light and click 'set download status to not downloaded'. Then LEFT click.

Does that start a new download?

imno007:
Yes, THAT works - thank you! Sorry, I somehow managed never to see that option.

And sorry for the long delays in responding. As it happens, I suddenly find that Inaudible is no longer working for me to convert .aax files (even with keys installed), so I was trying to troubleshoot that at the same time. ;)

rmcrackan:
You're welcome. You can also set this status from the top 'visible books' menu. It will apply your desired status to *all* books in the grid.

[...]

Issue #339, opened 2022-08-01
Linux and Mac TODO


Mbucari:
## Linux
- ~~Saving and restoring window size and position doesn't seem to work.~~
- ~~Make a .deb package and upload it to [Launchpad](https://launchpad.net). That will solve the install/update problem because users will just use apt-get.~~
- Add Linux arm64 native libraries. I'll need someone with an arm64 Linux machine to compile these for me.

## Mac
- ~~Mp3 is unavailable. This fix will require making an aac decoder for mac and then tracking down and adding the mac version of libmp3lame.~~
- ~~Figure out how to code sign the executables so users don't have to disable developer verification to use the app. This will [enrolling in the Apple Developer Program](https://developer.apple.com/developer-id/). TBH this seems onerous and maybe impossible without owning a mac.~~
- ~~Similar the the above point, we need to make an installer and updater for macOS. I don't know much about Linux, and I know even less about Mac.  Maybe we need to reach out to other open source devs who support macOS and see how they do it.~~
- ~Add command to open the file explorer to a folder (Go.To replacement)~ https://github.com/Mbucari/Libation/commit/d884d34c5b61c95c9341ff2524bc1b701281093f

## Both
- ~~The Expand/Collapse Process Queue symbols ( ❱❱❱ ❰❰❰ ) are Unicode and don't display properly. Replace with images.~~
- ~~Add the Illegal Character Replacement dialog.  This is a problem because Avalonia doesn't support Unicode, so the defaults don't display properly.~~

sammcj:
Hey @Mbucari - have you had a look at some automated options for this?

The one that popped to mind is using Github Actions which has macOS runners with something like this - https://github.com/marketplace/actions/code-sign-action

Great project by the way!

rmcrackan:
@Mbucari  is this still an active to do list? Or is it stale?

Mbucari:
>  is this still an active to do list? Or is it stale?

Sorry I missed that. Yes, it's stale.

timdmackey:
Thanks to those involved for your work on making this app cross-platform! I discovered Libation yesterday and was happy to be able to backup my Audible library to my mac.

_I see that this issue is closed, but my question is closely related so I didn't want to open a new issue unnecessarily._

Once it's set up, the app works great on MacOS, but the process of running the app could definitely be improved. The following issues could be solved by having the app live in an .app bundle:
- Can't add Libation to the Dock
- Need to sort through 300 files to find the executable each time
[...]

PR #522, opened 2023-03-03
Resolved Several Issues (It's not as bad as  2,453 lines suggests)

Mbucari:
I did a much better job organizing my commits this time, so it may be helpful to review changes on a commit-by-commit basis.

## [Add folder icon support to MacOS](https://github.com/rmcrackan/Libation/commit/b9c2a1cce3e627292e4506d29cf4ea137f785b23)

Added the [fileicon](https://github.com/mklement0/fileicon) script to the MacOsConfigApp (this alone is 733 lines)

I also fixed that background filesystem bug I told you about by not creating that temporary icon file with the GUID name.

## [Parallelize post-liberation tasks](https://github.com/rmcrackan/Libation/commit/abdf0e7261fa88934a57c2b865f899b10f857d5d)

Not only does this slightly speed up the download completion, but it frees the UI.

## [Move TrackedQueue to LibationUiBase](https://github.com/rmcrackan/Libation/commit/df224cc7f3886bc9b3f4c21a103172f11bfd830f)

This is just some housekeeping. I'm trying to move shared UI code into LibationUiBase where I can, but unfortunately opportunities are slim!

## [Add audiobook Trash Bin](https://github.com/rmcrackan/Libation/commit/52a863c62acba3ec9c6ce68963c0bc756ba76579)

I basically copied Hangovers' deleted tabs into Libation and called the new dialogs "Trash Bin". Trash Bin is accessible from the Settings menu. It offers undelete and permanent deleting.

Solves #433 and #514 

## [Add last download into to database](https://github.com/rmcrackan/Libation/commit/bd7e45ca3c5ceda43ef1dbedd2e7adf62dc05504)

I took your advise and used `HasConversion`. It's pretty clean.

## [Add Last Download column to grid](https://github.com/rmcrackan/Libation/commit/d94759d8689f19869bc5bb09743134b89910f05f)

Solves #498

## [Convert queue expand/collapse button text to images](https://github.com/rmcrackan/Libation/commit/26c60e8e7904ac031737ebfded7b882d4428a023)

Solves the Expand/Collapse Process Queue symbols issue in #339 

## Improved custom Folder Icons

It turns out a lot of that stuff you copied from [from here](https://github.com/dimuththarindu/FIC-Folder-Icon-Changer/blob/master/project/FIC/Classes/IconCustomizer.cs) isn't necessary. The icon and desktop.ini files don't need to be system files, and you don't need to refresh after changing the folder's attribute. Also, adding `ConfirmFileOp=0` to the ini file prevents the system from warning users that they're about to delete a system folder.

I also moved the icon encoder to LibationUiBase.IcoEncoder. It implements `SixLabors.ImageSharp.IImageEncoder`, so you can now call `Image.Save(Stream, IcoEncoder)`.


rmcrackan:
This is a big one. Thank you so much for breaking it down like that.

> I took your advise and used HasConversion. It's pretty clean.

I'm a big fan of how it keeps db logic abstracted to just the right place. EF is so much better than it used to be. I had to use it several jobs ago back in the .net framework era and it was a nightmare. I quite like it these days.

> you don't need to refresh after changing the folder's attribute

Color me surprised. In my tests I feel like the icons never showed up right for me without a refresh.

rmcrackan:
I just tried out the new trash and I dig it. Thank you. One details RE the new downloaded stats: can you please include them in the exports?

Mbucari:
> I just tried out the new trash and I dig it. Thank you.

So glad! It's basic, but it doesn't really need to be more than that. (though it would be _nice_ to see cover art).

> can you please include them in the exports?

Sure thing.

Issue #269, opened 2022-06-10
Libation doesn't know when an item has been removed from the audible library


Mbucari:
Described [here](https://github.com/rmcrackan/Libation/issues/263#issuecomment-1152781694)

If a book has been removed from the user's audible library (either by the user or by Audible) but is still in Libations database, Liberate All will always try and fail to liberate it.

My first though is to add a new LiberatedStatus enum value: Unavailable

rmcrackan:
Sure. We could use a new Unavailable enum status instead of Error when that happens.

Mbucari:
On a sort of but not really related note, should we make it possible to remove books from the library in a way that they are not added back on the next account sync?

**Reason**: There are some books that I bought and have no interest listening to or seeing.  I'd probably prefer to not even own them, but automatic return is not available so I'd have to go through contacting audible for them to remove it. Even then, I _may_ want access to the book again in the future.

**Possible Solution**:  Add a flag in the db for the book to be ignored. Maybe an Ignore value in LiberateStatus. By default, no books with that flag are returned when fetching the library.

**Problem** How do I remove the "Ignored" flag once it's set?  The book is ignored, so it's not in the grid for me to edit the UserDefinedItem.  Maybe a separate dialog (like the soon-to-be-deleted RemoveBooksDialog) that lists all books with the Ignored flag and a checkbox to un-ignore?

Mbucari:
> Sure. We could use a new Unavailable enum status instead of Error when that happens.

Now that you mention Error, it got me thinking that a LiberateStatus is not the way to do this.  If you have already liberated the book and the book has become unavailable, it should still show as liberated in your library.  If the book hasn't been liberated and has become unavailable, then... Not really sure what should happen in this case. But the more I think about it, the more this is beginning to sound related to https://github.com/rmcrackan/Libation/issues/204

wtanksleyjr:
> if the book hasn't been liberated and has become unavailable, then... Not really sure what should happen in this case. Maybe just automatically remove that book from the library?

Not automatically, please -- although this specific case would be solved by that, there are cases where having a book you owned vanish means you have the right to contact Audible's support and get a replacement credit. (If it's possible to tell whether the book had been "truly" owned, that could be automatic, since of course podcasts and borrowed books are different from purchased ones.)

rmcrackan:
> should we make it possible to remove books from the library in a way that they are not added back on the next account sync?

[...]

PR #524, opened 2023-03-08
Improve library scan speed and Track and display book availability

### LINKED ISSUE #523: VISIBLE 0 BOOKS
**Describe the bug**
I HAVE MANY BOOKS BUT AFTER NEW UPDATE IT SHOW VISIBLE 0 BOOK
Screenshot 2023-03-05 165128
Screenshot 2023-03-05 165128

**To Reproduce**

**Expected behavior**

**Screenshots**

**Platform**

[e.g. Windows 10, Windows 11, Mac, Linux (State distribution)]

**Log Files**
Attach your Libation log file here.

Mbucari:
Again, changes are broken up by commits (except for "Improve library scan performance" which is split in half because I forgot to pull in a merge).

# [Add last downloaded info to exports](https://github.com/rmcrackan/Libation/commit/cdb27ef712d7e23cace8d885dac27c1a8ec16d73)

You'd asked for this in my last PR

# [Improve library scan performance](https://github.com/rmcrackan/Libation/pull/524/files#diff-86df73cf7abdd921714ff352ba3bb2279ca9909ecdb4357e47f3d5e7cb3d6aa1)

This change is split across 3 commits, but only [ApiExtended.cs](https://github.com/rmcrackan/Libation/pull/524/files#diff-86df73cf7abdd921714ff352ba3bb2279ca9909ecdb4357e47f3d5e7cb3d6aa1) was changed.

Episode fetching was poorly batched. All episodes of a series were fetched synchronously in the same Task, so some podcasts with hundreds of episodes would bog down the scan. Now, Episode retrieval is done in batches of no more than 50, dramatically increasing scan speeds for users with large podcasts (my library scan speed increased by 60%).

Also notice that MaxConcurrency has increased from 5 to 10. Actual max concurrency prior to this change was actually 15. That's because AudibleApi's library scan had a concurrency limit of 10, and Libation would simultaneously download episodes with a max concurrency of 5. Now, episode scanning only begins after the Api has completed scanning the library.

# [Add AbsentFromLastScan](https://github.com/rmcrackan/Libation/commit/f6dcc0db1db8fca42c76d42f288b5494ea54afd1)

I decided on `AbsentFromLastScan` instead of `PresentInLastScan` to maintain default behavior immediately after a user migrates (since  post-migration values will be default false).

# [Show AbsentFromLastScan book status in grid](https://github.com/rmcrackan/Libation/commit/3ebd4ce24381df4fa01e7716a6bea440dfa87700)

If a book (or pdf) is not liberated and `AbsentFromLastScan ` is true, the grid button does nothing and the button image is grayed. Additionally, these absent books are excluded from unliberated counts and are not queued for download.

Solved #269

# [Fix #523](https://github.com/rmcrackan/Libation/pull/524/commits/e76f99ff28bb8bc19f71ec87761061ec839f9bb1)

If cover art fails to load, load the default image instead.

### DIFF
--- a/Source/ApplicationServices/LibraryCommands.cs
+++ b/Source/ApplicationServices/LibraryCommands.cs
@@ -467,6 +467,7 @@ namespace ApplicationServices
 
             var results = libraryBooks
                 .AsParallel()
+                .Where(lb => !lb.AbsentFromLastScan)
                 .Select(lb => Liberated_Status(lb.Book))
                 .ToList();
             var booksFullyBackedUp = results.Count(r => r == LiberatedStatus.Liberated);
--- a/Source/ApplicationServices/LibraryCommands.cs
+++ b/Source/ApplicationServices/LibraryCommands.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using System.Threading.Tasks;
 using AudibleApi;
 using AudibleUtilities;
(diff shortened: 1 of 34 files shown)


rmcrackan:
> Episode fetching was poorly batched

Good call. This is a nice evolution of the process.

rmcrackan:
I like the broken out and linked commits. Very clean. Following those links, I added some comments on the commit level rather than the full PR level. Let me know if you don't get notified of them.

rmcrackan:
Lookin good.

Regarding our other discussion on how to report unavailables: I'm fine with any reasonable choice as a first draft with the understanding that it could require tweaking if our way confuses users.

I'm good to merge after those 2 comments in "Add AbsentFromLastScan" are complete. No major issues.

Mbucari:
> This is a nice evolution of the process.

Yeah, there's only 1 step after this: Write episode asins to a channel, read them in batches of 50 and spin up a new get request, then set series info after all episodes have been downloaded. It sounds messy, but I'm sure it can be made neat and readable with some careful thought.

> how to report unavailables

Hold off on this merge and I'll try adding unavailable books to the taskbar count. I still think not adding unavailable to the Liberate menu items counts is the right choice, but status bar is a good idea.

rmcrackan:
> Hold off on this merge

Will do. Lemme know when you're ready. I'm quite happy with what I see.

Mbucari:
OK, it's done. New "Unavailable" counts for both books and PDFs. If a book or pdf is AbsentFromLastScan but liberated, it's included in the liberated count and not in the Unavailable count.

rmcrackan:
Verifying that this resolves #523 and #269 ?

Mbucari:
> this resolves #523 and #269 ?

Yes indeedy.

CharlieRussel:
Version 9.4.4 is noticeably faster returning from a scan. I didn't do any profiling, so can't quantify it, but I definitely noticed it after this latest install. Well Done!

PR #526, opened 2023-03-10
Add better AYCL detection and add verbose library scan logging

Mbucari:
I added a new plan name check for Audible plus titles. I also added some a new LogArchive class for compressing log files. If verbose logging is enabled, library scan JSONs are written to "LibraryScans.zip", but only the most recent 10 + #accounts are retained.

EDIT* I also added a new event trigger for when a user changes their rating and made Rating IComparable. I was in the middle of a big refactor when this issue came up, so I had to revert all that. But these don't affect anything so I thought I'd just push them now.

### DIFF
--- a/Source/ApplicationServices/LibraryCommands.cs
+++ b/Source/ApplicationServices/LibraryCommands.cs
@@ -7,8 +7,11 @@ using AudibleApi;
 using AudibleUtilities;
 using DataLayer;
 using Dinah.Core;
+using Dinah.Core.Logging;
 using DtoImporterService;
+using FileManager;
 using LibationFileManager;
+using Newtonsoft.Json.Linq;
 using Serilog;
 using static DtoImporterService.PerfLogger;
 
@@ -169,13 +172,21 @@ namespace ApplicationServices
         private static async Task<List<ImportItem>> scanAccountsAsync(Func<Account, Task<ApiExtended>> apiExtendedfunc, Account[] accounts, LibraryOptions libraryOptions)
         {
             var tasks = new List<Task<List<ImportItem>>>();
-            foreach (var account in accounts)
+
+            using LogArchiver archiver
+                = Log.Logger.IsDebugEnabled()
+                ? new LogArchiver(System.IO.Path.Combine(Configuration.Instance.LibationFiles, "LibraryScans.zip"))
+                : default;
+
+			archiver?.DeleteAllButNewestN(20);
+
+			foreach (var account in accounts)
             {
                 // get APIs in serial b/c of logins. do NOT move inside of parallel (Task.WhenAll)
                 var apiExtended = await apiExtendedfunc(account);
 
                 // add scanAccountAsync as a TASK: do not await
-                tasks.Add(scanAccountAsync(apiExtended, account, libraryOptions));
+                tasks.Add(scanAccountAsync(apiExtended, account, libraryOptions, archiver));
             }
 
             // import library in parallel
@@ -184,7 +195,7 @@ namespace ApplicationServices
             return importItems;
         }
 
-        private static async Task<List<ImportItem>> scanAccountAsync(ApiExtended apiExtended, Account account, LibraryOptions libraryOptions)
+        private static async Task<List<ImportItem>> scanAccountAsync(ApiExtended apiExtended, Account account, LibraryOptions libraryOptions, LogArchiver archiver)
         {
             ArgumentValidator.EnsureNotNull(account, nameof(account));
 
@@ -196,6 +207,8 @@ namespace ApplicationServices
             logTime($"pre scanAccountAsync {account.AccountName}");
 
             var dtoItems = await apiExtended.GetLibraryValidatedAsync(libraryOptions, Configuration.Instance.ImportEpisodes);
+
+            archiver?.AddFile($"{DateTime.Now:u} {account.MaskedLogEntry}.json", new JObject { { "Account", account.MaskedLogEntry }, { "ScannedDateTime", DateTime.Now.ToString("u") }, {"Items", JArray.FromObject(dtoItems) } });
 
             logTime($"post scanAccountAsync {account.AccountName} qty: {dtoItems.Count}");
 
--- a/Source/DtoImporterService/LibraryBookImporter.cs
+++ b/Source/DtoImporterService/LibraryBookImporter.cs
@@ -70,8 +70,11 @@ namespace DtoImporterService
 				}
 			}
 
-			//If an existing Book wasn't found in the import, the owning LibraryBook's Book will be null. 
-			foreach (var nullBook in DbContext.LibraryBooks.AsEnumerable().Where(lb => lb.Book is null))
+			var scannedAccounts = importItems.Select(i => i.AccountId).Distinct().ToList();
+
+			//If an existing Book wasn't found in the import, the owning LibraryBook's Book will be null.
+			//Only change AbsentFromLastScan for LibraryBooks of accounts that were scanned.
+			foreach (var nullBook in DbContext.LibraryBooks.AsEnumerable().Where(lb => lb.Book is null && lb.Account.In(scannedAccounts)))
 				nullBook.AbsentFromLastScan = true;
 
 			//Join importItems on LibraryBooks before iterating over LibraryBooks to avoid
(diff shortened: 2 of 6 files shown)


Source/ApplicationServices/LibraryCommands.cs
```diff
@@ -169,22 +171,33 @@ public static async Task<(int totalCount, int newCount)> ImportAccountAsync(Func
         private static async Task<List<ImportItem>> scanAccountsAsync(Func<Account, Task<ApiExtended>> apiExtendedfunc, Account[] accounts, LibraryOptions libraryOptions)
         {
             var tasks = new List<Task<List<ImportItem>>>();
-            foreach (var account in accounts)
+
+            LogArchiver archiver = default;
+
+            if (Log.IsEnabled(Serilog.Events.LogEventLevel.Verbose))
```
rmcrackan, line 177: Personal preference: `Log.Logger.IsVerboseEnabled()` extension in Dinah.Core. You're choice. Just wanted to point it out in case you want it also.

Source/ApplicationServices/LibraryCommands.cs
```diff
@@ -197,6 +210,8 @@ private static async Task<List<ImportItem>> scanAccountAsync(ApiExtended apiExte
 
             var dtoItems = await apiExtended.GetLibraryValidatedAsync(libraryOptions, Configuration.Instance.ImportEpisodes);
 
+            archiver?.AddFile($"{DateTime.Now:u} {account}.json", new JObject { { "Account", account.ToString() }, { "DateTime", DateTime.Now.ToString("u") }, {"Items", JArray.FromObject(dtoItems) } });
```
rmcrackan, line 213: This exposes account name to us directly which I try hard not to do. Since it's only in Verbose though, it's fine.

Source/DtoImporterService/LibraryBookImporter.cs
```diff
@@ -88,12 +91,29 @@ var items_lbs
...
+		 */
+
 
 		//This SEEMS to work to detect plus titles which are no longer available.
 		//I have my doubts it won't yield false negatives, but I have more
 		//confidence that it won't yield many/any false positives.
 		private static bool isPlusTitleUnavailable(ImportItem item)
 			=> item.DtoItem.IsAyce is true
-			&& item.DtoItem.Plans?.Any(p => p.PlanName.ContainsInsensitive("Minerva") || p.PlanName.ContainsInsensitive("Free")) is not true;
+			&& item.DtoItem.Plans?.Any(p =>
```
rmcrackan, line 113: Is AYCE still used anywhere or did they scrap that altogether?

Source/ApplicationServices/LibraryCommands.cs
```diff
@@ -169,22 +171,33 @@ public static async Task<(int totalCount, int newCount)> ImportAccountAsync(Func
         private static async Task<List<ImportItem>> scanAccountsAsync(Func<Account, Task<ApiExtended>> apiExtendedfunc, Account[] accounts, LibraryOptions libraryOptions)
         {
             var tasks = new List<Task<List<ImportItem>>>();
-            foreach (var account in accounts)
+
+            LogArchiver archiver = default;
+
+            if (Log.IsEnabled(Serilog.Events.LogEventLevel.Verbose))
+            {
+				var logfile = System.IO.Path.Combine(Configuration.Instance.LibationFiles, "LibraryScans.zip");
```
rmcrackan, line 179: Am I right that this means that there will be at most 1 stored zip file? ie: only one named LibraryScans.zip

Source/ApplicationServices/LibraryCommands.cs
Mbucari: The account name and audible user id are members of `Item`, so if we're not scrubbing those then there's no reason to not use the uncensored account in filenames.

Source/DtoImporterService/LibraryBookImporter.cs
Mbucari: `DtoItem.IsAyce` is still in use, but none of the "Plan" names use it as far as I'm aware. But my awareness only extends to US and Italy.

Source/ApplicationServices/LibraryCommands.cs
Mbucari: Correct. We just add more jsons into the same zip file so they're all in one place.

Source/ApplicationServices/LibraryCommands.cs
Mbucari: I changed it.

Source/ApplicationServices/LibraryCommands.cs
Mbucari: Nevermind, no they are not. I wouldn't sworn they were in there. Maybe they're in the raw json received from audible but not in the DTO class?

Anyway, I'll use the masked versions of username and then what do you think about writing the zip file on Debug level instead of verbose?

rmcrackan: Sounds good. I don't mind some PII leaking in Verbose. It shouldn't in other levels though, incl. Debug. This bring up something else: it might be better for debugging to save the raw json; not the DTO.

Mbucari: > it might be better for debugging to save the raw json; not the DTO

I fully agree, but we need to change the Api for that. I'm game to do it, but right now this is a bug that that's rendering at least Italian and Indian user's version unusable. I vote just merge and release this for now. These DTOs will be enough for us to get `Plan` data, which is what we need to fix Audible Plus detection for other regions.

rmcrackan: Agreed

github is doing something very frustrating: it keeps claiming there's only 1 commit and keeps updating it to your latest. Instead of letting me see all the actual commits.

Mbucari: Sorry, that's me. I squashed the change to reduce the number of commits because it's still part of [Add better AYCL detection and add verbose library scan logging](https://github.com/rmcrackan/Libation/pull/526/commits/230e014bb11a56be759abc3b0d2507b5bc3503ad). I'll leave it as its own commit next time.

rmcrackan: Thanks. From my end, github's PR reviewer page has the option to see all changes or select commits. Especially on large PRs it can be helpful to only see recent changes.

Source/ApplicationServices/LibraryCommands.cs
```diff
@@ -169,13 +172,21 @@ public static async Task<(int totalCount, int newCount)> ImportAccountAsync(Func
         private static async Task<List<ImportItem>> scanAccountsAsync(Func<Account, Task<ApiExtended>> apiExtendedfunc, Account[] accounts, LibraryOptions libraryOptions)
         {
             var tasks = new List<Task<List<ImportItem>>>();
-            foreach (var account in accounts)
+
+            using LogArchiver archiver
+                = Log.Logger.IsDebugEnabled()
+                ? new LogArchiver(System.IO.Path.Combine(Configuration.Instance.LibationFiles, "LibraryScans.zip"))
+                : default;
```
Mbucari, line 179: I really didn't think this work, but good ol' C# comes through again by making sure I don't shoot myself in the foot!

rmcrackan: The hell? So it created an idisposible ... null???

Mbucari: clr must use IDisposible?.Dispose() internally.

Source/ApplicationServices/LibraryCommands.cs
Mbucari: > it might be better for debugging to save the raw json; not the DTO

I've been brainstorming ways to do this, and everything I've come up with sucks. I thought of:

- Returning a tuple of items and the raw Json
- adding the source JObject as a member of Item
- adding an event to Api that's fired every time Json is read from the Api's HttpClient, and letting callers subscribe to it to handle logging raw responses.
- adding a static setting to Api that instructs it when and where to save raw Json


They all suck for different reasons, so I was wondering if you had any good ideas for how to capture and log raw responses.

Mbucari: @rmcrackan 

Thinking about this some more, I'm actually warming too adding the source JObject as a member of Item. Actually I was thinking of keeping the source JObject for all DTOs as a member of DtoBase. Thoughts?

rmcrackan: I'm not convinced I engineered that part of the audible api correctly when it assumed everything was json. Example: some errors are just strings. Also the api returns no metadata; only data. If I had to redo it, maybe have the response object include the original result json, error info, etc in addition to the actual data payload. That's letting great be the enemy of good though. I think your idea of adding data to that base class is a quick way to accomplish basically the same thing.

Issue #525, opened 2023-03-10
Can't download titles from plus catalog


adampalli:
**Describe the bug**
I have an audible (IN) subscription. When I try to download any title from the plus catalog, It's all unavailable for download. While clicking on the tile, it says "This book can't be downloaded because it wasn't found during the most recent library scan". The titles are still included in the plus catalog and I'm able to play them via the audible app.
**To Reproduce**

Add any title from the plus catalog and try to download them. 

**Expected behavior**
Audiobook should be available for download. 

**Screenshots**
![image](https://user-images.githubusercontent.com/117727805/224310926-dc33ee5e-83a3-4b63-a076-454bf48d69d3.png)

**Platform**

Windows 11, Latest version of Libation (9.4.4.1)

**Log Files**
Attach your Libation log file here.

JakobTischler:
I have the same problem with Audible.it, which doesn't use a credit system but has "included titles" which can be added to the library.

Using Libation v9.4.4.1

baldivox:
His discussed on Reddit _(I've just discovered and am trying out Libation and it seems to be great. I do have a Newbie question - I am getting the message that a lot of books in my library are unavailable and 'This book cannot be downloaded because it wasn't found during the most recent library scan'. Is this normal? Is it because they are free or expired books etc? Thank you.), I have the same query. You asked for the log and I'm not really an IT guy but hopefully this is what you need:

SessionID=a5d5050b-c57c-4d26-8e78-47ee34286f0a.1678446211838 Timestamp=2023-03-10T11:03:31:966+0000 ThreadID=8774 Component=ngl-lib_NglAppLib ErrorID=403 Description="GetCachedNglProfile Status: NGL Profile Unavailable"
SessionID=a5d5050b-c57c-4d26-8e78-47ee34286f0a.1678446211838 Timestamp=2023-03-10T11:03:31:970+0000 ThreadID=8774 Component=ngl-lib_NUSecureProfileDao Description="GetPrefetchedSecureProfile : Prefetched ASNP not found"
SessionID=a5d5050b-c57c-4d26-8e78-47ee34286f0a.1678446211838 Timestamp=2023-03-10T11:03:31:970+0000 ThreadID=8774 Component=ngl-lib_NUSecureProfileFetcher ErrorID=SPDaoLevelStatus:11 Description="ValidateAndGetPrefetchedSP : Failed to fetch prefetched ASNP"
[...]

PR #527, opened 2023-03-11
Improve Library Display performance and Refactor grid viewmodels

Mbucari:
# [Improve Library load and refresh performance](https://github.com/rmcrackan/Libation/commit/e1cd8b8f940ebec33c6c6a25b68a9a69cf5473e1)

- Writing library scans to zip file is now async
- Fixed the n^3 grid refresh problem
- Fixed another grid refresh bug on Avalonia
- Force garbage collection after a library scan. GC was blowing up my rap, each scan adding ~300MB. IDK why it wasn't collection on its own. Wait, maybe because I now have 64GB or ram and there's absolutely no reason to GC when I'm nowhere near capacity. lol

# [WinForms and Avalonia now share all GridEntry view models](https://github.com/rmcrackan/Libation/commit/fb9d0625459fc8c03fce017188086269af1d4d3a)

This is the big one. I successfully merged WinForms and Avalonia GridEntery view models and moved them into LibationUiBase. This removed over 600 lines of code.

I'd been working on this one for a while, and fixing #525 really threw a wrench into this.
This will make maintenance a LOT easier.


rmcrackan:
> GridEntery ... into LibationUiBase. This removed over 600 lines of code. ... This will make maintenance a LOT easier.

Thank you!!! This is exactly how I hoped LibationUiBase would eventually get used.

Source/ApplicationServices/LibraryCommands.cs
```diff
@@ -166,14 +167,15 @@ public static async Task<(int totalCount, int newCount)> ImportAccountAsync(Func
...
+                GC.Collect(GC.MaxGeneration, GCCollectionMode.Aggressive, true, true);
             }
         }
 
         private static async Task<List<ImportItem>> scanAccountsAsync(Func<Account, Task<ApiExtended>> apiExtendedfunc, Account[] accounts, LibraryOptions libraryOptions)
         {
             var tasks = new List<Task<List<ImportItem>>>();
 
-            using LogArchiver archiver
+            await using LogArchiver archiver
```
rmcrackan, line 178: I have never seen not contemplated `await using`; that's wild. What's being awaited? Isn't this just a ctor?

Source/FileManager/LogArchiver.cs
```diff
@@ -6,10 +6,11 @@
 using System.IO.Compression;
 using System.Linq;
 using System.Text;
+using System.Threading.Tasks;
 
 namespace FileManager
 {
-	public class LogArchiver : IDisposable
+	public sealed class LogArchiver : IAsyncDisposable
```
rmcrackan, line 13: Ah, I've never seen IAsyncDisposable. Answers my previous question.

Source/ApplicationServices/LibraryCommands.cs
Mbucari: It's actually a waiting [DisposeAsync](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync).

ZipArchive writes all changes to disk when it's disposed, so it's good to make it async.

Mbucari:
> > GridEntery ... into LibationUiBase. This removed over 600 lines of code. ... This will make maintenance a LOT easier.
> 
> Thank you!!! This is exactly how I hoped LibationUiBase would eventually get used.

You're very welcome. Definitely took quite a bit of work, and there is a tiny bit of redundancy when it comes to invoking PropertyChanged so that all bases are covered for both avalonia and wind forms, but on the whole I think it worked out rather nicely.

Issue #553, opened 2023-03-27
Won't Scan Library


lbhdan:
**Describe the bug**
When I try to scan my library, Libation just "stops responding" and doesn't add any of my new items, and eventually crashes. I'm using today's updated version (9.4.7.1).

**To Reproduce**

**Expected behavior**
It should add new items to my Libation library, so I can liberate them.

**Screenshots**

**Platform**
Windows 10

**Log Files**
I tried to attach the log file, but it is apparently too big. (448 MB, when only 25 MB are allowed)

Mbucari:
@lbhdan Please delete your log file, try to rescan, then pose the new log file here.

lbhdan:
Let's see if this works:

[Log202303.zip](https://github.com/rmcrackan/Libation/files/11083750/Log202303.zip)

Mbucari:
@lbhdan Found the culprit. Thanks for the log!

rmcrackan:
Fixed in v10

Issue #552, opened 2023-03-27
Crash when adding new Audible account


JakeShirley:
**Describe the bug**
I just downloaded Libation for the first time on MacOS Ventura and I cannot add an Audible account.  The app crashes when clicking on "Audible email/login".

**To Reproduce**

1. Click on "No accounts yet, add account"
2. Double click on "Audible email/login"
3. Crash

**Expected behavior**
No crash, let me log in.

**Screenshots**
<img width="912" alt="Screenshot 2023-03-27 at 12 46 37 PM" src="https://user-images.githubusercontent.com/1000311/228050313-3032f2c2-bb02-47f6-86e2-06d59d42ae12.png">

**Platform**

MacOS Ventura, using Libation 9.4.7.

**Log Files**
[crash.txt](https://github.com/rmcrackan/Libation/files/11082466/crash.txt)

Mbucari:
Please upload the Libation log file. It'll be in your user folder/Libation/Log_[date].log

JakeShirley:
[_Log202303.log](https://github.com/rmcrackan/Libation/files/11083759/_Log202303.log)

Here you go!

Mbucari:
@JakeShirley would you please do us a favor and [try this pre-release ](https://github.com/rmcrackan/Libation/actions/runs/4538595319) to see if the problem is fixed? I don't own a mac and after the latest UI framework update I've been unable to run libation on a virtual machine, so I can't debug anything on a Mac right now.

Thanks!

self-absorbed:
@Mbucari I tried the pre-release and I can confirm that the issue is fixed.

Thanks!

JakeShirley:
Glad to here, sorry I didn't get to test this.  The quick turn around is really appreciated, this tool rocks!

Mbucari:
@self-absorbed Thank you so much. This fix was incorporated into v10.

rmcrackan:
Fixed in v10. Released today

Issue #551, opened 2023-03-27
Blank book reported, causing error at download


CharlieRussel:
**Describe the bug**
When I ran the new build (9.4.7.1) against one of my accounts, it reported one title to liberate, but didn't show any titles that weren't downloaded. I attempted to download anyway and got a failed error. From the log, it appears to be a title that is no longer available, but if so it's not one I have any knowledge of, given that I was fully downloaded before this build. 

**To Reproduce**

Scan some libraries, find one item to liberate, click "Liberate Visible Books: 1", return error (see screen shot)

**Expected behavior**
Didn't really expect there to BE any to liberate, given that everything in my library has already been liberated. And it can't be pre-orders because I have two on that account, not one.

**Screenshots**
![image](https://user-images.githubusercontent.com/21000516/228018876-bc5968ed-07a3-4ded-ae83-15bb3581a922.png)
![image](https://user-images.githubusercontent.com/21000516/228018518-3a29e4f0-d380-4e6d-98a2-97dbc83fbeec.png)

**Platform**

Windows 11, Libation Classic 

**Log Files**
2023-03-27 10:18:02 AM	Download & Decrypt Step, Begin: [B0B4W5J4R7] [blank title]
2023-03-27 10:18:02 AM	Download & Decrypt Step, Completed: [B0B4W5J4R7] [blank title]
2023-03-27 10:18:02 AM	Download & Decrypt
2023-03-27 10:18:02 AM	ERROR. All books have not been processed. Most recent book: processing failed
2023-03-27 10:18:02 AM	ERROR: License response not "OK" for asin: [B0B4W5J4R7]

rmcrackan:
This is really weird. I just used Classic v9.4.7.1 to scan for and download my new books without issue. That blank entry has an ID so it's not a completely blank entry, but it's not an ID I can find on their website. Even google can't find it.

Does it have an entry in the grid? If so, what other details does it have when you click the 'tags and details' button?

Mbucari:
It's an episode of https://www.audible.com/pd/Quirks-and-Quarks-from-CBC-Radio-Podcast/B08JJP3RCY

It looks like that episode is no longer available for download. Some podcasts prune their catalog and older episodes become unavailable. If you re-scan your library, Libation should detect that that episode is no longer available and will mark it as unavailable.

CharlieRussel:
Ok. But then why did it not detect that in this scan?

Charlie.
(he/him/his)

[...]

Issue #529, opened 2023-03-13
Please add series # to grid display


CharlieRussel:
It would be useful in trying to see where I have blanks that I might want to infill if I could see the series # in the grid display. So, currently, it shows "Series", but if the series is "Inspector Lynley", having "Well-Schooled in Murder" show as 3rd in the series would be useful. And if it showed as 3 of 21, that might be even more useful. (Not sure you currently have the total number, plus it's something that's subject to change and even interpretation, so less important.)

chieftwidgit:
File sent.  Please verify

CharlieRussel:
Currently, you list series, sub-series, and merged series separated by a comma in the Series column. Not ideal, but workable given the limitations of most grid systems (what a perfect grid system would look like is a rabbit hole we probably don't want to go diving down!). Given that, the simplest answer is to list the books place in each series, in the order they appear in the Series column, separated by the same comma. Giving the primary series only is probably also acceptable. 

(if it weren't too annoying, I'd prefer semicolons as separators for both, but that's just me being cranky.)

chieftwidgit:
Sorry I was sure I replied to #528.

wtanksleyjr:
This is something I use Audiobookshelf for all the time, its Series view
lets you see what books are included in each subseries. (You know, if you
want to see how the competition does it ;) .

-Wm

Mbucari:
@rmcrackan @wtanksleyjr For viewing series, we could add a separate window like we did for "Records" (accessed via the book's context menu > "View Bookmarks/Clips". That window could live query the audible series ID and display all titles that are part of that/those series.

Mbucari:
Check this out:

[](https://user-images.githubusercontent.com/37587114/225118608-b9d20c54-716a-4cfb-afe2-93ea2a34dd5f.mp4)

Mbucari:
> So this is what people can do when they aren't stuck in meetings all day :)

One of the few perks of being overqualified for ones work.

> That's a really cool demo!

Thanks! I also figured out how to distinguish available books from unavailable books. If you query the catalog endpoint with an unauthenticated request, Audible doesn't know if you're able to purchase the item or not so the items look identical. But if you make an authenticated request, books that are unavailable to you will have a null Price.

rmcrackan:
Nice. As in: unavailable for purchase? Cool. Sounds like the perfect time for a convenience property on the DTO for `IsAvailable => Plan is not null` or something like that.

[...]

PR #554, opened 2023-03-28
Bug fixes and guided tour

Mbucari:
## [More migrations to Avalonia 11.0.0-preview6](https://github.com/rmcrackan/Libation/commit/090871f50db1ef1a1b3b403f749e9ca1c18c7bca)

Mostly just cleaning up unused namespaces that became unused in preview 6, but also updating code that was deprecated.

## [Add Series Order column](https://github.com/rmcrackan/Libation/commit/a71ccbac6eb2e40cb90a5c227c69493a68a6f0ce) (#529)

Also, SeriesOrder and LastDownload columns are now hidden by default.

## [Do not launch settings dialog after installation](https://github.com/rmcrackan/Libation/commit/cdf1a01457a9b0f9b2c8262b7f1f9fb33d2eec63) (#510)

What is says, plus a couple of Avalonia UI tweaks for window sizes. Also, Avalonia's setup dialog has a ComboBox to select the theme.

## [Add guided walkthrough](https://github.com/rmcrackan/Libation/commit/702219ee693aa0170c6a49bab7947093b370ed77) (also #510)
[Allow users to cancel walkthrough](https://github.com/rmcrackan/Libation/pull/554/commits/36005508a19a4c79142ee682f58cd82cbe473326)

I think this is really cool. The first time libation is launched (tracked by the new `FirstLaunch` setting), users are asked if they'd like a guided walk through. If they say yes, they're taken on a tour of the accounts dialog, settings dialog tabs, and the import > scan action. The menu items are opened and selected, with short pauses for the user to see how to get to everything.

## [Don't warn for blank password with external login](https://github.com/rmcrackan/Libation/commit/27d2ada5a432dd2aa346990fd4ef2815779aed38)

login eager dialog was not letting users do external login if the password box was blank.

## [Improve book availability detection](https://github.com/rmcrackan/Libation/commit/add31024dad4ddee112a797e903017c5c7dea6e8) (#551)

Based on my and Charlie's scans, it appears that an item is unavailable if it's content_type is null. 

## [Do not import orphaned episodes](https://github.com/rmcrackan/Libation/commit/ec9356b36e48f10f0a9583a34c4d20b59fb76545) (#553)

Some orphaned episodes were still getting through. For this issue, the offending episode claimed that it had a series parent, but the supposed parent's asin was the same as the episode. So now any episode that doesn't have a series will not be imported. Also, no logging of `ImportValidationException`. That's what caused the >400MB log file.

## [Replace editable DataGridTextColumn with TextBox](https://github.com/rmcrackan/Libation/commit/807725f6ffe0295702fefe2d7b2ea47ba6edb21c) (#552)

The user said Libation was crashing when he double-clicked the DataGridTextColumn. I replaced those column types with template columns that contain a TextBox. It's a bandaid, but hopefully it's sufficient.


rmcrackan:
I'm too tired for a proper PR tonight. I'll take a look manana. At first glance though :dancing_banana:

rmcrackan:
> I made the assumption that book ASINs were unique in the DB. I know they're not the primary key, but I still thought there weren't supposed to be duplicates.

This used to be true and first in wins. I believe this was only changed recently.

Mbucari:
> This used to be true and first in wins. I believe this was only changed recently.

Did you change this? Because I haven't touched the book importer. [And a quick look at the change log](https://github.com/rmcrackan/Libation/commits/master/Source/DtoImporterService/BookImporter.cs) doesn't reveal any change to importing unique books either.

rmcrackan:
> Did you change this?

Nope. I thought you did for a recent requester who had strange ideas about how to include duplicates. Guess I remembered wrong. To the best of my memory, books have always been 'first in wins' because there's no clean way to display duplicates of the same book from different accounts.

Mbucari:
Scratch that, it's not that there's a duplicate `Book` asin, it's that there's more than one `LibraryBook` pointing to the same `Book`. Still that should not happen either. We still check for existing `LibraryBook` entries by asin when importing, and don't import duplicates. This _shouldn't_ happen

rmcrackan:
That's actually what I meant above: multiple library book entries with the same book. Schema-wise it makes sense: you own it in multiple accounts. For our app though, how can you reasonably depict the same book twice? I remember some recent request pushed back on this but I don't see how it wouldn't just be confusion.

Mbucari:
OK, so existing LibraBook entries were found like this:

```C#
var existingEntries = DbContext.LibraryBooks.AsEnumerable().Where(l => l.Book is not null).ToDictionary(l => l.Book.AudibleProductId);
```

When Books are upserted during the BookImporter run, they are linked to their LibraryBook in the DbContext instance.  If a LibraryBook has a null book here, that means it's Book was not imported during by BookImporter. So far, so good.

Next wel loop through ALL ImportItems, and we only create and add a new LibraryBook if it was not found in `existingEntries`.

So the only way two LibraryBooks linked to the same book could happen is if, during a single import run, BookImporter somehow upserted an ImportItem but the DbContext's instance did not link the LibraryBook to that Book.

I'm stumped. I don't know how that is possible, but I'm still a db/EF novice.

rmcrackan:
Race condition? Could the recent parallelization improvements cause this?

rmcrackan:
> Do not launch settings dialog after installation

Cool. So this just launches it with full defaults out of the gate? I think people will like that.

rmcrackan:
> Add guided walkthrough

I'm looking forward to trying this out

rmcrackan:
> Don't warn for blank password with external login

Does this have an associated ticket number?

rmcrackan:
> Replace editable DataGridTextColumn with TextBox

Do you want to wait on OP's feedback before merging this code?

Mbucari:
> So this just launches it with full defaults out of the gate?

Yup.

> Does this have an associated ticket number?

No.

> Do you want to wait on OP's feedback before merging this code?

Yeah, let's give it a day. It will also give me more time to ponder the duplicate LibraryBook problem. The LibraryBookImporter is run sequentially after all scanning tasks complete and all Items are aggregated, so it can't be a race condition.

rmcrackan:
> The LibraryBookImporter is run sequentially after all scanning tasks complete and all Items are aggregated, so it can't be a race condition.

Oh good; that's how it should be.

I like what I see. Just ping me when you're ready to go live.

Mbucari:
OK, I've poured over the import process, and I just don't see how it's possible with the current system. This error is happening because there is already more than one `LibraryBook` linked to a book with the same asin in the db. For that to happen, `BookImporter` would have to import one Asin with two separate `BookId`s, otherwise `LibraryBookImporter` would have throw an exception when it tried to add a duplicate. But that also should not be possible because BookImporter only imports distinct asins:
https://github.com/rmcrackan/Libation/blob/e62f01d2a39e02dd77d8f54fb15742878666e282/Source/DtoImporterService/BookImporter.cs#L51-L53

So, I'm truly stumped. Maybe a change over the past few weeks temporarily allowed duplicates somehow (except as I've shown, `BookImporter` would have to have added two books with the same ASIN, and I haven't touched `BookImporter`). But whatever the cause, the distinct check I added today should solve it without side effects.
<br>
I did some more tweaking on the Accounts Dialog. It's possible the previous janky code was causing the crash instead of the DataGridTextColumn. If you ever look through the Avalonia code in detail, you'll notice that it tends to look cleaner and clearer the more recently that it's been edited. I haven't touched the Accounts Dialog since my initial port (except to update the file dialogs with new Avalonia versions). I was still figuring out MVVM and Avalonia back then, and a lot of my early efforts ware very hacky.

Mbucari:
Oh, I also added two new steps to the walkthrough: Filters and Quick Filters

Go ahead and review and pull. I'm not holding my breath that that guy will test and get back to me...

Mbucari:
Alright, she's ready. Can't wait for some feedback on the walkthrough!

rmcrackan:
Fuuuuuuuuck, that guided tour/walk through is tight. I can see why you're proud of it.

* I like the cancel option throughout
* I accidentally did things twice because I thought I was supposed to click it, then the walk through did it too. That is: something blinked so I clicked on it, then the tutorial also took the same action. Was especially confusing when multiple dialogs launched. Which come to think of it shouldn't have been possible
* Accounts. 'click save'. I only did this in Classic and this is correct. Don't remember if Save button exists in Chardonnay
* "Change where liberated books are stored." ... but I don't want to. What to do? Looks like "Save" worked. Same confusion on subsequent Settings tabs. Let user know that "Save" will take them to the next step
* Even the boxes' titles are right. Nice touch
* Pretty slick that it even initiates initial scan. However my first walk through attempt reset my existing accounts, prompting for password. Could not reproduce later
* Love that there's the Settings-menu option to take a guided tour in case you want to do so later. In later walk-throughs it will be especially important that the user be clear on how to progress through any step without making obligatory changes
* The '?' box filter cheat sheet needs serious work. I'm embarrassed that it's lived this long in its original state
* Legacy 'edit quick filters' bug: top item shouldn't have an enabled move-up button; bottom item shouldn't have an enabled move-down button
* Walk through abruptly stops after edit-filters. Should let users know it's complete by signing off, saying something positive, yada yada

Wow. This is powerful stuff. I wonder how much of the readme stuff could be replaced by just this.

rmcrackan:
I'm going to hold off on creating a release from this PR until you decide if you want to tweak the walk through first. It should get the first impression it deserves.

Mbucari:
> something blinked so I clicked on it, then the tutorial also took the same action. it shouldn't have been possible

You're right, that shouldn't be possible. I think I'll disable those menu items/buttons while they are flashing.

> 'click save'. I only did this in Classic and this is correct. Don't remember if Save button exists in Chardonnay

It does. I think you're thinking of the cancel button, which I almost entirely left out of the Chardonnay dialogues. To me it's redundant with the close window function.

> "Change where liberated books are stored." ... but I don't want to.

Maybe I'll change the language to something like "Here is where you can..."

> Same confusion on subsequent Settings tabs. Let user know that "Save" will take them to the next step

This is actually intentional. You can't close the settings dialog until every tab has been visited and that tab's message box has been displayed. It may feel a little heavy-handed, but it's the only way I could think of to allow the users the freedom to spend as much time as they want on each tab, but at the same time ensure that they visit every tab.

Maybe I'll programmatically change the text of that button too "Next".
 
> my first walk through attempt reset my existing accounts, prompting for password. Could not reproduce later

VERY weird. That definitely should not be possible because scanning is performed by performing a click action on the menu item. Maybe because I only performed that part of the walkthrough if they have actually added accounts which I check by opening their account settings. Maybe having the account settings open caused an error when a different thread was also trying to access the same file? I'll look into it. Maybe I need to manually dispose of the account settings before invoking the scan.

> Legacy 'edit quick filters' bug: top item shouldn't have an enabled move-up button; bottom item shouldn't have an enabled move-down button

I'll check it out.

> Walk through abruptly stops after edit-filters. Should let users know it's complete by signing off, saying something positive, yada yada

Good idea. I was on the fence about this but I think you're right. One thing that's especially nice about using the step runner is that this whole process is highly extensible. Adding more tutorials as simple as adding a new method to the step runner. I thought about adding walkthroughs for the grid, but programmatically interacting with the grid view is, well, problematic.
<hr>

Thank you so much for the detailed review! I'll work on it dressing these tomorrow morning, and hopefully we can have a release pushed by late afternoon!

rmcrackan:
> cancel button ... To me it's redundant with the close window function

Functionally, it's 100% redundant -- if you have the correct knowledge. I include them for clarity. If you change something and hit cancel, it's clear that nothing should have been saved. However, especially in many modern apps, as soon as you change something then it's saved; X-ing out just closes the box you're in. Lacking a Cancel button, it's unclear which paradigm is at play.

You could just end each of these tab steps with "\r\n\r\n(Click Save to continue)" or similar. This turns into the kind of thing you visually stop seeing after a few iterations. Which is a good thing: it doesn't take up much attention but it's still there in case it's useful.

> Maybe I'll programmatically change the text of that button too "Next".

That also works

> step runner is that this whole process is highly extensible

Yes! As I'm sure you've noticed throughout my code, I love creating polymorphic stuff like that too. So powerful and abstracted away from the business logic. Feels good man

> I thought about adding walkthroughs for the grid, but programmatically interacting with the grid view is, well, problematic.

Yeah, would be cool. It's already a lot though. Don't want to risk burning out the user.

PR #558, opened 2023-03-29
Refined Walkthrough

Mbucari:
- Flashing menu items and buttons are disabled until they're programmatically clicked (This was an absolute bitch to get working with Avalonia)
- `AccountsSettingsPersister` is disposed before scanning accounts. This should prevent file damage.
- Changed message verbage and added more info to some message boxes.
- Settings Cancel button is hidden, and the "Save" button is changed to "Next Tab" until all 4 tabs have been visited.
- ~~Disabling Quick Filter buttons is damned difficult. We'd have to:~~
  1. ~~Use binding to a custom QuickFilter class instead of DataGridView's internal list~~
  2. ~~Create a custom DataGridViewButtonColumn to override Paint~~
  3. ~~Create a custom BindingList that can change properties of the first and last elements when the collection changes.~~

~~I'm not doing that for such a minor graphical quirk.~~

Scratch that. My head is way too far into mvvm. I think you could just subscribe to the paint event and accomplish the same thing with a lot less code.

### DIFF
--- a/Source/DtoImporterService/BookImporter.cs
+++ b/Source/DtoImporterService/BookImporter.cs
@@ -46,7 +46,7 @@ namespace DtoImporterService
 			var productIds = importItems
 				.Select(i => i.DtoItem.ProductId)
 				.Distinct()
-				.ToList();
+				.ToHashSet();
 
 			Cache = DbContext.Books
 				.GetBooks(b => productIds.Contains(b.AudibleProductId))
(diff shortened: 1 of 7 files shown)


Mbucari:
OK, I've done the quick filter button fix. Ready for merge.

Mbucari:
Done.

Mbucari:
I found a bug in Quick Filters that allowed the new row buttons to be clicked

rmcrackan:
I'm ready to call this v10 is you are. Thoughts?

Mbucari:
That was my plan. Let her rip!

PR #564, opened 2023-03-31
Fix #563 and probably fix #534

### LINKED ISSUE #534: Shortcut ALT-L conflicted with Apple-@ (ALT/Option-L)
On MacOS the @ has keystroke ALT/Option-L. Typing this in filter-field i then open up the "Liberate"-menue

ALT-I ⁄
ALT-L @
ALT-X ≈
ALT-F ƒ
ALT-V √
ALT-S ‚

### LINKED ISSUE #563: Log in with browser command not working
**Describe the bug**
When you click on the link to log in through the browser, the login window closes but apart from this nothing else happens. I went into the account manager, added a new account, then after clicking save and getting the login screen I clicked to log in through the browser link. This closes the login window, but no browser ever opens.

I reproduced this with Libation 10.0.2 on 3 systems - 2 Windows 10 computers with Classic, and an M1 Mac with Mac OS Ventura.

I know this feature worked before because I was able to log in this way last year but discovered it's now broken after someone asked me for help logging in. Also, the reason I don't just go through the usual log in method is because I'm blind and use a screen reader, and on the browser Audible offers an audio captcha which isn't available in Libation.

Mbucari:
## [Fix external login failure error](https://github.com/rmcrackan/Libation/pull/564/commits/255e18eb5e997b705c40c27b8fdadfcb21a8e5ce) (#563)

The rest of this PR is stuff I was working on to address #534. Best case scenario is it works out of the box. Worst case scenario is menu item shortcut keys are broken on mac.

### DIFF
--- a/Source/LibationAvalonia/AccessKeyHandlerEx.cs
+++ b/Source/LibationAvalonia/AccessKeyHandlerEx.cs
@@ -11,26 +11,21 @@ namespace LibationAvalonia
 
         public AccessKeyHandlerEx(KeyModifiers menuKeyModifier)
         {
-            KeyModifier = menuKeyModifier;
-            ActivatorKeys = menuKeyModifier switch
-            {
-                KeyModifiers.Alt => new[] { Key.LeftAlt, Key.RightAlt },
-                KeyModifiers.Control => new[] { Key.LeftCtrl, Key.RightCtrl },
-                KeyModifiers.Meta => new[] { Key.LWin, Key.RWin },
-                _ => throw new System.NotSupportedException($"{nameof(KeyModifiers)}.{menuKeyModifier} is not implemented"),
-            };
-        }
+			KeyModifier = menuKeyModifier;
+			ActivatorKeys = menuKeyModifier switch
+			{
+				KeyModifiers.Alt => new[] { Key.LeftAlt, Key.RightAlt },
+				KeyModifiers.Control => new[] { Key.LeftCtrl, Key.RightCtrl },
+				KeyModifiers.Meta => new[] { Key.LWin, Key.RWin },
+				_ => throw new System.NotSupportedException($"{nameof(KeyModifiers)}.{menuKeyModifier} is not implemented"),
+			};
+		}
 
         protected override void OnPreviewKeyDown(object sender, KeyEventArgs e)
         {
-            if (ActivatorKeys.Contains(e.Key))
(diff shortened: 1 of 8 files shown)


rmcrackan:
Nice. Ready for merge?

rmcrackan:
There've been a lot of releases lately and I try to be mindful not to overwhelm users with constant updates. Is this a fix to release now? Or wait until some more accumulates?

Mbucari:
@rmcrackan Really up to you. It's a bug affecting login which, if they're unable to use the primary login method, prohibits their ability to use the program at all. But it is just a backup method, so maybe wait.

Also, I just tried backing up a book to mp3 and got a NRE from AAXClean. So hold off at least until I get that fixed. This is the only book I've encountered that breaks mp3 conversion.

Mbucari:
@rmcrackan [v1.0.2](https://www.nuget.org/packages/AAXClean.Codecs/1.0.2) is live.

rmcrackan:
Cool, I just updated to v1.0.2

I'll wait a bit on the release in case more gets done this weekend.

Mbucari:
> I'll wait a bit on the release in case more gets done this weekend.

Cool. Then I think I'll ask that mac user to try the preview to see if it fixes his hotkey problem.

Mbucari:
Speaking of preview builds, maybe we should make better use of that GitHub feature when trying to fix users' issues that we're not 100% sure will work. If we think we've fixed an issue, publish a new release as a preview first and have the user confirm that it's fixed. If it is, convert it to a regular release. If it's not, delete the release, delete the tag, and try again.

Mbucari:
Or for cases just like this, when we fixed a user's issue but don't want to push a new release to everyone just yet. Publish it as a preview, and then either convert it to a regular release later or delete it to make more changes before a regular release.

rmcrackan:
I'll give that a try. I didn't pay attention to whether that was an option when I was running my builds.

Mbucari:
Same steps as normal, but then you are editing the draft release, check the "Set as a pre-release" box. It won't show up as the latest release, and users won't get an auto update notification.

rmcrackan:
I know that's how it works in theory. In practice I haven't had a reason to see if that were an option. I'm about to try it.

rmcrackan:
[Pre-release v10.0.3](https://github.com/rmcrackan/Libation/releases/tag/v10.0.3) is live

Issue #536, opened 2023-03-16
Action on + sign acts on all episodes


CharlieRussel:
**Is your feature request related to a problem? Please describe.**
No, it's a pure enhancement. But the current behaviour is counter-intuitive

**Describe the solution you'd like**
Currently, when a podcast or other episodic title is shown in the grid, it's hidden behind a single + sign for the entire series. Which is fine. BUT, when you want to act on all of the episodes (ie, change download status, remove, etc) you have to act on each individual item, and they don't show in "visible" items unless you first expand the group. It makes more sense, IMHO, to allow any action to act on all of the items hidden by the + sign. (And, if you insist, you could add a warning dialog, though I've been an admin for too long to expect to be warned when I do something stupid.)

Mbucari:
At the core, it sounds like what your asking is for all book/episode entries to be counted among "visible" books if they pass the search filter. Is that right?

@rmcrackan What are your thoughts?

CharlieRussel:
@Mbucari -- Yeah, I think basically that works. Though my first banging my head up against it was right clicking on the + sign and not being able to actually do anything with all those ancient episodes that they used to treat as a "feature" or bonus.

rmcrackan:
I agree. "Visible" should be anything that matches the filter.

Although we can't actually see them, I think "visible" is more intuitive to the average user than "matches." I doubt many people even use or think about the search/filters. I'm open to suggestions though if you have a better idea.

Mbucari:
@CharlieRussel It's also pretty easy to add a context menu to the plus/minus button. What sort of actions would you like to see in there?

CharlieRussel:
It seems to me that it should include all the things you can do to the 'visible' books. If we're treating those hidden by the + sign as visible, then it should carry over to the context menu. Or am I missing something??

Mbucari:
Setting downloaded status and removing both make sense for groups of items. Operations that don't make sense are:

- locate the audiobook
- view bookmarks and records

Operations that may make sense:
- download all children 
- convert to mp3. Nearly all podcasts are MP3s to begin with, but if you are mp4.
- view other items in series. This is the future I'm working on to view all items on Audible that are a part of the series.

Any operation that acts on a group row will necessarily be a batch action. Some operations just don't make sense when batched.

CharlieRussel:
[...]

PR #565, opened 2023-04-02
About Dialog, mac menus, and hotkeys

Mbucari:
## [Add native menu for mac and refactor MainWindow](https://github.com/rmcrackan/Libation/commit/4a65d6bbd3faa9579193217fd29ecbebacfb5f70)

The majority of these changes were to move biz logic from `MainWindow` into its view model, `MainVM` (Previously named `MainWindow`). `MainVM` is still organized into multiple files, but the grouping is a little different than in classic. The whole reason for doind this is so I could use Avalonia's NativeMenus on macOS. Linux and Windows still use the original windowed menu, but macOS hides that menu and uses the OS's menu bar. Being native, customization is quite limited, so that menu is unhidden and used for the guided tour while the tour is running. All menu items are identical, so the point should be clear.

I've been meaning to refactor the view model for a while, and this was a good excuse.

## [Add About dialog](https://github.com/rmcrackan/Libation/commit/8d73f5cc7e0ee3e4b9bd587d00006c521335eda6)

Exactly as I showed in #534

## [UI Tweaks and new application hotkeys](https://github.com/rmcrackan/Libation/pull/565/commits/756d387238a06fab60dd6c7fc2bbfb8beef406a1)

Tweak the UI, and add application hotkeys to MainForm:

|Win/Linux|macOS|Action|
|-|-|-|
|ctrl+F|⌘+F|Focux the search filter textbox and select all text|
|ctrl+P|⌘+,|Settings|
|ctrl+shift+A|⌘+.|Accounts|
|F1 - F12|⌘+1 - ⌘+0|Apply nth Quick Filter|

## [Sort episodes by column beneath their parents](https://github.com/rmcrackan/Libation/pull/565/commits/2248d7b24eada32f8bec49853184f4c7b8562b3e)
[As discussed here](https://github.com/rmcrackan/Libation/issues/536#issuecomment-1483978394)


Mbucari:
I updated the about dialog.
image

rmcrackan:
Very cool. Minor suggestions

* I agree with CharlieRussel -- I'm not sure there's utility in showing loaded assemblies to users
* Please add your above shortcuts chart to the Mac's readme page

Mbucari:
All right, how about this?
image

> add your above shortcuts chart to the Mac's readme page

There's no need. When you open the menu by clicking on it, you can see the shortcut key combinations for each menu item.

rmcrackan:
Looks great for now and it will give us plenty of room to expand if needed. Goodbye shitty little messagebox with nothing but the version -- you'll not be missed.

Mbucari:
Hold off on releasing this. IDK how this happened, but during my editing of the Winforms version somehow event handlers got removed from two controls.

CharlieRussel:
Another thought that might be useful is a link/button to pull the logs and/or system information? Which could include details like which assemblies are loaded, what the underlying platform is, etc. And you could even have it zip them up to include with any bug report.

Issue #574, opened 2023-04-07
Classic. liberate new book from top menu


rmcrackan:
In Classic. Reproduce:

* permanently delete a book from libation (remove from library, trash bin, permanently remove from libation)
* scan
* using top menu: liberate all \*
* result: stop light still red. tags and details : not downloaded
* attempt to download again will re-download. ie: it's not just a display problem; Libation doesn't recognize that this was just downloaded
* restart libation: stop light is green, and 'tags and details' = downloaded

Note, if \* step occurs by clicking stoplight instead of using top menu: no error

rmcrackan:
Filter is fixed. Download state of new books is still wonky

* filter: `-[demotag]`
* permanently delete a book from libation (remove from library, trash bin, permanently remove from libation)
* scan
* using top menu: Liberate , Begin Book and PDF Backups
* result: stoplight is green. Click tags and details button: **Not** Downloaded
* edit book's tags: `demotag`
* result: book is hidden
* clear filter
* result: book is visible, stoplights are red

PR #577, opened 2023-04-09
Fixed your issues

Mbucari:
#574 was particularly difficult to nail down.

### DIFF
--- a/Source/LibationUiBase/GridView/GridEntry[TStatus].cs
+++ b/Source/LibationUiBase/GridView/GridEntry[TStatus].cs
@@ -154,6 +154,10 @@ namespace LibationUiBase.GridView
 			if (udi.Book.AudibleProductId != Book.AudibleProductId)
 				return;
 
+			//If UserDefinedItem was changed on a different Book instance (such as when batch liberating via menus),
+			//EntryStatu's Book instance will not have the current DB state.
+			Liberate.Book = udi.Book;
+
 			// UDI changed, possibly in a different context/view. Update this viewmodel. Call NotifyPropertyChanged to notify view.
 			// - This method responds to tons of incidental changes. Do not persist to db from here. Committing to db must be a volitional action by the caller, not incidental. Otherwise batch changes would be impossible; we would only have slow one-offs
 			// - Don't restrict notifying view to 'only if property changed'. This same book instance can get passed to a different view, then changed there. When the chain of events makes its way back here, the property is unchanged (because it's the same instance), but this view is out of sync. NotifyPropertyChanged will then update this view.
--- a/Source/LibationWinForms/GridView/GridEntryBindingList.cs
+++ b/Source/LibationWinForms/GridView/GridEntryBindingList.cs
@@ -24,7 +24,11 @@ namespace LibationWinForms.GridView
 	internal class GridEntryBindingList : BindingList<IGridEntry>, IBindingListView
 	{
 		public GridEntryBindingList() : base(new List<IGridEntry>()) { }
-		public GridEntryBindingList(IEnumerable<IGridEntry> enumeration) : base(new List<IGridEntry>(enumeration)) { }
+		public GridEntryBindingList(IEnumerable<IGridEntry> enumeration) : base(new List<IGridEntry>(enumeration))
+		{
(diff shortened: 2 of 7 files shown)


Mbucari:
I changed the original title because "Fixing your bugs" had an accusatory tone, but "fixing your issues" sounds even more judgemental 😂

Source/LibationWinForms/GridView/GridEntryBindingList.cs
```diff
@@ -24,7 +24,11 @@ namespace LibationWinForms.GridView
 	internal class GridEntryBindingList : BindingList<IGridEntry>, IBindingListView
 	{
 		public GridEntryBindingList() : base(new List<IGridEntry>()) { }
-		public GridEntryBindingList(IEnumerable<IGridEntry> enumeration) : base(new List<IGridEntry>(enumeration)) { }
+		public GridEntryBindingList(IEnumerable<IGridEntry> enumeration) : base(new List<IGridEntry>(enumeration))
+		{
+			SearchEngineCommands.SearchEngineUpdated += (_,_) => ApplyFilter(FilterString);
```
rmcrackan, line 29: This isn't functionally wrong, but it feels philosophically wrong. Something will update the object then persist to the database, possibly within the same method. The persistence will say when it's updated and will then update the search engine. It seems like the persistence-update would be the thing to subscribe to. I'm going off of memory though and you've obviously been more boots on the ground lately. Thoughts?

Source/LibationUiBase/GridView/GridEntry[TStatus].cs
```diff
@@ -154,6 +154,10 @@ private void UserDefinedItem_ItemChanged(object sender, string itemName)
 			if (udi.Book.AudibleProductId != Book.AudibleProductId)
 				return;
 
+			//If UserDefinedItem was changed on a different Book instance (such as when batch liberating via menus),
+			//EntryStatu's Book instance will not have the current DB state.
+			Liberate.Book = udi.Book;
```
Comment (rmcrackan) [inline Source/LibationUiBase/GridView/GridEntry[TStatus].cs:159]: That's wacky. I bet tracking down this idiosyncrasy was a challenge and a half.

btw -- typo on `EntryStatu's`

rmcrackan:
> fixing your issues

You don't begin to have the time to fix my issues

Thank you for being an advocate for: do it if and only if it's fun. I've felt guilty lately at how much you've done while I've very much not been in the mood. Especially on this batch when you're fixing the bugs I which submitted instead of fixing myself.

Source/LibationUiBase/GridView/GridEntry[TStatus].cs
Comment (rmcrackan) [reply Source/LibationUiBase/GridView/GridEntry[TStatus].cs:159]: Thank you for leaving responsible comments. This is perfect -- it's a note to us later about why this unintuitive thing was done and what exactly was wrong without it.

Source/LibationWinForms/GridView/GridEntryBindingList.cs
Mbucari: I'm not fond of this solution either, but this bug is a problem...

Here's where we detect changes and re-filter.
https://github.com/rmcrackan/Libation/blob/e52772826af681b3805e292b3bea19602975d305/Source/LibationWinForms/GridView/GridEntryBindingList.cs#L218-L227

The problem is that the item changes before the search has been re-indexed. I feel like this has been an ongoing problem.

Anyway, I'll download an old release tomorrow to see if I can find a time when it was working properly and see what we did then. I'm not in the best state to be thinking about this now.

Mbucari:
> You don't begin to have the time to fix my issues

Don't feel bad about that. It's people like us who keep the mental health profession thriving. We're job creators!

Source/LibationUiBase/GridView/GridEntry[TStatus].cs
Comment (Mbucari) [reply Source/LibationUiBase/GridView/GridEntry[TStatus].cs:159]: Yeah, I looked at that and thought "I'll probably try to remove this later". Lol

Source/LibationWinForms/GridView/GridEntryBindingList.cs
rmcrackan: Will `BookUserDefinedItemCommitted` accomplish the same?

> I'm not in the best state to be thinking about this now

Take care of you. Nothing about this is urgent.

PR #584, opened 2023-04-14
Web Browser Login for Windows

Mbucari:
## [Use WebLoginDialog as primary login method on Win10+](https://github.com/rmcrackan/Libation/commit/df2936e0b6f4c6dfd788c49192050ab041f8e129)

This is just for classic. If the user is running on win10 or above, they're presented with the web browser login. If they're on a lower version or for some reason the browser login doesn't work, login falls back to the existing Eager login.

## [Add WebLoginDialog for Windows Chardonnay](https://github.com/rmcrackan/Libation/commit/44564321169d93d572f2b529aecf61625a1db850)

This is the big one. It turns out it's possible to embed native windows inside of Avalonia. I managed to do it for the WebView2 control on windows using the LoadByOS arch (so chardonnay has winforms again). Doing this took significantly longer than 1 hour lol.

It also looks like it's possible to do this on macOS using webkit and .net macos framework. I tried, but ran into problem after problem. I solved several, but I've temporarily given up on the last one. The compiler was bitching about the AAXClean's native libraries. Mad no sense to me. I'll get back to it eventually, but I'm not really feeling up to it right now.

## [Grid refresh performance and behavior improvements](https://github.com/rmcrackan/Libation/commit/82fba7e75271341c2238118ee56b033da9e788e1)

Minor performance improvement. Suppress property changed notifications while doing a batch update, then reset the list at the end.

This also fixes the bug where the grid would scroll all the way to the top after updating.


rmcrackan:
> Use WebLoginDialog as primary login method on Win10+

I am shocked at how little code this took. Nice job. I noticed you're referencing a prerelease. Intentional?

> Add WebLoginDialog for Windows Chardonnay

Wow, yeah. That's a lot of code. Oof. You've been busy.

> Grid refresh performance and behavior improvements

With your new flag, are you concerned that something could break in the middle and you'd be stuck in that state. In other words, do you need a try/finally?

Mbucari:
> I noticed you're referencing a prerelease.

Yeah, I have my nuget client set up to show prereleases because of Avalonia.

In the [Add WebLoginDialog for Windows Chardonnay](https://github.com/rmcrackan/Libation/commit/44564321169d93d572f2b529aecf61625a1db850) commit, I removed that reference from LibationWinForms and use the interop object instead.

> With your new flag, are you concerned that something could break in the middle and you'd be stuck in that state. In other words, do you need a try/finally?

That flag is actually a member of [BindingSource](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.bindingsource.raiselistchangedevents?view=windowsdesktop-8.0). I'm confident that if anything breaks, it won't be throwing an exception in the UpdateGrid method. The error will happen on the UI thread when DataGridView and the bound list are out of sync. Regardless, I don't think there's a problem here since the list is reset after restoring `RaiseListChangedEvents`.

Mbucari:
> That's a lot of code.

Thankfully a lot of that was already written by one of the Avalonia devs. I just had to bring it up to date and add a couple of things.

PR #588, opened 2023-04-17
Use old activation bytes if present.

Mbucari:
This change is all that's needed to use the new aax fallback method in https://github.com/rmcrackan/AudibleApi/pull/45


rmcrackan:
AudibleApi is updated. Just update this PR to use 8.4.0.1 and I'll push it thru

rmcrackan:
Self-merging PRs -- I could get used to this

PR #616, opened 2023-06-03
Fix #612 and update Avalonia to v11-rc1

### LINKED ISSUE #612: "Save all podcast episodes to parent folder" option is not compatible with conditional naming
I have the following folder format string:

```<if podcast->Podcasts/<first author>/<series>/<series#[3]> - <title short><-if podcast><!if podcast->Audiobooks/<first author>/<if series-><series>/<series#[2]> - <-if series><title short><-if podcast>```

If the `Save all podcast episodes in a series to the series parent folder?` option is checked, the conditionals do not work, and I get incorrect folder output. With the option unchecked, it works fine.

Incorrect:
Incorrect

Correct:
Correct

### PR DESCRIPTION
**Pull request by Mbucari**
## [Fix SavePodcastsToParentFolder setting](https://github.com/rmcrackan/Libation/commit/75c07c3209f08517e6721e486cac2aa80236ec5b) #612

- `<if series>` and `<if podcast>` both evaluate true for series parents.
- When SavePodcastsToParentFolder is enabled, episode files are saved in the series parent folder and not in their own episode folder.
- `<bitrate>`, `<channels>`, `<series#>`, and `<samplerate>` evaluate to empty strings for podcast parents.
- New conditional tag `<if podcastparent>` for folder naming templates
- I tested setting and re-setting the folder icons, and there are no problems.

### CODE REVIEW

File: Source/LibationFileManager/TemplateTags.cs [hunk 1]
```diff
@@ -47,6 +47,7 @@ private TemplateTags(string tagName, string description, string defaultValue = n
         public static TemplateTags DateAdded { get; } = new TemplateTags("date added", "Date added to your Audible account. e.g. yyyy-MM-dd", $"<date added [{DEFAULT_DATE_FORMAT}]>", "<date added [...]>");
         public static TemplateTags IfSeries { get; } = new TemplateTags("if series", "Only include if part of a book series or podcast", "<if series-><-if series>", "<if series->...<-if series>");
         public static TemplateTags IfPodcast { get; } = new TemplateTags("if podcast", "Only include if part of a podcast", "<if podcast-><-if podcast>", "<if podcast->...<-if podcast>");
+        public static TemplateTags IfPodcastParent { get; } = new TemplateTags("if podcastparent", "Only include if item is a podcast series parent", "<if podcastparent-><-if podcastparent>", "<if podcastparent->...<-if podcastparent>");
```
rmcrackan, line 50: I love seeing good architectural decisions pay dividends like this. It's so clean! My only suggestion is to also update the tagging documentation.

rmcrackan: Never mind. I see it now. Not sure how I missed that before. Whoops

Issue #618, opened 2023-06-08
Searches on ID field are not working (10.3.3)


manalive:
**Describe the bug**
None of these fields are producing results:
![image](https://github.com/rmcrackan/Libation/assets/1682957/9715c71f-7fad-4f4a-90f5-220170280fa9)
If I just search the ID number without the including the field it _usually_ works but not always. I am not sure the pattern but I suspect books with all numeric IDs (no B0 prefix) are the most problematic.

rmcrackan:
@Mbucari The fix now appears to be case specific. `id:B06WLMWF2S` shows only Sherlock Holmes, `id:b06WLMWF2S` shows nothing

manalive:
This is now working for some IDs but not all numeric IDs where there are leading zeroes.

For example: https://www.audible.com/pd/Unfinished-Tales-Audiobook/0008371113

(Sorry I can't find any examples that are PLUS/Free for testing)

rmcrackan:
Reopening because of

> The fix now appears to be case specific

...

> This is now working for some IDs but not all numeric IDs where there are leading zeroes.

rmcrackan:
Let's try this again -- fixed in [v10.3.7](https://github.com/rmcrackan/Libation/releases/latest)

PR #623, opened 2023-06-10
Redesign query sanitizer (#618)

Mbucari:
It's often the most piddly shit that take the most time and effort to solve, and #618 is a prime example.

First of all, I was wrong about QueryParser converting everything to a number. That was done by `SearchEngine.FormatSearchQuery` The regex was being too aggressive, and the only way to _not_ convert ASINs to integers was to tokenize the whole query and not do it for an ID. Thank god Lucene exposes their built-in tokenizer.

But since I was tokenizing the whole query and looking at each term anyway, I took the opportunity to dispense with a lot of the regex that could be done more easily in the token loop.


Mbucari:
I see a nice way to simplify the "[tag]" expanded. Converting to draft.

Mbucari:
Ready for review!

rmcrackan:
> It's often the most piddly shit that take the most time and effort to solve

Truer words. Add this new huge feature? It's actually a minor refactor and new implementation of an existing well implemented feature -- done. Change the text on this button? Oh fuck, it has to go through the localization engine which has a hack for if you're pluralizing words with a capital letter in the middle and you can't use that exact string because it contains a word reserved by our ORM extension methods...

I haven't touched this part of the search engine in ages; I'm sure you can tell what a colossal effort it was at the time. I'm so glad I made those unit tests. They were indispensable while I was learning the weird nuances of lucene and they also let us run regressions during refactors like this.

Can you throw in some unit tests for your asin fixes?

rmcrackan:
RE unit tests, I'm thinking of: case-specific, leading numerals, and numerals-only IDs

rmcrackan:
> I took the opportunity to dispense with a lot of the regex

They were a carefully crafted solution that served me well for years, but I'm *never* surprised when regex is replaced with a specialized parser

rmcrackan:
Unrelated: WheelComboBox.axaml.cs has some notifications at the bottom of the PR. Looks like notes from the new version of avalonia

Mbucari:
> I'm sure you can tell what a colossal effort it was at the time.

I can. It was so hard to follow and I could feel your frustration reaching out across time and space.

> I'm so glad I made those unit tests.

I'm glad you made them to. They made my job easier, but it was still a royal PITA

> RE unit tests

Done.

> WheelComboBox.axaml.cs

Done.

PR #624, opened 2023-06-12
Fix #621

### LINKED ISSUE #621: Composer/Narrator [TCOM] tag not applying except in specific situations
**Describe the bug**
This is a follow up to #617. I already had the metadata setting checked, but none of the m4b files are getting the composer [TCOM] tag. I did some tests with the audiobook fix-up settings, and the results are below.

**To Reproduce**
1. My default settings: strip "unabridged" and download in original audio format settings chosen (secondary option not chosen), which results in the [TCOM] tag **not being applied**.
2. Making one change from my default - enabled split my books - [TCOM] tag is **not** applied. 
3. Making one change from my default - download as mp3 - [TCOM] tag is **not** applied.
4. Making two changes from my default - download as mp3 & split by chapter - [TCOM] tag **is** applied. 

**Expected behavior**
For the [TCOM] tag to apply in all cases listed above, particularly when split by chapter is not enabled and download original lossless setting is enabled. 

**Platform**
Windows 10, running latest version of Libation v10.3.1.1

**Log Files**
I have 2 log files (logging was set to verbose) - one with my default settings & the other with the chaptered & MP3 settings. I did not attach them at this time, as I am uncertain if it is safe to do so with verbose logging.

Mbucari:


Mbucari:
@rmcrackan This fixes #621. I know we just pushed a new release, but this is kind of important. ID3 tags were being written incorrectly. Some media players (like the one I was using for testing) could parse it, but others cannot.

PR #625, opened 2023-06-12
Refactor LibationSearchEngine

Mbucari:
Since I was just in there, I decided to refactor LibationSearchEngine.

There are two rule type, `LibraryBookRule` and `BookRule`, that implement `IIndexRule`. Each `IIndexRule` holds the delegate value getter, all field name aliases, and the field value type.

Rules are added to the `IndexRuleCollection`, and they're added and removed from the search index using `LuceneExtensions.RemoveRule()` and `LuceneExtensions.AddIndexRule()`.


Source/LibationSearchEngine/SearchEngine.cs
```diff
@@ -25,164 +22,47 @@ public class SearchEngine
...
         }
 
-        public static IEnumerable<string> GetSearchNumberFields()
+		// use these common fields in the "all" default search field
+		public static IndexRuleCollection FieldIndexRules { get; } = new IndexRuleCollection
         {
-            foreach (var key in numberIndexRules.Keys)
-                yield return key;
-        }
+            { FieldType.ID, Book => Book.AudibleProductId.ToLowerInvariant(), nameof(Book.AudibleProductId), "ProductId", "Id", "ASIN" },
```
rmcrackan, line 36: Wow, this is so much cleaner

Source/LibationSearchEngine/LuceneExtensions.cs
```diff
@@ -10,21 +10,58 @@ namespace LibationSearchEngine
...
+		}        
 
-        internal static void AddAnalyzed(this Document document, string name, string value)
-        {
-            if (value is not null)
-                document.Add(new Field(name.ToLowerInvariant(), value, Field.Store.YES, Field.Index.ANALYZED));
-        }
+        internal static void RemoveRule(this Document document, IIndexRule rule)
+		{
+			// fields are key value pairs. MULTIPLE FIELDS CAN POTENTIALLY HAVE THE SAME KEY.
```
rmcrackan, line 22: Thank you for retaining comments like these. They are hard won knowledge and I find they help considerably. Esp times like now years later when I've long since forgotten this gotcha

Source/LibationSearchEngine/BookRule.cs
```diff
@@ -0,0 +1,21 @@
+﻿using DataLayer;
```
rmcrackan, line 1: There's certainly nothing wrong with having a BookRule. I'm not clear on what it gains you. Since every LibraryBook has `.Book`

Source/LibationSearchEngine/SearchEngine.cs
```diff
@@ -25,164 +22,47 @@ public class SearchEngine
...
+            { FieldType.Raw, Book => Book.AudibleProductId, _ID_ },
+            { FieldType.String, Book => Book.Title, nameof(Book.Title), "ProductId", "Id", "ASIN" },
+            { FieldType.String, Book => Book.AuthorNames(), "AuthorNames", "Author", "Authors" },
+            { FieldType.String, Book => Book.NarratorNames(), "NarratorNames", "Narrator", "Narrators" },
+            { FieldType.String, Book => Book.Publisher, nameof(Book.Publisher) },
+            { FieldType.String, Book => Book.SeriesNames(), "SeriesNames", "Narrator", "Series" },
+            { FieldType.String, Book => string.Join(", ", Book.SeriesLink.Select(s => s.Series.AudibleSeriesId)), "SeriesId" },
+            { FieldType.String, Book => Book.CategoriesIds() is null ? null : string.Join(", ", Book.CategoriesIds()), nameof(Book.Category), "Categories", "CategoriesId", "CategoryId", "CategoriesNames" },
+            { FieldType.String, Book => Book.UserDefinedItem.Tags, TAGS.FirstCharToUpper() },
+            { FieldType.String, Book => Book.Locale, "Locale", "Region" },
```
rmcrackan, line 46: Ok, I'll bite. How the hell does it know to use the correct `Add` overload here for the Book vs LibraryBook func?

Source/LibationSearchEngine/BookRule.cs
Mbucari: I really didn't want to make it. This whole refactor would have been so much cleaner and more beautiful if I could have just used library book everywhere. The reason I didn't is because of how user-defined items are updated. Those methods take a Book, not a LibraryBook. I started looking up the stack to see if I could change all callers to use library book instead, but a lot of stuff would need to be changed. I decided the lesser of two evils was to do what I did here.

Source/LibationSearchEngine/SearchEngine.cs
Mbucari: Type inference from the methods called on the lambda arguments. I'll admit I was damn impressed too. Roslyn is pretty clever!

Source/LibationSearchEngine/BookRule.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace LibationSearchEngine;
+
+public class BookRule : IIndexRule
+{
+	public FieldType FieldType { get; }
+	public Func<Book, string> ValueGetter { get; }
+	public ReadOnlyCollection<string> FieldNames { get; }
+
+	public BookRule(FieldType fieldType, Func<Book, string> valueGetter, params string[] fieldNames)
```
rmcrackan, line 13: I recommend validation and sanitation on `fieldNames`. Something like

    ArgumentValidator.EnsureNotNull(fieldNames)
    fieldNames = fieldNames.Where(a => !string.IsNullOrWhiteSpace(a)).Select(a => a.Trim())
    if !fieldNames.Any() throw

Source/LibationSearchEngine/BookRule.cs
rmcrackan: Makes sense

Source/LibationSearchEngine/BookRule.cs
rmcrackan: Lemme know if you're intentionally not doing this so I'm not waiting on you before I merge.

Mbucari: I'm working on it now.

Source/LibationSearchEngine/BookRule.cs
Mbucari: We've talked about making such a change in various issues, and it probably won't happen, but if it does I think it will be a Many LibraryBook to One Book, so that would still be fine.

rmcrackan: LibraryBook is an ownership record of 1 book .: anything you can do to a book/user defined item you can do via LibraryBook. If we allow for duplicate books, it just won't be 1:1 in the other direction: a book could be owned by multiple library books; a library book still only owns 1 book. Even *if* we one day do expand to this (and 7 years later, this is nothing I've pursued for long), it would be trivial to have:

`void UpdateUserDefinedItem(this LibraryBook lb, var foo) => lb.Book.UpdateUserDefinedItem(foo)`

Source/LibationSearchEngine/SearchEngine.cs
rmcrackan: I've used VS for like 20 years and this is honestly blowing my mind. It just keeps getting better. Hot damn I love coding in C# + VS.

Source/LibationSearchEngine/BookRule.cs
rmcrackan: > We've talked about making such a change in various issues, and it probably won't happen, but if it does I think it will be a Many LibraryBook to One Book, so that would still be fine.

Yup, looks like we commented with the same idea at the same time. Great minds and all that

Source/LibationSearchEngine/SearchEngine.cs
Mbucari: > Hot damn I love coding in C# + VS.

+1000, which is what makes development in any other language fell like wading through knee-deep mud.

Source/LibationSearchEngine/LuceneExtensions.cs
```diff
@@ -10,21 +10,47 @@ namespace LibationSearchEngine
...
-        internal static void AddBool(this Document document, string name, bool value)
-            => document.Add(new Field(name.ToLowerInvariant(), value.ToString(), Field.Store.YES, Field.Index.ANALYZED_NO_NORMS));
+			foreach (var name in rule.FieldNames)
+			{
+				// fields are key value pairs and MULTIPLE FIELDS CAN HAVE THE SAME KEY.
+				// splitting authors and narrators and/or tags into multiple fields could be interesting research.
+				// it could allow for more advanced searches, or maybe it could break broad searches.
+
+				// all searching should be lowercase
+				// external callers have the reasonable expectation that product id will be returned CASE SPECIFIC
```
Mbucari, line 40: This is no longer true

rmcrackan: In what way?

Mbucari: https://github.com/rmcrackan/Libation/issues/618#issuecomment-1584924184

You expect it to be case insensitive, and that's now how the ID field works. Every id is converted to lowercase before indexing.

Issue #627, opened 2023-06-13
Option to overwrite existing files when saving books


pixil98:
**Is your feature request related to a problem? Please describe.**
When Libation moves a book's files to the books folder, it will append a number to the end if the files already exist. I'd like an option to overwrite existing files if they already exist. This came up with the problem I'm having there it's occasionally writing 10MB files out. I'd like to have it just redownload those and overwrite the existing bad ones. Additionally, it would allow for a somewhat heavy handed way to in place update the files if something like the moov atom changes again.

**Describe the solution you'd like**
I'd like an option to have the files overwrite existing files when saving books.

**Describe alternatives you've considered**
I have to clean things up manually more than I'd like.

rmcrackan:
Included in [Libation v10.4](https://github.com/rmcrackan/Libation/releases/latest)

Issue #626, opened 2023-06-13
CLI output log to stdout/stderr


pixil98:
**Is your feature request related to a problem? Please describe.**
I'm trying to debug some issues where my weekly job that downloads new books is getting errors and writing out m4b files that are only 10MB. Since this is running in a docker image, it's hard to debug because the logging all to a file that I don't have access to after the docker image finishes running.

**Describe the solution you'd like**
It would be nice if there was an option to for the cli to write it's logs to stdout and stderr. This is the expectation with a docker image and would allow those logs to be scraped and review easily.

**Describe alternatives you've considered**
There are some ways to trick a program into writing to stdout, depending on how it interacts with log files. I haven't actually tried using them yet.

rmcrackan:
I haven't tried this, but you might be able to just add this into your Settings.json in the WriteTo array

```
      ,{
        "Name": "Console",
        "Args": {
          "outputTemplate": "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level:u3}] (at {Caller}) {Message:lj}{NewLine}{Exception} {Properties:j}"
        }
      }
```

rmcrackan:
Included in [Libation v10.4](https://github.com/rmcrackan/Libation/releases/latest)

Issue #629, opened 2023-06-13
Store Different Libraries in Different Locations


CaptainLexington:
I like helping my less-technical friends liberate their media, and for some of them this means using my own computer to download their libraries. Right now, their library will save to the same location as my library. I could set each library to manual updates only, and then manually switch the output dir each time I update, but it would be nicer if each account had a separately-configured output directory.

CLHatch:
> I like helping my less-technical friends liberate their media, and for some of them this means using my own computer to download their libraries. Right now, their library will save to the same location as my library. I could set each library to manual updates only, and then manually switch the output dir each time I update, but it would be nicer if each account had a separately-configured output directory.

I know you are looking for a way to set up a new library, with a new location, but you can  SORT OF do what you want already with the renaming patterns.  This is my folder naming pattern currently:

`<account>\<if podcast->Podcasts\<if series-><series>\<-if series>(<pub date>) - <title> [<id>] {<narrator[separator(; )]>}<-if podcast><!if podcast->AudioBooks\<author[separator(; )]>\<if series-><series>\<series#[2]> - <-if series>(<year>) - <title> [<id>] {<narrator[separator(; )]>}<-if podcast>`

Notice the first part is `<account>`, so the books from each person's account gets put into their own folder (the folder name ends up being the owner's email address).

CaptainLexington:
Actually, that's pretty much exactly what I wanted! Probably I would only request to be able to use the account nickname instead of the email address, but I can use this indefinitely

CharlieRussel:
@CLHatch -- That's actually a quite slick solution. I like it. Thanks.

CLHatch:
> Actually, that's pretty much exactly what I wanted! Probably I would only request to be able to use the account nickname instead of the email address, but I can use this indefinitely

I had the same idea, about an option to use the account nickname.  Would be nice.

rmcrackan:
For an easy fix, the `<account>` idea above is great! If you want more control:

Libation isn't made to be multi-user. However, it **is** made with extreme flexibility around how to manage files. If you're comfortable mucking around in the settings files, you can set up completely separate installations of Libation for each user.

I'll detail this for Windows although it's very similar for Mac and Linux. Before you manually edit or move settings files, please **back up everything first**.

[...]

PR #631, opened 2023-06-14
Add features #626 and #627 and Fix #628

### LINKED ISSUE #628: Output file validation
**Is your feature request related to a problem? Please describe.**
I'm running into a problem where book being liberated are encountering errors, and it results in writing out partial m4b files. I'm still tracking down the cause of that, but it would be nice if there was the ability to validate the file was written correctly and not mark the book as liberated if it wasn't.

**Describe the solution you'd like**
I'd like the the ability to have Libation validate that the file was written correctly. This is more of a vague enhancement request. There are a couple points in the process that feel like they could be validated, when the book is initially downloaded and when the book is moved to the final destination. For the first one, it would largely depend on what kind of metadata you can get from audible. For the second one it could check that the file has the same checksum as the one in the temp folder.

**Describe alternatives you've considered**
I've considered writing a script to run periodically and flag books that are suspiciously small files.

Mbucari:
## [Add feature](https://github.com/rmcrackan/Libation/commit/46f6ba17101c70119c14284d806a4d3a60b7c162) #627 [and fix bug](https://github.com/rmcrackan/Libation/commit/46f6ba17101c70119c14284d806a4d3a60b7c162) #628

- Add setting to overwrite files when moving from temp folder to Books folder (#627)
- Do not set `LiberatedStatus` to `LiberatedStatus.Liberated` if moving files threw an exception (#628)

## [Write error info to StdErr](https://github.com/rmcrackan/Libation/commit/ec9854212a509013de338779cdf53b9739808c0d) #626

### DIFF
--- a/Source/LibationCli/Options/_OptionsBase.cs
+++ b/Source/LibationCli/Options/_OptionsBase.cs
@@ -18,11 +18,11 @@ namespace LibationCli
 			{
 				Environment.ExitCode = (int)ExitCode.RunTimeError;
 
-				Console.WriteLine("ERROR");
-				Console.WriteLine("=====");
-				Console.WriteLine(ex.Message);
-				Console.WriteLine();
-				Console.WriteLine(ex.StackTrace);
+				Console.Error.WriteLine("ERROR");
+				Console.Error.WriteLine("=====");
+				Console.Error.WriteLine(ex.Message);
+				Console.Error.WriteLine();
+				Console.Error.WriteLine(ex.StackTrace);
 			}
 		}
 
--- a/Source/FileLiberator/ConvertToMp3.cs
+++ b/Source/FileLiberator/ConvertToMp3.cs
@@ -73,8 +73,23 @@ namespace FileLiberator
 						}
 						else
 						{
-							var realMp3Path = FileUtility.SaferMoveToValidPath(mp3File.Name, proposedMp3Path, Configuration.Instance.ReplacementCharacters, "mp3");
-							OnFileCreated(libraryBook, realMp3Path);
+							try
+							{
+								var realMp3Path
+									= FileUtility.SaferMoveToValidPath(
+										mp3File.Name,
+										proposedMp3Path,
+										Configuration.Instance.ReplacementCharacters,
+										extension: "mp3",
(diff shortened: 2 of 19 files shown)


Source/LibationCli/Options/_OptionsBase.cs
```diff
@@ -18,11 +18,11 @@ public async Task Run()
 			{
 				Environment.ExitCode = (int)ExitCode.RunTimeError;
 
-				Console.WriteLine("ERROR");
-				Console.WriteLine("=====");
-				Console.WriteLine(ex.Message);
-				Console.WriteLine();
-				Console.WriteLine(ex.StackTrace);
+				Console.Error.WriteLine("ERROR");
```
rmcrackan, line 21: Good call. I always forget `Console.Error` exists

Mbucari:
@rmcrackan Please see latest commit: [Fix Stupid](https://github.com/rmcrackan/Libation/pull/631/commits/7a9c6720c73f8a1ff3209342e90e03a4fd9df154)

rmcrackan:
Since you've got everything open, do you want to [throw this in there too](https://github.com/rmcrackan/Libation/issues/629#issuecomment-1590434045)? I feel like it should be 1 line of code plus another line of markdown.

I was thinking this morning: a completely masturbatory backend task could be a post-build runner which auto-updates stuff like nuspec and the tags readme.

rmcrackan:
> Fix Stupid

There's a saying around here "you can't fix 'stupid'" :P

Mbucari:
> Since you've got everything open, do you want to [throw this in there too](https://github.com/rmcrackan/Libation/issues/629#issuecomment-1590434045)?

Sure thing.

> I was thinking this morning: a completely masturbatory backend task could be a post-build runner which auto-updates stuff like nuspec and the tags readme.

I'll keep that in my back pocket for when boredom strikes again.

> you can't fix 'stupid'

I'm an optimist :D

rmcrackan:
> I'm an optimist :D

At every poker table, there is a fool. And if you can't find him, I've got some bad news...

Mbucari:
#629 is done

Issue #638, opened 2023-06-19
Option to set timestamps for created directories and files


uffefl:
**Sorting by "Date Modified" doesn't make much sense on an initial download**
After liberating my entire Audible library in one go, I'm sad to see that all the directories have creation and modification timestamps equal to the time they were downloaded and not the time that I purchased them. This makes it harder to find recent purchases than it should be. Especially on the initial download since a lot of books were downloaded in a semi-random order.

I'm sure that after using the program for a while the dates will make a little more sense since I will probably download new titles immediately after buying them. Even more so with the ease of Libation compared to my previous inAudible process.

However some times I know there will be a lag where I will just use the Audible app for whatever reason and then come back to downloading multiple books in a single sitting. In this case it would be nice if the timestamps on the files and folders created reflected the purchase history.

**Suggestion: Make an option to set timestamps on created files and folders**
Make an option to set the timestamps on created files and folders (checkbox in Download/Decrypt settings?) to reflect purchase date would be nice.

Maybe an option to reflect the publication date as an alternative as well? I don't think I would be using this though.

rmcrackan:
Included in [Libation 10.4.2](https://github.com/rmcrackan/Libation/releases/latest)

Issue #637, opened 2023-06-19
Handling of multiple spaces in the file/directory naming templates


uffefl:
**Handling multiple spaces/other replacement chars in file/directory names**
In the illegal character replacement dialog I have several rules that replace illegal characters with spaces, or with strings that contain spaces. This sometimes lead to a replaced character space next to a natural space, which creates unsightly double-spaces in the final name.

The best example I can give is the typical `Book title: Book subtitle` pattern that most of my books follow. I would like to replace `:` with ` - `, to ensure spacing on both sides of the (now no longer present) colon. However this turns into `Book title -  Book subtitle` which looks dumb. My workaround for now is to replace with ` -`, but obviously this would fail if the colon doesn't have a trailing space. It would also look dumb if there was a space before the colon (eg. French language maybe? I don't have any books with this issue, so this is speculative).

(I'm assuming this also applies to other characters than spaces. Eg. the default underscore replacement could have the same issue, but I have not tried this.)

**Possible solution?**
I think the simplest solution would be to have an option (checkbox?) to simply trim the whitespace in the final string. This obviously wouldn't work with other replacement characters than spaces.

Making a general solution that handles any replacement text is a bit trickier though. What if the replacement text is more than one character, eg. the colon example above.

rmcrackan:
Included in [Libation 10.4.2](https://github.com/rmcrackan/Libation/releases/latest)

PR #642, opened 2023-06-25
Bug fixes and minor features

Mbucari:
## [Add options to set file created/modified timestamps](https://github.com/rmcrackan/Libation/commit/ac87d706137d1e8cb4ebd76a103d97eb3c34d5cb) #638

Two new settings:
Libation_4VqvGhQlhn

## [Remove multispace instances from template filenames](https://github.com/rmcrackan/Libation/commit/1b09b1fd481221b141a75fa12da5c5de051a7b2d) #637

Replace all instances of two or more spaces with a single space. The reason I did before concatenation of the parts is because the concatenation process involves truncating parts to fit the file system length limits. If you first concatenate and then remove multispaces you risk losing more data than is necessary.


rmcrackan:
Nice batch of upgrades

PR #646, opened 2023-06-26
feat: add Book subtitle capturing so TitleShort reflects titles better

Alanoll:
I ran into some books on Audible that contained semi-colons in their actual Titles (e.g. Star Trek: Ashes of Eden by William Shatner).

When using TitleShort in the naming, it was removing a portion of the title. At least in my library, it looks like the Semi-Colon for others was added due to combining Title and Subtitle early on, and then later when templating resplitting it.

I added Subtitle to be captured when scraping, so now TitleShort is just Title as on Audible. Title still retains the {Title}: {Subtitle} syntax.


Mbucari:
@Alanoll This is really great!

One thing I'm not sure about is if we should still try splitting titles containing colons.
Looking through my library I found 8 books and ~2500 podcasts whose titles contain a colon and do not have a subtitle. Here are the books.

```
The Handmaid's Tale: Special Edition
Get In Get Out: An Amish, Sci-fi, Gender Bender Adventure
The Other Side of History: Daily Life in the Ancient World
Understanding Literature and Life: Drama, Poetry and Narrative
War, Peace, and Power: Diplomatic History of Europe, 1500-2000
Death, Dying, and the Afterlife: Lessons from World Cultures
Early Christianity: The Experience of the Divine
Tin Man: A Galaxy's Edge Prequel
```

The podcast episodes folloing this pattern are usually in the form of `[Episode #]: [Episode Title]`.

Currently, all of these examples are split so that `<titleshort>` is everything before the colon, but in this PR `<titleshort>` is everything including the colon. My first though it to split these titles in the DB. Something like:

```C#
if (item.Title.Contains(':') && string.IsNullOrEmpty(item.Subtitle))
{
  var firstIndex = item.Title.IndexOf(':');
  book.Title = item.Title.Substring(0, firstIndex).Trim();
  book.Subtitle = item.Title.Substring(firstIndex + 1).Trim();
}
else
{
  book.Title = item.Title;
  book.Subtitle = item.Subtitle;
}
```

Alanoll:
I think I'll try and scrape a couple of API calls to see how the data is formatted.

@Mbucari Can you provide a link to the one of the podcast episodes on the Audible site?

Mbucari:
> ETA: Actually, I wonder if that will break the purpose of what I'm trying to fix. In my scenario, `Star Trek: Ashes of Eden`, that IS the title. It's exactly the same scenario as one of your's.

Your right, it is the title. It just differs from how it's been done. I have no preference in this matter and would be fine with the change so long as @rmcrackan is. He's the boss, after all! Let's wait for him to weigh in.

> I think I'll try and scrape a couple of API calls to see how the data is formatted.

FYI, `AudibleApi.Common.Item.SourceJson` contains the verbatim JSON returned by the API. As far as we're aware, we're querying all available data for each item. If it's not in that JSON then it probably doesn't exist.

> Can you provide a link to the one of the podcast episodes on the Audible site?

https://www.audible.com/pd/Chapter-84-Taboo-Tradeoffs-Aftermath-2-Podcast/B0C37Q4482

Mbucari:
> ETA: Actually, I wonder if that will break the purpose of what I'm trying to fix. In my scenario, `Star Trek: Ashes of Eden`, that IS the title. It's exactly the same scenario as one of your's.

You know we can solve this whole thing by keeping everything you did and just adding a couple of new tags

Keep the `<title>` and `<title short>` definitions and add new tags for `<subtitle>` and `<title with subtitle>`.

I'm ambivalent about `<title with subtitle>` because on the one hand users would format `<title>: <subtitle>` themselves, but on the other hand if subtitle is null then the file name will have a stray colon after the title.

Alanoll:
Just thinking through, we have the following possible formats for titles:

1. Title with no semi-colon, no subtitle (e.g. `Greenlights`)
    1. Easy `title` parse
    2. https://www.audible.com/pd/Greenlights-Audiobook/0593294181
2. Title with a semi-colon, no subtitle (e.g. `Star Trek: The Ashes of Eden`)
    1. Should be parsed as `title`
    2. https://www.audible.com/pd/Star-Trek-The-Ashes-of-Eden-Adapted-Audiobook/B002VA94PY
3. Title with a semi-colon, and should be subtitle (e.g. `Tin Man: A Galaxy's Edge Prequel`)
    1. Should be parsed as `title` and `subtitle`
    2. https://www.audible.com/pd/Tin-Man-A-Galaxys-Edge-Prequel-Audiobook/1774242834
4. Title with no semi-colon, with subtitle (e.g. Savage Wars: Galaxy's Edge: Savage Wars, Book 1)
    1. https://www.audible.com/pd/Savage-Wars-Audiobook/B084KR193P
5. Title with a semi-colon, with subtitle
    1. I don't know an example here, but complicates things

Scenario `2` complicates things, as it SHOULD be the filename when using `title`, but **Tin Man** should be the title from `3`

Mbucari:
> 5. Title with a semi-colon, with subtitle
>     1. I don't know an example here, but complicates things

I've only found one example: https://www.audible.com/pd/Bond-of-a-Dragon-Publishers-Pack-Audiobook/1774246171

> Scenario `2` complicates things, as it SHOULD be the filename when using `title`, but **Tin Man** should be the title from `3`

Audible's data has lots of little inconsistencies like this, and we just need to make a choice about how to handle it knowing that it won't be 100%.

I see 2 choices:

1. Use what you've done and let `<title short>` be the `Item.Title` and `<title>` be `{Item.Title}: {Item.Subitle}`.
    1. This is a departure from the current system.
    2. Case `3` will be wrong.
2. Split Title at colons and use the first part as the title and the second part as the subtitle.
    1. Maintains current behavior
    2. Case `2` will be wrong.
    3. Case `5` will be wrong, but we've only found one example of case `5` and I think that it's actually a case `3` will extraneous data in the Subtitle.

Like I said, I have no preference. Waiting for @rmcrackan

rmcrackan:
I like where this is going in theory. After the edge cases are considered, I can't tell if we're solving the problem, or squeezing the balloon to create different problematic edge cases.

> This is a departure from the current system.

Whatever we do, we can't break people's existing custom templates.

Basically, I'm open to suggestions which are backward compatible and which aren't creating new problems.

Mbucari:
> > This is a departure from the current system.
> 
> Whatever we do, we can't break people's existing custom templates.

I hate to quibble over semantics, but what do you consider breaking? In my option 1 above the template would still evaluate to a valid string, but what those strings are would be different after the update.

rmcrackan:
> those strings are would be different after the update

On the surface that sounds like a breaking change to me. Like you said earlier though: audible is rife with inconsistencies so a subtle change probably won't bother anyone.

> <title short> be the Item.Title and <title> be {Item.Title}: {Item.Subitle}

I dislike this. Title isn't a short title; it's just the title. "Title with subtitle" captures the intent better and is very clear, but what a terrible syntax. "Title extended"? "Title expanded"? After brainstorming, it's hard to come up with anything more clear than "title with subtitle".

CLHatch:
> > those strings are would be different after the update
> 
> On the surface that sounds like a breaking change to me. Like you said earlier though: audible is rife with inconsistencies so a subtle change probably won't bother anyone.
> 
I suppose it all depends on how you look at it.  It's a "breaking change" in that  the output is different.  But it could be looked at as a "fixing change" (bugfix) since it appears it is more accurate in how Audible is calling things.

ddirty830:
I guess the real question: what is the source of truth?

If it's Audible, then we have to accept that it's data pattern is right. Meaning their Title field is the title, and their Subtitle field is subtitle. No additional parsing.

Then then translates to `title` being `{Title}: {Subtitle}` and `titleshort` being `{Title}`. It's a "breaking" change to what's currently expected with regard to `titleshort` but respects the data as is.

Could also introduce an additional: `albumtitle` which is `{Title}` and then define `titleshort` as `{Title upto first colon}`

Alanoll:
I guess the real question: what is the source of truth?

If it's Audible, then we have to accept that it's data pattern is right. Meaning their Title field is the title, and their Subtitle field is subtitle. No additional parsing.

Then then translates to `title` being `{Title}: {Subtitle}` and `titleshort` being `{Title}`. It's a "breaking" change to what's currently expected with regard to `titleshort` but respects the data as is.

Could also introduce an additional: `albumtitle` which is `{Title}` and then define `titleshort` as `{Title upto first colon}`

Mbucari:
> Title isn't a short title; it's just the title.

Yeah, I agree with that.

> I guess the real question: what is the source of truth? I'm inclined to take Audible at their word, even though we know it's wrong some of the time. After this discussion I've developed a preference for keeping `<title>` and `<title short>` definitions and adding new tags for `<subtitle>` and <title and subtitle>` (or whatever the hell you want to call it).

CLHatch:
> > Title isn't a short title; it's just the title.
> 
> Yeah, I agree with that.
> 
> > I guess the real question: what is the source of truth?
> 
> I'm inclined to take Audible at their word, even though we know it's wrong some of the time. After this discussion I've developed a preference for keeping `<title>` and `<title short>` definitions and adding new tags for `<subtitle>` and <title and subtitle>` (or whatever the hell you want to call it).

But isn't the definition of `<title>` basically what the `<title and subtitle>` would be?  BTW, perhaps `<title long>` would be a less wordy name for it.

Mbucari:
> But isn't the definition of `<title>` basically what the `<title and subtitle>` would be?

I don't think so. I think the title is the title, especially when so many of the subtitles are in the vein of "Book n of Series X"

> BTW, perhaps `<title long>` would be a less wordy name for it.

I like that. I like that a lot.

> Edit: I just realized that it seems you were talking about what to call the internal names for them, not the tags we would use in our naming patterns.

Nope, you had it right. I was still not sure what to call title + subtitle, but I think you hit the nail on the head with `<title long>`.

<hr>

So it sounds like we can settle this? Revert `<title>` and `<title short>` to their original definitions and add `<subtitle>` and `<title long>` tags?

CLHatch:
> > But isn't the definition of `<title>` basically what the `<title and subtitle>` would be?
> 
> I don't think so. I think the title is the title, especially when so many of the subtitles are in the vein of "Book n of Series X"
> 
> > BTW, perhaps `<title long>` would be a less wordy name for it.
> 
> I like that. I like that a lot.
> 
> > Edit: I just realized that it seems you were talking about what to call the internal names for them, not the tags we would use in our naming patterns.
> 
> Nope, you had it right. I was still not sure what to call title + subtitle, but I think you hit the nail on the head with `<title long>`.
> 
> So it sounds like we can settle this? Revert `<title>` and `<title short>` to their original definitions and add `<subtitle>` and `<title long>` tags?

Unless I'm mistaken on the current definition of `<title>` and `<short title>`, `<short title>` is currently the `<short title>` is basically the title without the subtitle, and `<title>` is the title plus the subtitle.  So really all this change would be doing is changing the `<title>` to be what the current `<short title>` is, and making a new `<title long>` to be what the current `<title>` is.  Or am I wrong on what the current usage is?

Mbucari:
> Or am I wrong on what the current usage is?

I think we're talking about two different things here. When we query book information from audible, it returns a field named `title` and another field named `subtitle`. Currently libation only stores and makes use of the `title` field, and the `<short title>` is the the title up to the first colon.

The primary change in this PR is to store and make use of the `subtitle` field returned by audible, and we're trying to decide how to use that data without it being confusing. 

> `<short title>` is currently the title without the subtitle

I think this is how most people understand the current system, so adding a new subtitle tag is bound to cause confusion.

CLHatch:
> > Or am I wrong on what the current usage is?
> 
> I think we're talking about two different things here. When we query book information from audible, it returns a field named `title` and another field named `subtitle`. Currently libation only stores and makes use of the `title` field, and the `<short title>` is the the title up to the first colon.
> 
> The primary change in this PR is to store and make use of the `subtitle` field returned by audible, and we're trying to decide how to use that data without it being confusing.
> 
> > `<short title>` is currently the title without the subtitle
> 
> I think this is how most people understand the current system, so adding a new subtitle tag is bound to cause confusion.

Ah, ok.  So I was right in that we were partially talking about the internal usage, plus how to combine those for the "external usage".

Alanoll:
> Currently libation only stores and makes use of the title field
Currently, Libation stores the Title and Subtitle as Title. The token `title` is then the stored Title and Subtitle (if present), and `titleshort` is everything up to the first colon.

This change currently changes it so we store Title and Subtitle separately. The definitions of the tokens are the same `title` is still Title + Subtitle, but `titleshort` is now just the Title returned from Audible directly.

CLHatch:
> > Currently libation only stores and makes use of the title field
> 
> Currently, Libation stores the Title and Subtitle as Title. The token `title` is then the stored Title and Subtitle (if present), and `titleshort` is everything up to the first colon.
> 
> This change currently changes it so we store Title and Subtitle separately. The definitions of the tokens are the same `title` is still Title + Subtitle, but `titleshort` is now just the Title returned from Audible directly, no additional string manipulation is occuring.

Ah,  makes sense. So that way, the user can use the same tags they had been, but get the "more accurate" `<title short>`, for those titles that actually have colons in them.  This would mean there is no real need for a `<title long>`, because that is still what `<title>` outputs.  The "breaking change" would be to make `<title>` output the title without the subtitle, and have `<title long>` output the title plus the subtitle.  Which is still a possibility, since it reflects that is being received.

Mbucari:
> > Currently libation only stores and makes use of the title field
> 
> Currently, Libation stores the Title and Subtitle as Title. The token `title` is then the stored Title and Subtitle (if present), and `titleshort` is everything up to the first colon.

My mistake, thank you for the correction.

rmcrackan:
> Revert `<title>` and `<title short>` to their original definitions and add `<subtitle>` and `<title long>` tags?

This sounds like the answer for tagging. Do we still need an additional db field?

wtanksleyjr:
So we get <title> and <subtitle> from Audible, and report them accurately.
I can see a need to process data after that, since Audible can do an
incredible number of silly things like including the subtitle in the title
(so you want a "short title" that removes it).

My old Python code for this took opportunities like searching for the
subtitle in the title (and vice versa) to see if they copied the data. The
problem is always that we can get smarter about the heuristics we use, so
they're going to change over time. So what if we slap a name on those
processed tags that makes it possible to document that they might change
with different Libation releases? Just for the sake of starting the
discussion, suppose the new tags were <processed title> and <processed
subtitle>. Likewise, my old Python program had to determine the most likely
series name to use (because Audible sometimes hides the series in the
subtitle, and sometimes provides many series names and we can only use
one). So <processed series name> and <processed series index>.

Don't think "processed" is the best idea, it's just AN idea. The point is
to clearly mark when we're doing something "magic" so that the user who
reads the docs will know we're doing work to get that data.

-Wm

Alanoll:
How about this solution:

We keep the Title/Subtitle as returned from Audible in the database fields `Title` and `Subtitle` (this one being added by my PR). 
We keep the tag `title` defined `The full title and subtitle` as has been
We keep the tag `titleshort` defined as `The title and subtitle up to the first semi-colon` as has been (so we'd combine the two values, and then find first semi-colon and trim everything after.
I add two new tags `audible title` which is the raw Audible returned Title, and `audible subtitle` which is the raw Audible returned Subtitle.

Then nothing changes for for current users using tags, but allows others to get untouched values as well. We can then add in future some form of further processed tags in addition to the four here.

Mbucari:
> How about this solution:
> 
> We keep the Title/Subtitle as returned from Audible in the database fields `Title` and `Subtitle` (this one being added by my PR). We keep the tag `title` defined `The full title and subtitle` as has been We keep the tag `titleshort` defined as `The title and subtitle up to the first semi-colon` as has been (so we'd combine the two values, and then find first semi-colon and trim everything after. I add two new tags `audible title` which is the raw Audible returned Title, and `audible subtitle` which is the raw Audible returned Subtitle.
> 
> Then nothing changes for for current users using tags, but allows others to get untouched values as well. We can then add in future some form of further processed tags in addition to the four here.

I like this solution. TBH there's no problem with adding more tags. The list gets longer when editing, but who cares?

<hr>
As an aside, I just had an idea for a new conditional tag: `<if hasvalue[tag]-><if hasvalue>`
Would let users check for null tag values.

Mbucari:
> > Revert `<title>` and `<title short>` to their original definitions and add `<subtitle>` and `<title long>` tags?
> 
> This sounds like the answer for tagging. Do we still need an additional db field?

@rmcrackan 
I think so. Without separate fields we're relying on string parsing to determine title and subtitle, but as shown in the examples above this is unreliable. Many titles have a colon in the 'Title' property and no defined subtitle. One even has a colon in the title AND a subtitle, so the concatenation has 2 colons. Users (specifically @Alanoll) wants to distinguish between what Audible says is a Title and what it says is a Subtitle, and the only reliable way to do that is with separate DB entries.

rmcrackan:
> I like this solution. TBH there's no problem with adding more tags. The list gets longer when editing, but who cares?

As long as the documentation is up to date and clear, I agree. In the case of these new tags (which required pages of discussion here to nail down nuance), the documentation is particularly important.

> `<if hasvalue[tag]-><-if hasvalue>`

To me this sounds like a debugging headache waiting to happen. If you want to tackle it though, on your own head be it lol

> > This sounds like the answer for tagging. Do we still need an additional db field?

> I think so

Fair enough.

I like how this review went down; it has my blessing. Ping me when you're done discussing other new tags.

Mbucari:
@Alanoll

So TODO is:
- Keep the title short tag the same (title substring up to first colon)
- make the \<title\> tag evaluate to {Title}: {Subtitle}
- add the 2 new tags for title and subtitle
- update [the docs](https://github.com/rmcrackan/Libation/blob/master/Documentation/NamingTemplates.md)

Thanks!

rmcrackan:
@Mbucari It's hard to tell from github's ui, but it looks like you created a duplicate entry for `Title` instead of editing.

Mbucari:
> @Mbucari It's hard to tell from github's ui, but it looks like you created a duplicate entry for `Title` instead of editing.

Yeah, I was trying to figure out how to make commits to someone else's PR. I think I got it now. Take a look.

Source/LibationFileManager/Templates.cs
```diff
@@ -274,8 +276,10 @@ private static void RemoveSpaces(List<string> parts)
 		{
 			new PropertyTagCollection<LibraryBookDto>(caseSensative: true, StringFormatter)
 			{
-				{ TemplateTags.Title, lb => lb.Title },
+				{ TemplateTags.Title, lb => lb.TitleWithSubtitle },
 				{ TemplateTags.TitleShort, lb => getTitleShort(lb.Title) },
```
Alanoll, line 280: To maintain compatibility with the old tag, should this get `getTitleShort(lb.TitleWithSubtitle)`? I think it'd end up the same in 90% of cases, but old `lb.Title` was with Subtitle.

Alanoll:
This looks like you've gone through and made/corrected all the changes. Anything else that needs to get gone in this?

Source/LibationFileManager/Templates.cs
Mbucari: > old `lb.Title` was with Subtitle.

True, but they were concatenated with a colon so a substring on just the title will always return the title short.

Mbucari:
> Anything else that needs to get gone in this?

I don't think so, I just want to give @rmcrackan a chance for a final review.

Source/DataLayer/EfClasses/Book.cs
```diff
@@ -101,20 +102,25 @@ public class Book
...
+            UpdateTitle(title, subtitle);
+			Description = description?.Trim() ?? "";
             LengthInMinutes = lengthInMinutes;
             ContentType = contentType;
 
             // assigns with biz logic
             ReplaceAuthors(authors);
             ReplaceNarrators(narrators);
 		}
+        public void UpdateTitle(string  title, string subtitle)
```
rmcrackan, line 114: Thanks to which ever of you did this. I like that updating title and subtitle: A: has to be done together and B: is really intentional. Title and Subtitle should typically not change. Using this rather than casual property assignments makes this more clear.

rmcrackan:
This looks great. Before I merge, has anyone actually tested going from the earlier version of the db, doing this migration, then doing a scan? Are their any 'gotcha's I need to look out for?

Mbucari:
> Are their any 'gotcha's I need to look out for?

I did test that case and it did work. However, I wouldn't mind you checking again. I won't be in a position to do it myself until tomorrow.

Mbucari:
@rmcrackan I tested it again on both classic and chardonnay, and everything works as expected.

rmcrackan:
Thanks!! Released: https://github.com/rmcrackan/Libation/releases/tag/v10.4.3

PR #659, opened 2023-07-07
Fix classic scaling on high dpi displays

Mbucari:
Finally fixed all high DPI scaling issues. Some of it is a bit hacky because DataGridView is broken and [nobody plans to do anything about it](https://developercommunity.visualstudio.com/t/datagridview-does-not-scale-column-widths-when-aut/807917), but thankfully it didn't require too much change since we are already manually setting column widths.

I've tested it on both Normal and High DPI displays and it looks identical on both.


rmcrackan:
Thanks!

Issue #666, opened 2023-07-10
New Image Size


OldScotsman:
Firefox on Windows 10

I downloaded the new version (???) today on  Firefox on Windows 10 and there are 9 book images viewable.  Has this been reduced?  I feel there used to be more.  Is there a way to try the previous version like I used to be able to by looking for them under "Classic-Libation.?.?.?-windows-classic.zip" references which I can't find any more.  I suppose what I am saying is there a way to undo an upgrade.
All the best.
Bill

[Log202307_004.zip](https://github.com/rmcrackan/Libation/files/12000498/Log202307_004.zip)

OldScotsman:
I can no longer log in and get this message

[Log202307_005.zip](https://github.com/rmcrackan/Libation/files/12000616/Log202307_005.zip)


SplitterDistance must be between Panel1MinSize and Width - Panel2MinSize.

   at System.Windows.Forms.SplitContainer.set_SplitterDistance(Int32 value)
   at System.Windows.Forms.SplitContainer.ApplyPanel2MinSize(Int32 value)
   at LibationWinForms.Form1.Configure_ProcessQueue() in D:\a\Libation\Libation\Source\LibationWinForms\Form1.ProcessQueue.cs:line 19
   at LibationWinForms.Form1..ctor() in D:\a\Libation\Libation\Source\LibationWinForms\Form1.cs:line 52
   at LibationWinForms.Program.Main() in D:\a\Libation\Libation\Source\LibationWinForms\Program.cs:line 74


Thank you.
Hope you can help
Bill

Mbucari:
@OldScotsman 
> there are 9 book images viewable. Has this been reduced

If I understand you correctly, you're saying that when you open libation you can only see 9 rows of books where before you could see many more? And let me guess, you have a 4K monitor?

If so, this was a recent "fix" (fix in quotes because one man's fix is another man's new bug). You can still scroll down to see the rest of your books, right?

>  is there a way to undo an upgrade.

Yes, you can download old releases [here](https://github.com/rmcrackan/Libation/releases). Not that, if I correctly understand your problem, this behavior will not be reverted in future releases and older versions of Libation may stop working at any time when Audible makes changes to their service.

If you want everything to be "smaller", you can change your computer's display scaling. [See these instructions](https://support.microsoft.com/en-us/windows/view-display-settings-in-windows-37f0e05e-98a9-474c-317a-e85422daa8bb).

<hr>
> I can no longer log in and get this message...

[...]

PR #667, opened 2023-07-10
Fix setting Panel2MinSize min width bug (#666)

Mbucari:


rmcrackan:
Can I release now? Or should I wait?

Mbucari:
Go for it

Issue #692, opened 2023-08-01
Windows Classic Settings scaling layout


rmcrackan:
This has always been bad between my work monitors (100% scaling) and my weird laptop screen (250%). Before you fixed the scaling though it wasn't worth mentioning. Now that scaling is fixed, hopefully this is possible to resolve.

100% scaling:

![libation settings 100 percent](https://github.com/rmcrackan/Libation/assets/22817130/d35d26ae-d3b2-47f5-b37e-35c9d0458545)

250% scaling:

![libation settings 250 percent](https://github.com/rmcrackan/Libation/assets/22817130/5b705c30-3844-4b60-b92f-9e728cf1c0ef)

Mbucari:
I've had problems with user controls from the beginning. For some reason they don't seem to obey the same scaling rules as built-in winforms controls. Or they do and I'm not correctly defining some property in the user control. Probably the latter.

Mbucari:
So this appears to be a known scaling issue with anchors in user controls in .NET Core winforms. 

[This issue](https://github.com/dotnet/winforms/issues/6381) is kind of hard to follow, but it looks like they thought they fixed it in 7, some people still had problems in 7, and now they think they've fixed it in 8 preview.

So for now I think the only "fix" is to not use anchors in user controls and instead resize child controls manually inside `OnResize`

rmcrackan:
Thanks for the research.

Closed because screw it. It's been an issue since forever and maybe .net8 will fix it.

Mbucari:
I was able to fix this without too much trouble.

In `DirectorySelectControl` I used Docking to dock Top and Bottom instead of anchoring.

Then in `DirectoryOrCustomSelectControl` I just need to manually position the button and change the textbox and DirectorySelectControl widths.

rmcrackan:
Sweet, thanks!

rmcrackan:
Fixed in [version 10.6.4](https://github.com/rmcrackan/Libation/releases/latest)

PR #694, opened 2023-08-01
Fix DPI scaling bug (#692)

Mbucari:
Do me a favor and run this through it paces, will you?

I'd tested it on 100% and 175% scaling and get good results on both, but who knows what happens at higher.


rmcrackan:
Looks great at 100% and 250%. Thanks again!

Issue #695, opened 2023-08-02
Frozen "Audible Login" dialog on first import


mattalexx:
**Describe the bug**
I was following the [very cool] tour and clicked the "Import > Scan Library" menu option. A dialog appeared as an outline only, with a transparent body, and the application froze. I can move the main window but the "Import" dropdown stays frozen in place. I can move the dialog, but I cannot close it.

**To Reproduce**

1. Start First Time User tour.
2. Start first import.
3. See dialog.

**Expected behavior**
I expected a dialog that would ask for my Audible/Amazon password.

**Screenshots**
![](https://i.imgur.com/Hx0uGC4.png)

**Platform**

Arch Linux

```
% uname -a
Linux m 6.4.7-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 27 Jul 2023 22:02:18 +0000 x86_64 GNU/Linux
% kwin_x11 -v
kwin 5.27.6
```

**Log Files**

```
2023-08-02 00:52:08.180 -05:00 [INF] (at AppScaffolding.LibationScaffolding.logStartupState(LibationFileManager.Configuration)) 

 {}2023-08-02 00:52:08.200 -05:00 [ERR] (at LibationFileManager.InteropFactory.get_InteropFunctionsType()) Unable to locate *ConfigApp.dll
 {}2023-08-02 00:52:08.208 -05:00 [INF] (at AppScaffolding.LibationScaffolding.logStartupState(LibationFileManager.Configuration)) Begin. {"AppName":"Libation","Version":"10.6.3.1","ReleaseIdentifier":"LinuxAvalonia","OS":"Linux","InteropFunctionsType":null,"Mode":"Release","LogLevel_Verbose_Enabled":false,"LogLevel_Debug_Enabled":false,"LogLevel_Information_Enabled":true,"LogLevel_Warning_Enabled":true,"LogLevel_Error_Enabled":true,"LogLevel_Fatal_Enabled":true,"BetaOptIn":false,"UseCoverAsFolderIcon":false,"LibationFiles":"/home/matt/Libation","BooksDirectory":"/home/matt/Libation/Books","InProgress":"/tmp/Libation","DownloadsInProgressDirectory":"/tmp/Libation/DownloadsInProgress","DownloadsInProgressFiles":0,"DecryptInProgressDirectory":"/tmp/Libation/DecryptInProgress","DecryptInProgressFiles":0}
[...]

PR #696, opened 2023-08-03
Bug fixes

Mbucari:
## Fix walkthrough causing freeze (#695)
Weird bug caused by avalonia changing their dispatcher implementation.

## Fix account nickname retrieval (#629)

He hasn't answered my question, but I feel certain that the only way he would be getting the same nickname for two different accounts is if both accounts had the same account ID but different locales.

### DIFF
--- a/Source/FileLiberator/UtilityExtensions.cs
+++ b/Source/FileLiberator/UtilityExtensions.cs
@@ -30,7 +30,7 @@ namespace FileLiberator
 			using var persister = AudibleApiStorage.GetAccountsSettingsPersister();
 			var nickname
 				= persister.AccountsSettings.Accounts
-				.FirstOrDefault(a => a.AccountId == libraryBook.Account)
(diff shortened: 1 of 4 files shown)


Source/FileLiberator/UtilityExtensions.cs
```diff
@@ -30,7 +30,7 @@ public static LibraryBookDto ToDto(this LibraryBook libraryBook)
 			using var persister = AudibleApiStorage.GetAccountsSettingsPersister();
 			var nickname
 				= persister.AccountsSettings.Accounts
-				.FirstOrDefault(a => a.AccountId == libraryBook.Account)
+				.FirstOrDefault(a => a.AccountId == libraryBook.Account && a.Locale.Name == libraryBook.Book.Locale)
```
rmcrackan, line 33: Damn, this bug has been hanging out here for yeeeeeears. Good find.

Issue #698, opened 2023-08-04
Cannot edit custom naming for files


BearBryant:
**Describe the bug**

**To Reproduce**

1. Go to settings
2.  Download/Decrypt tab
3. Click edit button for files or for chapters (editing folder names works)
4. Get error

**Expected behavior**
Be able to edit template for file names

**Screenshots**

**Platform**
Windows 10

**Log Files**
Attach your Libation log file here.
[_Log202308.zip](https://github.com/rmcrackan/Libation/files/12260838/_Log202308.zip)

Object reference not set to an instance of an object.

   at LibationWinForms.Dialogs.EditTemplateDialog.<templateTb_TextChanged>g__slashWrap|11_0(String val, <>c__DisplayClass11_0&) in D:\a\Libation\Libation\Source\LibationWinForms\Dialogs\EditTemplateDialog.cs:line 66
   at LibationWinForms.Dialogs.EditTemplateDialog.templateTb_TextChanged(Object sender, EventArgs e) in D:\a\Libation\Libation\Source\LibationWinForms\Dialogs\EditTemplateDialog.cs:line 99
   at System.Windows.Forms.TextBoxBase.OnTextChanged(EventArgs e)
   at System.Windows.Forms.Control.set_Text(String value)
   at System.Windows.Forms.TextBoxBase.set_Text(String value)
   at System.Windows.Forms.TextBox.set_Text(String value)
   at LibationWinForms.Dialogs.EditTemplateDialog.EditTemplateDialog_Load(Object sender, EventArgs e) in D:\a\Libation\Libation\Source\LibationWinForms\Dialogs\EditTemplateDialog.cs:line 43
   at System.Windows.Forms.Form.OnLoad(EventArgs e)
   at System.Windows.Forms.Control.CreateControl(Boolean fIgnoreVisible)
   at System.Windows.Forms.Control.CreateControl()
   at System.Windows.Forms.Control.WmShowWindow(Message& m)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ScrollableControl.WndProc(Message& m)
   at System.Windows.Forms.Form.WndProc(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, WM msg, IntPtr wparam, IntPtr lparam)

rmcrackan:
[...]

PR #699, opened 2023-08-04
Fix broken template editor (#698)

Mbucari:


rmcrackan:
> This one is probably pretty urgent since templates are a main feature

On it. Thanks for the heads up and the quick fix.

PR #703, opened 2023-08-06
Fix #701: Replace filter with basic stroke in `./Source/LoadByOS/Linu…

### LINKED ISSUE #701: Icon not visible when pinned to dark taskbar in KDE Plasma
**Describe the bug**
I have Libation pinned to my KDE Plasma taskbar, but because I am using a dark theme, the icon is invisible.

I don't have enough experience with Plasma to know if it provides developers with a solution to this issue. A kludgy solution might be to include "Use light-colored icon in taskbar" in Settings or something, but that's pretty grim.

If this bug ends up WONTFIX, I could easily probably make a light-colored version of the icon SVG and write a custom .desktop file. But what about the next guy? :-)

**To Reproduce**

1. Install Linux versions of KDE Plasma and Libation.
2. Open Plasma.
3. Open "System Settings > Appearance > Global Theme" and choose "Breeze Dark".
4. Open Libation.
5. Right click on the taskbar icon and choose "Pin to taskbar".
6. Exit Libation (to view the default state of the icon)
7. Notice the dark icon on the dark background.

**Expected behavior**

To be able to look at the pinned icon and know it would open Libation if clicked.

**Screenshots**

Screenshot

**Platform**

```
% grep -v _URL /etc/os-release
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
LOGO=archlinux-logo
% uname -a
Linux m 6.4.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 03 Aug 2023 16:02:01 +0000 x86_64 GNU/Linux
% plasmashell --version
plasmashell 5.27.7
% kwin_x11 --version
kwin 5.27.7
% libationcli version
Libation Chardonnay v10.6.5
```

**Log Files**
n/a

mattalexx:
…xConfigApp/libation_glass.svg` to support dark mode in linux installs

Well, the research I did was interesting, but the whole thing seemed too complex when I stepped back and realized that removing that filter and using some trusty old CSS instead would probably fix Plasma. And all the other desktops too, since the fix is not Plasma-centric.

So, I dumbed down that file and it worked. And personally, I think the solid stroke line looks pretty spiffy:

Preview of light mode vs dark mode

Notice the `nnn` icon below it (the rook). That was the inspiration.

### DIFF
--- a/Source/LoadByOS/LinuxConfigApp/libation_glass.svg
+++ b/Source/LoadByOS/LinuxConfigApp/libation_glass.svg
@@ -1,28 +1,17 @@
-<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="256.0px" height="256.0px" viewBox="0 0 512 512" enable-background="new 0 0 512 512">
-<path id="glass" d=
-"M139,2
-	A 192,200 0 0 0 103,84
-	A 222,334 41 0 0 241,320
-	V478
-	H160
-	A 16,16 0 0 0 160,510
-	H352
-	A16 16 0 0 0 352,478
-	H271
-	V320
-	A 222,334 -41 0 0 409,84
-	A 192,200 0 0 0 373,2
-M355,32
-	A 192,200 0 0 1 381,127
-	A 187.5,334 -35 0 1 256,286
-	A 187.5,334 35 0 1 131,127
-	A 192,200 0 0 1 157,32
-	H355
-z" />
-
-<path id="wine-level" d=
-"M146,128
-	A 168,300 35 0 0 256,270
-	A 168,300 -35 0 0 366,128
-z"/>
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   version="1.1"
+   viewBox="0 0 524 524"
+   enable-background="new 0 0 524 524"
+   id="svg2"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <path
+     id="glass"
+     style="fill:#000000;stroke:#ffffff;stroke-width:15;stroke-opacity:1"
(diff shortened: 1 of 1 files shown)


Mbucari:
@mattalexx Can you please try this version and tell me what you think?

https://github.com/Mbucari/Libation/blob/master/Images/libation_glass.svg

rmcrackan:
> Thoughts?

On using your svg with the glow? Looks good to me.

mattalexx:
> @mattalexx Can you please try this version and tell me what you think?
> 
> https://github.com/Mbucari/Libation/blob/master/Images/libation_glass.svg

rmcrackan:
Sorry! I forgot this PR was still going on. Are y'all still discussing it? Or is it ready for merge?

Mbucari:
@rmcrackan I think we will go with his stroke idea, but this one shouldn't be merged. I need to fix the top of the wine glass so I'll do it myself.

@mattalexx Thanks for the help testing!

mattalexx:
I had suggested using PNG, did you see that above? I think you'd be able to control this better if you used raster graphics instead of vector.

> @mattalexx Thanks for the help testing!

@Mbucari Thank you for letting me contribute!

Mbucari:
> I had suggested using PNG, did you see that above? I think you'd be able to control this better if you used raster graphics instead of vector.

I did, and I only want to do that is the last resort. Hopefully these changes will be good and everybody will be happy. If not then we'll go PNG, but I have a lot less confidence with those. How many of the resolutions should we support? If we don't support one of the resolutions, will that cause some distros to show a blank icon at some resolutions?

mattalexx:
Thing is, a vector icon's stroke may look great at large size or at 48x48, but when it's being used for a smaller icon like 24x24, that stroke looks like a hairline. If you provide a 24x24 too, you can make sure the stroke is exactly how you want it to look at those smaller sizes (24x24 and smaller).

> If we don't support one of the resolutions, will that cause some distros to show a blank icon at some resolutions?

No, to the best of my knowledge, they round up to the closest larger size, then scale down if needed.

If you provide an SVG file too, they'll use that if they need anything larger than your largest PNG. 

> How many of the resolutions should we support?

I would probably suggest:

scalable SVG for 49x49 and larger
48x48 PNG for 25x25 to 48x48
24x24 PNG for 24x24 and smaller

Issue #715, opened 2023-08-17
Failed Import


mlaforge64:
[Log202308.zip](https://github.com/rmcrackan/Libation/files/12369035/Log202308.zip)
Following the instructions after a failed library import.
I tried several times on two computers and receive the same error

Mbucari:
@mlaforge64 Thanks for the report. This will be fixed shortly.

Out of curiosity, are you subscribe to [the THE JORDAN
HARBINGER SHOW](https://www.audible.com/pd/The-Jordan-Harbinger-Show-Podcast/B08K57HXZS)? I ask because I got the same error today and it was caused by the most recent episode of that podcast.

rmcrackan:
Oh god don't tell me about podcast like this one. In my transition from podcast addict to audiobook addict I tried to keep up with several very good long form interview podcasts and also have time for audiobooks. It was ugly.

Mbucari:
The secret is to live very far away from your office and have an atrocious commute.

rmcrackan:
[Fixed in version 11.0.1](https://github.com/rmcrackan/Libation/releases/latest)

mlaforge64:
I don't subscribe to that one, But I agree with you about the commute. I live an hour and an half away from my shop and can't imagine life without audiobooks and Podcasts

mlaforge64:
Outstanding! Thank you for your quick response and repair!

PR #716, opened 2023-08-17
Fix NRE when item has no category ladders (#715)

Mbucari:
This should be released asap because it's preventing the user from importing his library.

### DIFF
--- a/Source/DtoImporterService/CategoryImporter.cs
+++ b/Source/DtoImporterService/CategoryImporter.cs
@@ -23,9 +23,11 @@ namespace DtoImporterService
 			loadLocal_categories();
 
 			// upsert
+			//Import item may not have no (null) categories
 			var categoryLadders = importItems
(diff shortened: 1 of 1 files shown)


rmcrackan:
On it

Mbucari:
All-in-all a pretty good record for v11.0.0! It survived nearly a week and is only being replaced because of a bug from a much earlier release.

CLHatch:
> All-in-all a pretty good record for v11.0.0! It survived nearly a week and is only being replaced because of a bug from a much earlier release.

Alas, v11.0.0... We knew him well.

rmcrackan:
Agreed. I'm pleased. New version building now

rmcrackan:
In case you're wondering, github only gives me a few concurrent build actions at a time, so I often cancel some of the automatic runs to rush the version built to the top of the queue.

PR #751, opened 2023-09-18
Minor Bug Fixes

Mbucari:
Two easy issues fixed.

As I said in another thread, I'm a bit burned out now and I've got lots of other life stuff happening now through October 1.


rmcrackan:
Thanks for this fix

> I'm a bit burned out now and I've got lots of other life stuff happening now through October 1.

I assumed such might be the case. I started to reach out to you but my wording attempts always sounded like passive aggressive requests, which wasn't the intent so I deleted that email draft.

There's never not going to be more requests. Pause or stop anytime you feel like it. Maybe it'll sound like fun again later; maybe not.

Issue #977, opened 2024-09-16
Folder cover art setting not applying on Docker container


muchtall:
**Describe the bug**
When copying the Settings.json from a Windows installation, and then subsequently running with these settings in Docker, the "Set cover are as the folder's icon" setting does not apply, and appears to be ignored here: https://github.com/rmcrackan/Libation/blob/d1d0a7e4871b4e581a21a81755778d30951e0441/Source/LibationFileManager/WindowsDirectory.cs#L18

Since this is a headless installation which requires some advanced knowledge of the settings involved, I would expect this setting to apply as intended from the JSON. Even though the app is running in Linux via Docker, it's still desired that I be able to populate the directory with the ico and ini files so that I get the custom thumbnail when viewing the data directory in Windows.

I would suggest that the config line above be modified to state this instead:
`if (!Configuration.Instance.UseCoverAsFolderIcon)`

**To Reproduce**

1. Install Libation in Windows
2. Enable "Set cover art as the folder's icon"
3. Copy Settings.json to the /config volume in the Docker container
4. Start the Docker container and allow it to download a title
5. The desktop.ini and Icon.ico files are missing

**Expected behavior**
The UseCoverAsFolderIcon Settings.json setting should be honored, even if the application isn't running in Windows

**Platform**
Linux/Docker

rmcrackan:
Your PR is now available in [pre-release v11.4.1](https://github.com/rmcrackan/Libation/releases/tag/v11.4.1)

PR #978, opened 2024-09-16
Allow Non-Windows installations to populate Windows folder artwork

muchtall:
Proposed fix to https://github.com/rmcrackan/Libation/issues/977

### DIFF
--- a/Source/LibationFileManager/WindowsDirectory.cs
+++ b/Source/LibationFileManager/WindowsDirectory.cs
@@ -15,7 +15,7 @@ namespace LibationFileManager
             try
             {
                 //Currently only works for Windows and macOS
-                if (!Configuration.Instance.UseCoverAsFolderIcon || Configuration.IsLinux)
+                if (!Configuration.Instance.UseCoverAsFolderIcon)
                     return;
 
                 // get path of cover art in Images dir. Download first if not exists


rmcrackan:
Thanks for the PR! This removes the check but it doesn't fix the underlying issue: the developer who put this here said it doesn't work for linux and the check is to avoid errors. After removing the check, does it actually work for linux?

wtanksleyjr:
I would assume he means that it doesn't do the work to set a folder image
for Linux, but that's not what the requester wants -- he wants a Windows
image set for the folder.

Technically, we should probably allow the user to set the kind of image
that's going to be set - currently supporting Mac folder images and Windows
folder images (or, I wonder, both?). There's no reason why the kind of
system Libation runs on should constrain the kind of folder image it
creates. (As opposed to illegal characters in filenames - the system one
runs on will cause a crash or no file creation if you ever write an invalid
character to a filename. Actually it would be nice to have a simple control
for that as well, so I can always know my Linux Libation won't write an
invalid filename for Linux, but I can also set a simple checkbox and also
be 100% sure it won't write an invalid NTFS filename, or an invalid MacOS
filename, or both.)

-Wm

rmcrackan:
Am I correct that this is the case being addressed: Libation is running on a Linux system and saving to a Windows system. The user would like the saved folder and files to be subject only to Windows limitations.

If this is not the case, please try again since I misunderstood.

If this is the case, I need to reject this PR. Taking off the safety rails to support that mismatch will hard more users than it helps.

You should be able to cobble together a script by using the Libation setting to export the image (settings > audio file options > save cover image alongside audiobook). Here's an example for code setting the dir icon:

* C# https://stackoverflow.com/q/41392818
* batch https://stackoverflow.com/q/23909922

wtanksleyjr:
That's not the case. Sorry to be unclear.

THIS PR is only about Windows folder icons. Those have no safety-rail
implications. Turning them on when writing on Linux will do nothing bad to
Linux. It'll create some additional files that Linux will ignore.

I was generalizing the request to allow turning ON the safety rails for
other systems, not turning off any safety rails ever. But ignore me; that's
actually a much bigger PR.

-Wm

rmcrackan:
If I'm understanding correctly:

If I allow this check to be removed, then the check will not be hit by Libation on linux systems and thus Libation on linux will attempt to set the directory image. This is the PR author's desire, per this from their ticket

> Even though the app is running in Linux via Docker, it's still desired that I be able to populate the directory with the ico and ini files so that I get the custom thumbnail when viewing the data directory in Windows.

Removing the check will allow them to do this. The other side effect of removing this check is that for typical linux users (who are running libation on linux and who are saving to linux) that this check will no longer be hit. This typical scenario is why the check exists.

wtanksleyjr:
Correct. However, it's not clear why Linux users would want this check;
they can get the same effect by turning off the folder images, which I
believe is the default anyhow. Since we don't claim support for Linux
folder images, it's not clear why they would expect folder images to work
anyhow. (And if they do turn it on, nothing bad happens; the only result is
that a small amount of additional space is taken up.)

-Wm

rmcrackan:
Your PR is now available in [pre-release v11.4.1](https://github.com/rmcrackan/Libation/releases/tag/v11.4.1)

muchtall:
Sorry I missed the comment party. @wtanksleyjr understood the use-case correctly. I'm a Windows desktop user, but I use Docker on Linux wherever I can. Having the Docker instance populate the Windows-specific folder icon files is still helpful, thus the PR.

Thanks for merging!

Issue #993, opened 2024-10-10
Disable annoying shrink/expanding scrollbars


cbordeman:
I've fixed this, submitting a small PR to do it.

Scrollbars everywhere (particulary on the main product list) are so freaking tiny, it's really hard to locate them and click and start to scroll.  When you mouse over scrollbars, they grow to a normal size, but I think it would be so much better if they were a normal size all the time.

This does _not_ refer to autohide, where scrollbars are not present when the contents are small.

cbordeman:
The PR is #994

cbordeman:
Nice!  Although I'm technically just over the border in Fort Mill, SC now.

rmcrackan:
Fair enough. I have family in Rock Hill and we still say they're in Charlotte.

PR #994, opened 2024-10-10
Disable shrink/expand on mouseover for all scrollbars.

cbordeman:
Implements feature request #993


rmcrackan:
Thanks so much! I also see other notifications from you which I'll look at soon.

Issue #995, opened 2024-10-10
Enable friendly naming quick filters (I will PR)


cbordeman:
Would like to have a Name column on the edit filters windows.  If present, the Name would display on the menu.

I'll PR this.

rmcrackan:
Hey, what's the matter? My hack ain't good enough for you?

![Screenshot 2024-10-10 07 34 05](https://github.com/user-attachments/assets/144112bd-c0df-4958-8c01-d9f6c0ceeddb)

:smile: Thanks for the new feature.

A word of warning: I keep classic (winforms) and chardonnay (avalonia/xaml) as close to feature parity as possible. Any UI changes will need to be made to both. (Obviously unless it's framework-specific like the PR you did for the xaml scroll bars.)

cbordeman:
I actually didn't notice you could do the hack, but I think making it formal has some valye.  

I'm going to go ahead and have a go at a separate field, though my WinForms experience is very limited.  IMO the chardonnay (Avalonia) is so much better, is there a reason it is still around?  Pardon my ignorance.

cbordeman:
Also, I realize that while loading the filters, I'll need to handle the missing name field gracefully so as to not nuke existing data.

rmcrackan:
> is there a reason it is still around?

Fair question and one I've considered. Despite the legion reasons not use winforms:

* Years before Libation was *Libation*, this was just an app I made for myself in winforms. (Winforms was the only version until another dev'er did the full Avalonia port a few years ago.) It's simply what I knew better so I went with it.
* Every time I try to do more than trivial things in xaml I end up tearing out my hair. Of which I have little to spare.
* It's absolutely rock solid and supported officially by MS. Every so often bugs come up in Avalonia. I don't know if I've ever found a proper bug in winforms since I started using it over 20 years ago.
* It's time-tested and filled with mature things. It's had out of the box screen reader support since forever. I don't have accessibility needs but I have close family who do so I'm especially sensitive to such things.

CharlieRussel:
As a devoted user of Libation in its Classic (winform) implementation, I'd be really unhappy if it went away. It has the one thing going for it that beats everything, it JFW.

rmcrackan:
> JFW

lol. No worries friend -- classic isn't going anywhere. Classic and Chardonnay are here to stay. One day I'll remember to write up an explanation in the faq. One day...

cbordeman:
@CharlieRussel What doesn't just work in the chardonnay?

[...]

PR #997, opened 2024-10-11
Implemented Name on Quick Filters.

cbordeman:
#995

### DIFF
--- a/Source/AppScaffolding/AppScaffolding.csproj
+++ b/Source/AppScaffolding/AppScaffolding.csproj
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net8.0</TargetFramework>
-    <Version>11.4.1.1</Version>
+    <Version>11.5</Version>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Octokit" Version="11.0.1" />
--- a/Source/AppScaffolding/LibationScaffolding.cs
+++ b/Source/AppScaffolding/LibationScaffolding.cs
@@ -7,6 +7,7 @@ using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using Serilog;
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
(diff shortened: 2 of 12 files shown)


rmcrackan:
I'm back in town and slowly working through my mountain of emails.

This PR looks good. I'm a little hesitant about the migration code in the QuickFilters static ctor. It's just that if anything goes wrong in a static ctor, we can get in a weird state. Historically I've handled migrations in a central place. Eg:

https://github.com/rmcrackan/Libation/blob/f7a4a95e3b2e5c0eec0b1c161da96fef567d93d6/Source/AppScaffolding/LibationScaffolding.cs#L89

If it's not too much trouble, could you do this with your migration? The next version can be 11.4.2 or 11.5 -- your choice.

RE version: a lot of users had issues last year when I updated .net. Since then, I rolled back and have been slowly adding to the code base and I *think* I fixed the issue they were having. So fairly soon, we can create an actual new release version. I see I have some other stuff from you and I'll take a look at that before creating a new release. Before releasing, I'd also like to see where that conversation goes with adding a player.

cbordeman:
Sure, I usually use a more formal upgrade process, like you do.  I will migrate to that.

cbordeman:
I'll use 11.5, seems like a good time.

cbordeman:
Moved the migration.  If I understood the migration strategy a little better, I might have written the code differently, though.  It seems like if it fails, the app fails a migration method; is that what we want to happen?  Currently, if it fails, I'm creating a new, empty QuickFilters.cs...

Source/AppScaffolding/AppScaffolding.csproj
```diff
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net8.0</TargetFramework>
-    <Version>11.4.1.1</Version>
+    <Version>11.5</Version>
```
rmcrackan, line 5: For stupid legacy reasons that I swear I'll fix one day (don't ask how long I've been telling myself this), please always use 4 numbers. Eg: 11.5.0.1. Don't worry if you don't get around to this; I'll fix it if you don't.

rmcrackan:
Included in latest pre-release -- https://github.com/rmcrackan/Libation/releases/tag/v11.5.0

Let me know when you have a better idea about the player. That would be one with a big wow-factor and I'd wait for it to be finalized before making a new public release.

cbordeman:
Occurs to me I didn't update the version # on the Winforms side.  Can you do that?

rmcrackan:
What version on the winforms side?

cbordeman:
Is there not a version in there somewhere?  Like in the Avalonia project file or a Properties.cs?  You're the expert.

rmcrackan:
The only version should come from that shared spot in appscaffolding

https://github.com/rmcrackan/Libation/blob/6dcf456d06af28ba8a3bbb7874a2402832f292d5/Source/AppScaffolding/AppScaffolding.csproj#L5

Issue #998, opened 2024-10-12
Add "System" option for theme and default to it on new installs (I will PR)


cbordeman:
Like many diabetics, I am photosensitive.  White backgrounds not super easy to look at!

There IS a "Dark" option, but it would be much easier for it to just default to the system theme automatically.

Also, I noticed the WinForms version doesn't have the option to select the theme, or I just can't find it.  This change benefits WinForms, too, for new installs by at least using the system theme automatically.

cbordeman:
PR #999

rmcrackan:
Thank you! Wow -- I had no idea this was a diabetic trait. My wife is super photosensitive and her sister is juvenile onset diabetic and somehow this has never come up. Makes me wonder if there's a connection.

cbordeman:
@rmcrackan Diabetes causes severe nerve degredation, to the extremities and especially to the optic nerve.  Does she have a dry, sore throat, too?  Same cause.  Only solution is to control blood sugar religiously and hope that keeps it from getting worse.

rmcrackan:
Something we only discovered about a year ago which may help you: my photosensitive wife has always gotten migraines from screens. Little by little we've found a list of good screens: Kindles, CRTs, she could could use her phone much longer back when we upgraded from iPhone 8 => 13, and she can watch our friend's TV. We finally worked it out: her bad screens are backlit led laptops/monitors and/or have low refresh. We just got her a pricy OLED laptop with a 120 refresh and she's been able to use her laptop for the first time headache-free in ages.

Our friend's TV was a top-end OLED with high refresh, ver 8 was the last iPhone without OLED, kindles aren't backlit at all, and CRTs are such a different technology that we have no good guess and it obviously doesn't come up much.

As with *everything* migraine-related: ymmv. Every migraine sufferer is basically a unique case. (If you want to anger one, tell them that you solved yours and all they have to do to fix theirs is listen to you.)

cbordeman:
Yeah, I don't get migraines often, but a huge OLED with high reresh rates helps.

My suggestion for her: use lots and lots of eye drops any time she can, the slightest pain or irritation.  But only one brand works well: Bion Tears $17-18$ on Amazon.  All other brands will make it worse over time.  Tons and tons of people like me will swear the same thing, my opthamologists tell me.  I swear they are a god send.

Tell her to put in a few drops every morning, as soon as she wakes up, as a preventative measure.

PR #999, opened 2024-10-12
Implemented "System" option for theme.

cbordeman:
I couldn't find the option to change this in Winforms, however the default for new installs is now "System" so it will still benefit from the change.  

Issue #998


rmcrackan:
Thanks! I'll include it in the same release as #997 when that's finalized.

PR #1026, opened 2024-10-29
Player preliminary work

cbordeman:
@rmcrackan I've done just a little work on the playlist.  I renamed the right side popup window from ProcessQueue to Sidebar and added a tab for "Playlist."  

I did my vm implementation mainly in LibationUiBase, and did some work to allow the WinForms side to use viewmodels, namely a ViewModelBase.  I added a subclass of DataGridView with a bindable SelectedItem property (why doesn't the base control doesn't have this??) called DataGridViewEx.  This should allow a lot more code to be in common with WF/Av.

I also added some Container stuff and an EventAggregator to make it possible to get global stuff (like the PlayerViewModel and the EventAggregator) without using static classes, and in a common way between Winforms and Avalonia.  Types are registered in two different RegisterTypes() methods called at startup, one in WinForms and one in Avalonia.


rmcrackan:
Is this still on-going? Or abandoned?

rmcrackan:
@cbordeman Is it fair to say that this isn't going to happen? Or if it is: it'll be merge hell

cbordeman:
I'm sorry, I've been just too swamped.  I have to back away from this.

rmcrackan:
No worries. It happens. I'm going to close this PR and you're welcome back anytime it's convenient.

cbordeman:
Thanks.

Issue #810, opened 2023-12-18
Flatpak - Linux


cadric:
**Is your feature request related to a problem? Please describe.**
Flatpak's are a much better universal, sandboxed way of distributing apps on linux. It works on a majory of linux distro's. And it is much easier to handle and update for a user. And for some immutable distro's, like Fedora Silverblue, its the only way to install apps.
You can see distro's where flatpaks can be setup for here: [https://flathub.org/setup](https://flathub.org/setup)

**Describe the solution you'd like**
While I am happy that you now create RPM and Deb files for libation, it would be better if you did it as a flatpak. They update automatically and it is possible for the user to download the app at either [flathub.org](https://flathub.org) or in their software center.
There is plenty of information on how to build flatpaks at [https://docs.flatpak.org/en/latest/first-build.html](https://docs.flatpak.org/en/latest/first-build.html) and I am sure if you reached out to [cassidyjames](https://discourse.flathub.org/u/cassidyjames/summary), he would be more then happy to help you get setup with building a flatpak. He has helped a lot of apps to get on flathub.

**Additional context**
A Verified version of Libation on flathub would be so hot ;)

P5421135-2:
Wholeheartedly agree, I am on void linux and I cannot install it. A flatpak would bring in many more users.

RyanOrigens:
Also unable to install on Arch Linux due to the AUR package being broken, and no other alternative for downloading, flatpak would be a game changing!

rmcrackan:
I don't know how to script github to create a "flatpak". If you do though, let me know

jjcf89:
If anyone cares, I did find a snap of the project. https://snapcraft.io/libation

RyanOrigens:
It doesn't work out! The snap build is also broken!

jwillikers:
@rmcrackan I think most users will want the app to be available on Flathub, which will take care of building everything for you. You'll need to submit a PR to add your app to the [flathub](https://github.com/flathub/flathub/) repository.

There's a [Flatpak GitHub Action](https://github.com/flatpak/flatpak-github-actions) available, which is really easy to use if you just want to build your application. It's a bit more involved to set up your own Flatpak repository for users to actually obtain the Flatpak, but it is possible through GitHub Pages.

[...]

PR #1036, opened 2024-11-10
Add some necessary files for the Flatpak

jwillikers:
Related to #810.


rmcrackan:
Sorry for the delay. Is this PR ready for my approval? After this is done, what else do I need to do so the linux folks can benefit from it? This is all greek to me

jwillikers:
> Sorry for the delay. Is this PR ready for my approval? After this is done, what else do I need to do so the linux folks can benefit from it? This is all greek to me

@rmcrackan I think it should be good. The only thing I haven't tested is the GitHub Actions, one of which will fail until this is merged. I set up the workflows to only run when the relevant files change, so even if I need to work out a kink, it shouldn't effect anyone else.

rmcrackan:
Cool. Here goes nothing.

jwillikers:
@rmcrackan I'll fix that

rmcrackan:
Gracias

rmcrackan:
included in https://github.com/rmcrackan/Libation/releases/tag/v11.5.4

PR #1039, opened 2024-11-13
Fix metainfo validation

jwillikers:


jwillikers:
@rmcrackan That fixed it.

PR #1042, opened 2024-11-14
Exclude docker build report artifact from release packing

pixil98:
When I redid the docker build process, it added a new artifact to the workflow that contains a report on the docker image that was built. This was causing an error when the release process was trying to download all of the artifacts for the release. This PR adds a pattern to include only the artifacts that should get included in the release.

You can see a successful run of the fixed workflow at https://github.com/pixil98/Libation/actions/runs/11846675142

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,6 +43,7 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           path: artifacts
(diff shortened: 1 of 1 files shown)


rmcrackan:
Perfect. Much thanks for the quick response!

rmcrackan:
You latest changes are included in https://github.com/rmcrackan/Libation/releases/tag/v11.5.4

PR #1050, opened 2024-11-20
set maxdepth to prevent traversal into subdirectories

stickystyle:
This restricts the `find` command to the `DBPATH` and prevents it from traversing into subdirectories when checking for multiple *.db files. I just encountered this issue because I copy my DB and settings files to a `bak` directory before making any changes and found that in the most recent pull of the container, it wouldn't start up anymore.

```
.
├── AccountsSettings.json
├── bak
│   ├── AccountsSettings.json
│   ├── Settings.json
│   └── LibationContext.db
├── FileLocations.json
├── LibationContext.db
├── SearchEngine
│   ├── _7y6.cfs
│   ├── segments_7w5
│   └── segments.gen
└── Settings.json
```

### DIFF
--- a/Docker/liberate.sh
+++ b/Docker/liberate.sh
@@ -80,7 +80,7 @@ setup_db() {
   # Figure out the right databse file
   if [[ -z "${LIBATION_DB_FILE}" ]];
   then
-    dbCount=$(find "${DBPATH}" -type f -name "${dbpattern}" | wc -l)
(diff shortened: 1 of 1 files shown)


rmcrackan:
Great, thanks! Does this merit a new build?

stickystyle:
My gut says it's a pretty edge case, if this were my project I'd wait to roll it into something larger unless issues start to get reported.

Issue #1158, opened 2025-02-26
Libationcli crashes during liberation


Flameborn:
Hello,

When I  try to liberate any book via the latest stable libationcli (12.0), I get a crash with the abort signal. The last entry in the log in *Verbose* mode is:

```
 {}2025-02-26 13:45:06.626 +01:00 [INF] (at FileManager.FileUtility+<>c__DisplayClass25_0.<SaferMove>b__0()) File successfully moved: {"source":"/Volumes/hd3/Incoming/ltmp/DecryptInProgress/book.m4b","destination":"/Volumes/hd3/Incoming/book/book.m4b"}
 {}
```

On the command line, I see:

```
at System.Collections.Generic.HashSet`1[[System.__Canon, System.Private.CoreLib, Version=9.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].AddIfNotPresent(System.__Canon, Int32 ByRef)
   at System.Collections.Generic.HashSet`1[[System.__Canon, System.Private.CoreLib, Version=9.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].UnionWith(System.Collections.Generic.IEnumerable`1<System.__Canon>)
   at System.Collections.Generic.HashSet`1[[System.__Canon, System.Private.CoreLib, Version=9.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]..ctor(System.Collections.Generic.IEnumerable`1<System.__Canon>, System.Collections.Generic.IEqualityComparer`1<System.__Canon>)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.SnapshotFactoryFactory.SnapshotCollection(System.Collections.Generic.IEnumerable`1<System.Object>)
   at DynamicClass.lambda_method494(System.Runtime.CompilerServices.Closure, Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntrySubscriber.SnapshotAndSubscribe(Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.StartTracking(Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry)
[...]

PR #1160, opened 2025-02-27
Fixed stack overflow crash when movifying large databases

Mbucari:
This is mostly to address #1158

The [DbContext.Attach](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbcontext.attach?view=efcore-9.0#microsoft-entityframeworkcore-dbcontext-attach-1(-0)) recursively traverses the database looking for referenced entities that are not being tracked, and starts tracking them. The recurrsion will cause a stack overflow with large DBs. 

The solution is to use [DbContext.Entry](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbcontext.entry?view=efcore-9.0) and then manually [Reference](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.changetracking.entityentry-1.reference?view=efcore-9.0) related properties.

Also, I was looking through some of my other code, winged (as one does looking at something one wrote year(s) ago), and refactored it.

EDIT* I also edited the release.yml workflow to remove the [dwenegar/upload-release-assets](https://github.com/dwenegar/upload-release-assets) package because it's redundant with softprops/action-gh-release@v2. I could have sworn it was out of support (because I just did had to do something similar this week for one of my projects), but it now it looks like it is supported. Oh well, it's still redundant.

### DIFF
--- a/Source/ApplicationServices/LibraryCommands.cs
+++ b/Source/ApplicationServices/LibraryCommands.cs
@@ -369,12 +369,12 @@ namespace ApplicationServices
 
                 using var context = DbContexts.GetContext();
 
-                // Attach() NoTracking entities before SaveChanges()
-                foreach (var lb in removeLibraryBooks)
+				// Entry() NoTracking entities before SaveChanges()
+				foreach (var lb in removeLibraryBooks)
                 {
-                    lb.IsDeleted = true;
-                    context.Attach(lb).State = Microsoft.EntityFrameworkCore.EntityState.Modified;
-                }
+					lb.IsDeleted = true;
+					context.Entry(lb).State = Microsoft.EntityFrameworkCore.EntityState.Modified;
+				}
 
                 var qtyChanges = context.SaveChanges();
                 if (qtyChanges > 0)
@@ -398,12 +398,12 @@ namespace ApplicationServices
 
                 using var context = DbContexts.GetContext();
 
-                // Attach() NoTracking entities before SaveChanges()
-                foreach (var lb in libraryBooks)
-                {
-                    lb.IsDeleted = false;
-                    context.Attach(lb).State = Microsoft.EntityFrameworkCore.EntityState.Modified;
-                }
+				// Entry() NoTracking entities before SaveChanges()
+				foreach (var lb in libraryBooks)
+				{
+					lb.IsDeleted = false;
+					context.Entry(lb).State = Microsoft.EntityFrameworkCore.EntityState.Modified;
+				}
 
                 var qtyChanges = context.SaveChanges();
                 if (qtyChanges > 0)
@@ -518,17 +518,18 @@ namespace ApplicationServices
                 if (libraryBooks is null || !libraryBooks.Any())
                     return 0;
 
-                foreach (var book in libraryBooks)
-                    action?.Invoke(book.Book.UserDefinedItem);
-
                 using var context = DbContexts.GetContext();
 
-                // Attach() NoTracking entities before SaveChanges()
-                foreach (var book in libraryBooks)
+				// Entry() instead of Attach() due to possible stack overflow with large tables
+				foreach (var book in libraryBooks)
                 {
-                    context.Attach(book.Book.UserDefinedItem).State = Microsoft.EntityFrameworkCore.EntityState.Modified;
-                    context.Attach(book.Book.UserDefinedItem.Rating).State = Microsoft.EntityFrameworkCore.EntityState.Modified;
-                }
+					action?.Invoke(book.Book.UserDefinedItem);
+
+					var udiEntity = context.Entry(book.Book.UserDefinedItem);
+
+					udiEntity.State = Microsoft.EntityFrameworkCore.EntityState.Modified;
+					udiEntity.Reference(udi => udi.Rating).TargetEntry.State = Microsoft.EntityFrameworkCore.EntityState.Modified;
+				}
 
                 var qtyChanges = context.SaveChanges();
                 if (qtyChanges > 0)
--- a/Source/ApplicationServices/LibraryCommands.cs
+++ b/Source/ApplicationServices/LibraryCommands.cs
@@ -222,7 +222,7 @@ namespace ApplicationServices
             {
                 int qtyChanged = await Task.Run(() => SaveContext(context));
                 if (qtyChanged > 0)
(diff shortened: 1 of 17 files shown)


Source/DataLayer/QueryObjects/LibraryBookQueries.cs
```diff
@@ -21,8 +21,8 @@ public static List<LibraryBook> GetLibrary_Flat_NoTracking(this LibationContext
                 .AsNoTrackingWithIdentityResolution()
                 .GetLibrary()
                 .AsEnumerable()
-                .Where(lb => !lb.Book.IsEpisodeParent() || includeParents)
-                .ToList();
+				.Where(c => !c.Book.IsEpisodeParent() || includeParents)
+				.ToList();
```
Mbucari, line 26: I had changed something here during a test then reverted it.

Mbucari:
While I was dusting off Avalonia and looking back at the UI, I noticed I couldn't preview the Main window because `QuickFilters` were causing exceptions (due to Avalonia loading the preview window without running Migration code. I fixed this by making `QuickFilters.InMemoryState` nullable and handling all null checks.

Mbucari:
I made a change to Avalonia so that it loads the main window before the library grid is loaded, which is how classic works.

I'm not going to add anything else right now, so you can start your review.

Mbucari:
Hold on, I messed up library counts.

Mbucari:
@rmcrackan Are you familiar with the [split queries](https://learn.microsoft.com/en-us/ef/core/querying/single-split-queries)? 

I was working playing around with that enormous (> 10k book) database, looking for performance gains, and I came across `IQueryable.AsSplitQuery()`. I Applied it to `GetLibrary_Flat_NoTracking()` by changing it from :

https://github.com/rmcrackan/Libation/blob/211f15af251720b3d96bff49d2d1cb3a25519116/Source/DataLayer/QueryObjects/LibraryBookQueries.cs#L18-L25

to:
```c#
public static List<LibraryBook> GetLibrary_Flat_NoTracking(this LibationContext context, bool includeParents = false)
  => context
     .LibraryBooks
     .AsNoTrackingWithIdentityResolution()
     .GetLibrary()
     .AsSplitQuery()
     .AsParallel()
     .Where(c => !c.Book.IsEpisodeParent() || includeParents)
     .ToList();
```

I ran 10x benchmarks on each version in release mode, and here are the results. All values are time, in milliseconds, to create the context and call `GetLibrary_Flat_NoTrackin()`.

| |Original|Split Query|
|-|-|-|
|Minimum|7493|3552|
|Maximum|7958|4076|
|**Average**|**7680**|**3766**|

I'm not nearly as familiar with EF as you, so are there any pitfalls with split queryies that I'm missing? If it _seems_ to work, does that mean it did work and will continue to work with the current DB config?

Source/ApplicationServices/LibraryCommands.cs
```diff
@@ -445,10 +445,14 @@ public static int PermanentlyDeleteBooks(this IEnumerable<LibraryBook> libraryBo
         #endregion
 
         // call this whenever books are added or removed from library
-        private static void finalizeLibrarySizeChange() => LibrarySizeChanged?.Invoke(null, null);
+        private static void finalizeLibrarySizeChange()
+        {
+            var library = DbContexts.GetLibrary_Flat_NoTracking(includeParents: true);
```
rmcrackan: Damn fine optimization

Source/ApplicationServices/SearchEngineCommands.cs
```diff
@@ -48,6 +48,8 @@ public static void UpdateBooks(IEnumerable<LibraryBook> books)
 		}
 
 		public static void FullReIndex() => performSafeCommand(fullReIndex);
+		public static void FullReIndex(List<LibraryBook> libraryBooks)
+			=> performSafeCommand(se => fullReIndex(se, libraryBooks.WithoutParents()));
```
rmcrackan, line 52: Oooh, I've never used WithoutParents before. That looks interesting

Source/LibationWinForms/Form1.BackupCounts.cs
```diff
@@ -17,7 +17,9 @@ protected void Configure_BackupCounts()
 			beginPdfBackupsToolStripMenuItem.Format(0);
 
             LibraryCommands.LibrarySizeChanged += setBackupCounts;
-            LibraryCommands.BookUserDefinedItemCommitted += setBackupCounts;
+			//Pass null to the runner to get the whole library.
```
rmcrackan, line 20: Thank you for including comments like this. `(null, null)` is cryptic without it

Source/ApplicationServices/LibraryCommands.cs
```diff
@@ -445,9 +445,9 @@ public static int PermanentlyDeleteBooks(this IEnumerable<LibraryBook> libraryBo
         #endregion
 
         // call this whenever books are added or removed from library
-        private static void finalizeLibrarySizeChange()
+        private static void finalizeLibrarySizeChange(LibationContext context)
```
rmcrackan, line 448: I'm not *opposed* to these new params, I just can't tell what we've gained.

rmcrackan:
> I'm not nearly as familiar with EF as you

Oooooooh you might be surprised :)

I've never heard of that split query. DBAs have shamed me for doing exactly this manually for so many damn years, rather than doing all the 'correct' joins. I'm happy (read: vindicated) to see it's an actually useful feature. Quality discovery.

Source/ApplicationServices/LibraryCommands.cs
Mbucari: The context is already opened, so it saves having to reopen the context before grabbing the library. Can take upwards of a second.

rmcrackan:
I see your last change was quite recent. Do you want more time before I accept?

Mbucari:
> I see your last change was quite recent. Do you want more time before I accept?

 Nope, go ahead. I just had an insight about using The already open contacts, and I added the split query. But that's it.

Source/ApplicationServices/LibraryCommands.cs
rmcrackan: > Can take upwards of a second.

Fuuuck. Then by all means, yes: let's do this.

PR #1172, opened 2025-03-03
Thread safety and AccountSettings.json error handling

Mbucari:
[Add thread safety and comments re threading](https://github.com/rmcrackan/Libation/pull/1172/commits/b36e110b495789626de7b4059cdcb8a1b6cff179)

This is about addressing the threading issue introduced in 12.0.1. In 12.0.2 I added a bunch of UI thread invocations in a shotgun approach because I didn't understand exactly where the error was coming from. Since then, I did a deep dive and figured out exactly where the cross-threading operations were happening. I couldn't reproduce them organically (i.e. by just running Libation), but I could trigger the errors by executing a `Task.Run()` which always executes on a new thread. I was able to find exactly which parts need to be executed on the UI thread and added lots of comments so I hopefully never screw this up again.

[Handle and notify users of invalid account settings file](https://github.com/rmcrackan/Libation/pull/1172/commits/ac036f65f1283cb5c6d533fc339e8eb17e046443)

I added an error handling function which gets triggered if the account settings fail to load. If triggered, Libation will try to back y the accountsettings.json file, create a new, empty file, notify the user of the issue, and continue loading. If Libation is unable to backup the problem accountsettings.json file and create a new one, it notifies the user and throws.

[Add a border around dialogs with CanResize=true](https://github.com/rmcrackan/Libation/pull/1172/commits/981a183992bd8c5465c19ee0886ab14b7c28f7c4)

This is a UI tweak. Dialog windows which are fixed size had no border.

### DIFF
--- a/Source/AudibleUtilities/AudibleApiStorage.cs
+++ b/Source/AudibleUtilities/AudibleApiStorage.cs
@@ -5,19 +5,58 @@ using Newtonsoft.Json;
 
 namespace AudibleUtilities
 {
+	public class AccountSettingsLoadErrorEventArgs : ErrorEventArgs
+	{
+		/// <summary>
+		/// Create a new, empty <see cref="AccountsSettings"/> file if true, otherwise throw
+		/// </summary>
+		public bool Handled { get; set; }
+		/// <summary>
+		/// The file path of the AccountsSettings.json file
+		/// </summary>
+		public string SettingsFilePath { get; }
+
+		public AccountSettingsLoadErrorEventArgs(string path, Exception exception)
+			: base(exception)
+		{
+			SettingsFilePath = path;
+		}
+	}
+
 	public static class AudibleApiStorage
 	{
 		public static string AccountsSettingsFile => Path.Combine(Configuration.Instance.LibationFiles, "AccountsSettings.json");
+
+		public static event EventHandler<AccountSettingsLoadErrorEventArgs> LoadError;
 
 		public static void EnsureAccountsSettingsFileExists()
 		{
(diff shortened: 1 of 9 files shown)


rmcrackan:
Before merging your PR, I spent a lot of time poking around the persistence stuff again. I see room for improvement but nothing strikes me as particularly prone to file corruption. I put some safety measures in months ago that I thought fixed this crap. I'm disappointed that I must have missed whatever is causing it.

Mbucari:
@rmcrackan I have an idea to track this bug a little deeper.

https://github.com/rmcrackan/Dinah.Core/blob/7abbe015b8980331ea6f25613798f4e60da10377/Dinah.Core/_IO/JsonFilePersister.cs#L114-L160

Based on the logs I'm seeing, everyone who has this error has an empty json file, so in the persister, split the serialization and file writing operations. Serialize to string, then check if that string is null or empty. If it is, don't write it and instead throw an exception.

Mbucari:
I also noticed something else in that method that you might want to address.

In saveFile(), you're calling user code inside a lock, which is dangerous. If any users of OnSaving() or OnSaved() caused the file to update and change, you'll have a deadlock.

rmcrackan:
Yeah, and also the instance lock does nothing unless you use the same instance every time. Which I *think* I do but I might not. And I also remember seeing something recently about a transaction where I didn't try/final it correctly. As I said, there's for sure room for improvement but nothing stuck out as a cause for corruption. Maybe I'm looking at it wrong. Maybe the corruption part is fixed and I should be looking at the things you mentioned.

Issue #1173, opened 2025-03-04
Mac startup error and crashes in 12.0.2.1


codykonior:
**Describe the bug**
I was on some version of v11 and had to upgrade as downloads stopped working. So I tried 12.0.2, and then 12.0.2.1, and both give the same issue.

Bad filter string on startup:

![Image](https://github.com/user-attachments/assets/715909ef-6845-4e8a-832b-68b980ab78a4)

There is no bad filter string, it has not changed.

Then it seems to randomly crash after.

**Platform**

MacOS 15.3.1 (24D70) on an M2 Max.

**Log Files**
```
4/3/2025 10:24:07 pm - Libation Crash
 OS                    MacOS
 Version               12.0.2.1
 ReleaseIdentifier     MacOSAvalonia_Arm64
 InteropFunctionsType  MacOSConfigApp.MacOSInterop
 LibationFiles         /Users/.../Libation
 Books Folder          /Users/.../Libation/Books
 === EXCEPTION ===
 System.NullReferenceException: Object reference not set to an instance of an object.
   at LibationAvalonia.ViewModels.ProductsDisplayViewModel.UpdateGridAsync(List`1 dbBooks) in /Users/runner/work/Libation/Libation/Source/LibationAvalonia/ViewModels/ProductsDisplayViewModel.cs:line 148
   at LibationAvalonia.ViewModels.MainVM.LibraryCommands_LibrarySizeChanged(Object sender, List`1 fullLibrary) in /Users/runner/work/Libation/Libation/Source/LibationAvalonia/ViewModels/MainVM.cs:line 42
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
   at System.Threading.QueueUserWorkItemCallback.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
   at System.Threading.Thread.StartCallback()
```

rmcrackan:
@codykonior Can you attach your latest log file to this github issue?

@Mbucari It feels like this could possibly be related to your recent work? Hopefully the log file will be illuminating.

Mbucari:
@rmcrackan That's very diplomatic of you. "Hey buddy, you think this might just have been caused by all that recent mucking about you've been doing with this exact section of the codebase?" XD

@codykonior Yeah, please attach the full log if you would.

rmcrackan:
[...]

PR #1174, opened 2025-03-04
Null safety & UI tweak

Mbucari:
[Null safety checks.](https://github.com/rmcrackan/Libation/commit/56e25697e12521c9a38adf307064d3ee96b6d446)

I enabled NRTs in ProductsDisplayViewModel.cs and addressed every warning. NRE's shouldn't be possible anymore. That said, I don't know how #1173 was possible in the first place. But I suppose that's the appeal of NRTs: just because you can't envisage an NRE doesn't mean it's impossible.

[Only show buttons on mouse over](https://github.com/rmcrackan/Libation/commit/7ba34306722d52a6d2876d089969d15ec14a5f3d)

This is purely aesthetics. Hide the queue position and cancel buttons until the mouse is over the queued item. I tried this for winforms too, but it was leading to lots of flickering.


rmcrackan:
Muchas gracias

PR #1175, opened 2025-03-04
Additional null safety

Mbucari:
[Additional null safety](https://github.com/rmcrackan/Libation/pull/1175/commits/c3938c49a9690e75e69f0249e8539af43d8a8459)

This is the big one. I enabled NRTs in a lot of the chardonnay files and a few of the biz logic files, then addressed all compiler warnings.

[Fix tags font color in dark mode](https://github.com/rmcrackan/Libation/pull/1175/commits/7658f21d7ca990e5274caef9757370dbb9129b9b) and [Fix dark theme text color in DataGridTextColumn](https://github.com/rmcrackan/Libation/pull/1175/commits/e37abbf2760a929f81a942437b4f147bf56d3ae0)

More UI tweaks to correct text font color in dark mode.

### DIFF
--- a/Source/AppScaffolding/AppScaffolding.csproj
+++ b/Source/AppScaffolding/AppScaffolding.csproj
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net9.0</TargetFramework>
-    <Version>12.0.2.1</Version>
+    <Version>12.0.2.2</Version>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Octokit" Version="14.0.0" />
--- a/Source/ApplicationServices/LibraryCommands.cs
+++ b/Source/ApplicationServices/LibraryCommands.cs
@@ -15,12 +15,13 @@ using Newtonsoft.Json.Linq;
 using Serilog;
 using static DtoImporterService.PerfLogger;
 
+#nullable enable
 namespace ApplicationServices
 {
     public static class LibraryCommands
     {
-        public static event EventHandler<int> ScanBegin;
-        public static event EventHandler<int> ScanEnd;
+        public static event EventHandler<int>? ScanBegin;
+        public static event EventHandler<int>? ScanEnd;
 
         public static bool Scanning { get; private set; }
         private static object _lock { get; } = new();
@@ -100,7 +101,7 @@ namespace ApplicationServices
         }
 
         #region FULL LIBRARY scan and import
-        public static async Task<(int totalCount, int newCount)> ImportAccountAsync(Func<Account, Task<ApiExtended>> apiExtendedfunc, params Account[] accounts)
+        public static async Task<(int totalCount, int newCount)> ImportAccountAsync(Func<Account, Task<ApiExtended>> apiExtendedfunc, params Account[]? accounts)
         {
             logRestart();
 
@@ -236,7 +237,7 @@ namespace ApplicationServices
         {
             var tasks = new List<Task<List<ImportItem>>>();
 
-           await using LogArchiver archiver
+           await using LogArchiver? archiver
                 = Log.Logger.IsDebugEnabled()
                 ? new LogArchiver(System.IO.Path.Combine(Configuration.Instance.LibationFiles, "LibraryScans.zip"))
                 : default;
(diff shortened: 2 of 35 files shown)


rmcrackan:
Good golly miss molly, you weren't kidding about it being a big one. I ain't combing though all these null ref changes. Imma just have to trust you.

rmcrackan:
I'm not personally a fan of the nullref feature. How do you like it? I mean, ok, that's a dumb question. You obviously like it since you just did this huge PR for it. But you know what I mean: can you describe your experience?

Mbucari:
I have mixed feelings about them, but I definitely skew towards liking them.  One of the really annoying things is that the analyzer can't tell that your code is setting a field/property in the constructor if it's being set by a helper method called by the constructor. Take [the settings view models](https://github.com/rmcrackan/Libation/pull/1175/commits/c3938c49a9690e75e69f0249e8539af43d8a8459#diff-15ba1a7fb3cacdd3c25daf5451ef24188e771edc03bbc181066133e4fee09d8a) as an example. I used to call LoadSettings() from the constructor which would set all the non-nullable properties, but the analyzer can't see that and so it produces warnings. I prefer that style because it's self-documenting and cleans up the constructor, but when NRTs are enabled, I'd rather refactor the code to satisfy the analyzer than suppress the warning. In general, I prefer refactoring over suppressing the warnings or using the null-forgiving operator(!). Refactoring to satisfy the analyzers can create more boilerplate code and can add some ugliness, but I think those prices are worth the benefits.

The main benefit is obviously a lower chance of NREs (and I believe dramatically lower if done well). #1173 would never have happened if NRTs had been enabled in the view model. The reason we didn't do a null check is because we assumed that the namedFilter would be non-null here:
https://github.com/rmcrackan/Libation/blob/85b679246830386ff70d4405417854c31adafcfa/Source/LibationAvalonia/ViewModels/MainVM.Filters.cs#L59-L60

But if NRTs were enabled, we would have been warned about this line of code which could pass a null:
https://github.com/rmcrackan/Libation/blob/85b679246830386ff70d4405417854c31adafcfa/Source/LibationAvalonia/Views/MainWindow.axaml.cs#L129

Another smaller benefit of NRTs is that they can sometimes help you write _fewer_ null checks. Sometimes we are excessive in our null checks for safety, resulting in multiple redundant null checks or excessive use of the null coalescing operator. With NRTs enabled, if you check for a variable's nullity, then use the null coalescing operator on that variable, the analyzer will again mark that variable as "maybe null". So you can sometimes clean up your code by only checking for nullity once.

And finally, I think NRTs make me a more thoughtful developer. If I quickly whip up a new class, I'm forced to make good choices about which members are nullable and which aren't from the very beginning. Then as I continue developing, I'm forced to write code that either respects those choices, or re-design the code to make a different choice. But at all times I must be thoughtful and address the possibility of NREs, lest I be pestered by warnings!

rmcrackan:
Thanks for the thoughtful analysis.

> analyzer can't tell that your code is setting a field/property in the constructor if it's being set by a helper method

I feel ya. For literally decades I've wished the ctor could call a more descriptive/shared subroutine to set a readonly but not only does the analyzer not allow it, the compiler doesn't.

> null-forgiving operator

I love that this is also called the "damn it!" operator

> If I quickly whip up a new class

This touches on my painpoints with NRTs: when I create a new class, my brain is in architecture mode and I find NRT warnings jarring. When it's existing code, I'm worried about legacy compatibility. I guess I just need to break out of my habits.

> interacting with something you wrote years ago or with a third-party library

I guess at these moments I should think of such interactions like apis. When you call them, validate/sanitize the result. Then you can forget about/ignore where it came from, knowing that for the rest of the code you've already validated it.

Mbucari:
>> When I create a new class, my brain is in architecture mode and I find NRT warnings jarring. When it's existing code, I'm worried about legacy compatibility. I guess I just need to break out of my habits.

I've been thinking about this statement for a while now. I believe that how we handle null objects _is_ part of architecture. Null references are inevitable. Often they're part of the design. How they are handled and what assumptions we make about where they may or may not be is part of the architecture of a solution. When you ignore the possibility of NRES because they're unlikely, but not impossible, you may make architectural choices that lead you into a blind alley. If you're forced to fix a design flaw in the future, you may find yourself having to undo a lot of architectural choices you'd made while discounting nre possibilities. Not being allowed to sweep under the rug small possibilities of NRES will lead to better architectural choices in the long run.

Issue #1185, opened 2025-03-09
Possible zip file corruption when log level is set to Debug


rmcrackan:
Per #1184 - when log level is set to Debug, error: `A local file header is corrupt.`

rmcrackan:
Fixed in PR #1189 . Will be available in the next release (coming soon)

rmcrackan:
[Version 12.0.3](https://github.com/rmcrackan/Libation/releases/latest) is now available.

Issue #1183, opened 2025-03-08
Scans library but is unable to download until reopening


myles2gobe4isleep:
Libation opens, scans my Audible library and updates with new books, but seems to freeze and does not start downloading. Closing Libation opens up an error window. Reopening Libation allows normal download of books.

1. Add new books to Audible library
2. Open Libation
3. Libation recognizes new books but does not start download
4. Close Libation - error window pops up
5. Reopen Libation
6. Previously scanned books download without issues

Windows 10

[_Log202503.zip](https://github.com/user-attachments/files/19140085/_Log202503.zip)

rmcrackan:
Does this happen if you're on another network? example: wi-fi at work, friend's house, phone hotspot, etc

rmcrackan:
[Version 12.0.3](https://github.com/rmcrackan/Libation/releases/latest) is now available.

Issue #1187, opened 2025-03-09
Crashes on opening: andcec


rmcrackan:
Continuation of #1182 , for @andcec

You can remove all accounts by just deleting the `AccountsSettings.json` file

Also, if you're comfortable mucking with settings, you can open `Settings.json` and set autoscan to false

andcec:
It send this is not the case. Removed all directories for Libation but still have the account connected. I'll try again and I'll key you know if this will solve the issue

andcec:
Remove Libation and dependencies. Start again Libation without any success.
Renamed the directory ~/Libation to ~/old:Libation
Start again Libation. Finally able to create my account and import my Library.

Fingrer crossed but it seems the problem is finally solved.
Thanks a lot for your patience and your help!

rmcrackan:
Thank you so much for working through this with me. I apologize for the headache. Glad it all worked out in the end.

andcec:
Sorry to bother you again but the program is not working.
It use a LOOOT of RAM and abort every time I'll try to liberate one audiobook.
And, after closing, I have to cancel the folder with settings (this time in ~/.local/share/Libation) anc recreate the account.
But still no luck when try to liberate some audiobooks (tried with 3 different books)

Mbucari:
@andcec Which desktop environment are you using?

rmcrackan:
[Linux EndeavourOS, Libation installed via yay package manager](https://github.com/rmcrackan/Libation/issues/1182#issuecomment-2708244270). I've never heard of this distro

andcec:
I know there are many Linux distros but this one is quite famous: I think is the most popular among the Arch-based distros. Anyway is using the same repos as Arch.
I've seen that on Debian-based there's a new version and I'll wait until I can get the update through AUR repos.
Let me know if I can do something in order to help you in sorting this problem.

Mbucari:
@rmcrackan @andcec The reason I asked about "desktop environment" is because when trying to install EndeavourOS on a VM, I'm presented with the following choice. I want to make sure I'm getting an Apples-to-apples comparison so I want to run whichever desktop that @andcec is running.

![Image](https://github.com/user-attachments/assets/c41d3290-59eb-43a7-ac9e-87a356c9be7f)

andcec:
I'm using i3 but I have also KDE of you think is easier for you to sort it all in KDE 😊

Mbucari:
[...]

PR #1189, opened 2025-03-10
Minor bug fixes.

Mbucari:
## [Add default window dimensions](https://github.com/rmcrackan/Libation/commit/65e12d9a8f805db6e7b8100a2ee7984e55f4e6c3)

These windows's contents wouldn't display correctly without initial Width and Height.

## [Handle corrupted LibraryScans.zip file](https://github.com/rmcrackan/Libation/commit/ee05ca4eb218596f5548f48f17b8ff42546f8554) (#1185)

If we can't open the LibraryScans.zip, try deleting it and creating a new one. Log all errors.

## [Remove inadvertently committed debugging code](https://github.com/rmcrackan/Libation/commit/4e067f5b5b2c58bffbbbd64b3d05a51fd1215a58)

My oopsie.

## [Fix Auto download not working sometimes](https://github.com/rmcrackan/Libation/commit/653381b1dfe6ac1be5a76cd316c34fea4a4de3cd) (#1183)

This error was crossed by cross-thread calls. Also, I changed `stats.booksNoProgress` to `tats.PendingBooks` because `PendingBooks` includes partially downloaded files.

### DIFF
--- a/Source/ApplicationServices/LibraryCommands.cs
+++ b/Source/ApplicationServices/LibraryCommands.cs
@@ -233,13 +233,42 @@ namespace ApplicationServices
             }
         }
 
+		private static LogArchiver? openLogArchive(string? archivePath)
+        {
+            if (string.IsNullOrWhiteSpace(archivePath))
+                return null;
+
+            try
+            {
+                return new LogArchiver(archivePath);
+            }
+            catch (System.IO.InvalidDataException)
+            {
+                try
+                {
+					Log.Logger.Warning($"Deleting corrupted {nameof(LogArchiver)} at {archivePath}");
+                    FileUtility.SaferDelete(archivePath);
+					return new LogArchiver(archivePath);
+				}
+                catch (Exception ex)
+				{
+					Log.Logger.Error(ex, $"Failed to open {nameof(LogArchiver)} at {archivePath}");
+				}
+            }
+            catch (Exception ex)
+            {
+                Log.Logger.Error(ex, $"Failed to open {nameof(LogArchiver)} at {archivePath}");
+            }
+            return null;
+        }
+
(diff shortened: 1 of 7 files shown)


rmcrackan:
Nice work, thanks. This feels like enough to merit a new release after the Dinah.Core stuff is done.

rmcrackan:
Want me to re-open them?

Mbucari:
@rmcrackan Well it looks like #1187 may be resolved. What do you think about changing AutoScan to default to false? If not, then I think you should go ahead and make a new release.

rmcrackan:
> They are open

Ah, you're right. I was thinking about the other 2 tickets that I closed in reply to this PR.

> AutoScan to default to false

I don't actually like AutoScan as true. I don't use AutoScan at all. I did this for the benefit of the 'don't make me think' users who wondered where their new books are. When I created it, I was getting a lot of those. I think for now I'll leave AutoScan default : true

wtanksleyjr:
May I suggest that defaulting autoscan to false is actually pretty sensible; the problem actually is that there's no "affordance" to guide the user to what they need to do. You get a spreadsheet (account info), then save it ... then wait and wonder what's next. What if the account info page had "scan all accounts" and "scan only new accounts" buttons, and maybe an additional column with a traffic light (or just a colorful lozenge) showing whether each account had been scanned?

CLHatch:
I would tend to agree that "false" is a more sensible setting for autoscan.  Especially since the user could attach accounts, then it starts downloading things before they have a chance to set up the rename patterns.

andcec:
> @rmcrackan Well it looks like #1187 may be resolved. What do you think about changing AutoScan to default to false? If not, then I think you should go ahead and make a new release.

As per yesterday I haven't any update i AUR so I'm stll with the .2 release. The program is quite working, but I'm experiencing errors with two audiobooks. As soon as possible I'll try with another one just to make another try. maybe I'll recreate the account just in case.
Anyway from now on I'll keep the AutoScan disabled, but I understand why it active by default

Issue #1079, opened 2024-12-14
Libation trying to download books that are not released yet


Tamba3998:
**Describe the bug**
Audible Japan recently(?) added a feature where you can "preorder" audiobooks that will get released with the flatrate plan at a future date. 
These audiobooks get picked up by Libation, but fail when you try to download them (because they are not released yet).
All of these titles have a "date pub'd" in the future in the misc column, and it's possible to filter them out with a search like 
datepublished:[20000101 to 20241214]

**To Reproduce**
1. Go to https://www.audible.co.jp/coming-soon and add any of the audiobooks with the orange button to your library. I'm assuming the same will happen with the regular preorders (the magenta buttons that are also available on the other region's Audible), but I have not tried it
2. Try to download them. The download will fail with the error "Book has no codec specified"

**Expected behavior**
Libation should ignore those books because the download can never work

**Platform**
Linux (EndeavourOS)

**Log Files**
The only relevant part in the log files is this error (which makes sense because it's trying to download something that is not released yet):
System.ApplicationException: Book has no codec specified. Cannot continue with download. 
   at AudibleApi.Api.getAllCodecsAsync(String asin) in /home/runner/work/AudibleApi/AudibleApi/AudibleApi/Api.Download.cs:line 247

rmcrackan:
Thank you for reporting this since it is a legit bug in relation to this new feature. I have no access to a .jp account so I'm not able to do anything about it right now. I'll leave this ticket open so it doesn't get lost.

As a bandaid: when your book does become available, if it has an error in Libation, you should be able to right-click > download to force Libation to attempt to download it again.

ajdonim:
Just wanted to add that this happened to me as well, but in the US and on MacOS. It occurred with _Amorph_ by Chene Lawson and N.K. Jemisin and _Hitchhikers_ by Ben H. Winters.

rmcrackan:
Fixed in #1223

Issue #996, opened 2024-10-10
Libation not downloading best available quality 


gitaiuser20:
Libation when set to "Audio quality to request from Audible: High" and "Download my books in the original audio format (Lossless)" is not getting the best available quality.

For the book Words of radiance by brandon sanderson read by kate reading and Michael kramer, when i attempt to download it i get a file that is 1.28 gb in size, when i set "Audio quality to request from Audible: Normal" i get a file that is around 600-700mb, however on the Audible app on android on high quality setting I'm getting a file that is 2.7ish gb, and has much better quality than the high quality one on audible. i believe this higher quality version is a relatively recent addition that may not have been available when the book first came out a few years ago.

rmcrackan:
I don't have Words of Radiance, but I do have Way of Kings in my library. Does this give you a similar discrepancy?

gitaiuser20:
Yes, i just attempted WoK, on android file size is 2539mb, the m4b from Libation is 1.21 GB when "Audio quality to request from Audible: High", and 616 mb when "Audio quality to request from Audible: Normal"

gitaiuser20:
update: I just tried to do the same with the book Leviathan Wakes by James F. A. Coley, and there is the same issue. The file I got from Libation set to high quality is 522 MB, whereas the file I got from Audible Android is 861 MB, and the one from Audible is much clearer.

mkb79:
@gitaiuser20 
Can you compare the files downloaded with Libation and the Android app using MediaInfo or another tool which show you the codec. I’m think you have activated „Spatial Audio“ on the Android Audible app? Then the „huge“ audiobook downloaded with Android can be in Dolby Atmos format.

gitaiuser20:
i tried turning off spatial audio and redownloaded, I'm getting the same size file.

as for checking them, when i try to look for the audiobook files on my android phone, i found loads of folders with exo files, so not sure what to check tbh

mkb79:
How do you know that the audiobook downloaded with the Android app are larger? You have to look for files with the extension `aax`, `aaxc` or `mp4` (if spatial is enabled)!

gitaiuser20:
when i click download on the app as it is downloading i get the file size.

now that I'm checking again, there are folders with names ending in .mp4, but inside it are more folders and inside them are the exo files.

mkb79:
[...]

Issue #597, opened 2023-04-29
Spatial Audio?


mslmn:
**Is your feature request related to a problem? Please describe.**
No. 

**Describe the solution you'd like**
Is there a difference between the files from standard Audible books and Spatial Audio ones? If so, can they be fetched by Libation?

**Additional context**
Example link:
https://www.audible.co.uk/pd/Zero-G-Audiobook/B07K3XYXFC

rmcrackan:
Yes. The only caveat is: make sure this item is **disabled** -- Settings, Audio File Options tab, Downsample stereo to mono. It will still download, but will sound wrong.

Mbucari:
>  Is there a difference between the files from standard Audible books and Spatial Audio ones?

Yes, but please don't ask me what they are.

> If so, can they be fetched by Libation?

Nope. Libation only downloads the standard stereo (actually duplicated mono) AAC audio files. In theory it _could_, but I don't own a device that supports Dolby Atmos so I can't reverse engineer what we'd need to do to get those files.

rmcrackan:
> Nope. Libation only downloads the standard stereo (actually duplicated mono) AAC audio files.

@Mbucari Why wouldn't Libation be able to support anything, as long as it's just a file wrapped in encryption?

Mbucari:
> @Mbucari Why wouldn't Libation be able to support anything, as long as it's just a file wrapped in encryption?

It probably could, but I don't know how to request a Dolby Atmos file. We can get High and Low quality, but those are your standard stereo audio.

![Screenshot_20230510-064118~2](https://github.com/rmcrackan/Libation/assets/37587114/543972e1-8b5e-4ae9-994a-75a26fe282c2)


[See the supported devices](https://help.audible.com/s/article/does-my-device-support-dolby-atmos-playback?language=en_US)

gilesfun:
> > @Mbucari Why wouldn't Libation be able to support anything, as long as it's just a file wrapped in encryption?
> 
> It probably could, but I don't know how to request a Dolby Atmos file. We can get High and Low quality, but those are your standard stereo audio.
> 
> 
> [See the supported devices](https://help.audible.com/s/article/does-my-device-support-dolby-atmos-playback?language=en_US)

I have a compatible device and could supply a pcap if you know of a way to disable certificate pinning for Audible Android.

CLHatch:
Seems my phone also supports it. Let me know if there's anything I can do to help you.

mkb79:
[...]

PR #1223, opened 2025-04-27
New features, including spatial audio support

Mbucari:
I've organized the commits fairly well, each one pertaining only to the changes called out in the commit title.

**Don't be too concerned with the 16,583 additions**. 15,195 of them are the Widevine/CDM code, and 14,552 line of that are an auto-generated profobuf file.

## [Move contributors to UI Base](https://github.com/rmcrackan/Libation/commit/0a9e489f48f7387d96f5a9002c61b6b474de1c92)

Move all contributors to the UI base to allow easier expansion in the future.

## [Increase tag template options for contributor and series types](https://github.com/rmcrackan/Libation/commit/7d806e0f3ecd24f47d95ee826837fdc7ef6c9da5)

This is second biggest feature of this PR.
- Add template tag support for multiple series
- Add series ID and contributor ID to template tags
- \<first author\> and \<first narrator\> are now name types with name formatter support
- Properly import contributor IDs into database
- Updated docs
- Added Tests

## [Mark unreleased books as unavailable](https://github.com/rmcrackan/Libation/commit/5d7ac699e66660ef876375ca152926b8e68e1d6d) (#1079)

Self-explanatory.

## [Allow re-adding completed queued items](https://github.com/rmcrackan/Libation/commit/e2aae85fd7a69de9833897118973f2e42fe70af6)

I thought this would be a nice quality of life feature, but I've reconsidered it. If a book is in the queue and completed successfully, users should have to clear it before re-adding it to prevent too many accidental duplicates. Reverted in [Revert "Allow re-adding completed queued items"](https://github.com/rmcrackan/Libation/commit/9ca879cc3d26b940eb0e4b0c3e89312c19bd0c86)

## [Add CDM API url list](https://github.com/rmcrackan/Libation/commit/bffaea6026d3395f6cef442ebc057a49564f37b1)

This is to address the good call you made in your [AudibleApi PR comment](https://github.com/rmcrackan/AudibleApi/pull/55#issuecomment-2833449132).

## [Add spatial audio support](https://github.com/rmcrackan/Libation/commit/ece48eb6d7d9059bfcb0117512a0d7a562a4b076)

This is the main feature. Downloading spatial audio (#597) and the highest quality audio(#996)

NOTE: I have used Adrm and AAX(C) interchangeably. Adrm is the name of the DRM used on AAX(C) files.

How this works:
1. A new option has been added to the settings quality combobox. The options are now "Normal", "High", and "Spatial (if available)". In order for users to select Spatial, the must have at least one account which has been registered as an android device (the latest [AudibleApi PR](https://github.com/rmcrackan/AudibleApi/pull/55)). If they don't, a message box will inform them that they need to remove and re-add their account to enable Spatial audio support.
2. Libation will get a CDM for the user from the API. If successful, the CDM is stored in the AccountSettings.json file. 
3. If we have a CDM in the AccountSettings, use it to request a widevine license. Regardless of the Quality setting, Libation will try to request a Widevine license if it has a valid CDM. This is to address #996. Unfortunately, it's not always clear if the Widevine audio files are better than the Adrm files, so I had to use some logic to decide which files to download. It's explained pretty well in the comments and logic flow of the [DownloadOptions.ChooseContent() method](https://github.com/Mbucari/Libation/blob/9ca879cc3d26b940eb0e4b0c3e89312c19bd0c86/Source/FileLiberator/DownloadOptions.Factory.cs#L34C1-L34C113).
4. If anything goes wrong in the chain from provisioning a CDM -> ... -> Licensing a Widevine Download, Libation falls back to downloading an Adrm audiobook (just like now). There is one exceptional case where Libation will fail to download the the Audiobook. If...
   1. the user has selected the "Spatial (if available)" quality option in their settings, and
   2. Audible grants a license for a spatial audiobook (still uncommon), and
   3. for some reason, the Audible DRM challenge/response fails (possibly due to blacklisting)

Then the download will fail without fallback to ADRM. I think this is appropriate because if the challenge/response fails, we will want to know about it. And the user can still download the title by changing their quality setting to "High".


.github/workflows/build-windows.yml
```diff
@@ -89,14 +89,6 @@ jobs:
...
           $bin_dir = "${{ matrix.os }}-${{ matrix.release_name }}\"
           $delfiles = @(
-            "libmp3lame.x64.so",
-            "libmp3lame.arm64.so",
-            "libmp3lame.x64.dylib",
-            "libmp3lame.arm64.dylib",
-            "ffmpegaac.x64.so",
-            "ffmpegaac.arm64.so",
-            "ffmpegaac.x64.dylib",
-            "ffmpegaac.arm64.dylib",
```
Mbucari, line 99: I re-jiggered the AAXClean.Codecs nuget package so only the native libs for the runtime identifier being built are copied to the output directory.

Source/AaxDecrypter/NetworkFileStream.cs
```diff
@@ -110,10 +115,14 @@ public NetworkFileStream(string saveFilePath, Uri uri, long writePosition = 0, D
...
-			RequestHeaders["Range"] = $"bytes={WritePosition}-";
 			try
 			{
-				Updated?.Invoke(this, EventArgs.Empty);
+				if (DateTime.UtcNow > NextUpdateTime)
+				{
+					Updated?.Invoke(this, EventArgs.Empty);
+					//JsonFilePersister Will not allow update intervals shorter than 100 milliseconds
+					NextUpdateTime = DateTime.UtcNow.AddMilliseconds(110);
+				}
```
Mbucari, line 125: The recent change to JsonFilePersister created a throttle to the downloader. I spent hours trying to figure out why it was so slow!

Source/AaxDecrypter/NetworkFileStream.cs
```diff
@@ -151,39 +159,82 @@ public async Task BeginDownloadingAsync()
...
+
+			try
+			{
+				using var client = new HttpClient();
+				while (WritePosition < ContentLength && !IsCancelled)
+				{
+					using var response = await RequestNextByteRangeAsync(client);
+					await DownloadToFile(response);
+				}
+			}
```
Mbucari, line 190: Unlike the ADRM download links which can be reliably downloaded in their entirety in a single Http connection, the widevine links tend to close the connection prematurely. Upon inspecting the Audible app's network traffic, I discovered that the App will only download 64MB chunks at a time. I changed the network stream to do the same.

Source/AudibleUtilities/AccountsSettings.cs
```diff
@@ -47,6 +47,22 @@ private List<Account> _accounts_json
...
+			get => _cdm;
+			set
+			{
+				if (value is null)
+					return;
+
+				_cdm = value;
+				update_no_validate();
+			}
+		}
```
Mbucari, line 64: Store the CDM file as a base64  string in the AccountSetting.json file. A CDM isn't linked with a specific account, so a user only need one CDM for all their accounts.

Source/AudibleUtilities/Widevine/Cdm.Api.cs
```diff
@@ -0,0 +1,189 @@
...
+using System.Net.Http;
+using AudibleApi.Cryptography;
+using Newtonsoft.Json.Linq;
+using Dinah.Core.Net.Http;
+using System.Text.Json.Nodes;
+
+#nullable enable
+namespace AudibleUtilities.Widevine;
+
+public partial class Cdm
```
Mbucari, line 15: This is my C# implementation of the https://github.com/devine-dl/pywidevine project.

Source/AudibleUtilities/Widevine/Device.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+	public DeviceTypes Type { get; }
+	public int FileVersion { get; }
+	public int SecurityLevel { get; }
+	public int Flags { get; }
+
+	public RSA CdmKey { get; }
+	internal ClientIdentification ClientId { get; }
+
+	public Device(Span<byte> fileData)
```
Mbucari, line 25: This is a CDM Devide file compatible with the https://github.com/devine-dl/pywidevine project.

Source/FileLiberator/DownloadDecryptBook.cs
```diff
@@ -122,15 +122,13 @@ private async Task<bool> downloadAudiobookAsync(LibraryBook libraryBook)
 
             downloadValidation(libraryBook);
 
-            var quality = (AudibleApi.DownloadQuality)config.FileDownloadQuality;
             var api = await libraryBook.GetApiAsync();
-            var contentLic = await api.GetDownloadLicenseAsync(libraryBook.Book.AudibleProductId, quality);
-			using var dlOptions = BuildDownloadOptions(libraryBook, config, contentLic);
 
+            using var dlOptions = await DownloadOptions.InitiateDownloadAsync(api, libraryBook, config);
```
Mbucari, line 127: Moved a lot of the download initialization code into the DownloadOptions class.

Source/FileLiberator/DownloadDecryptBook.cs
```diff
@@ -158,194 +156,40 @@ AaxcDownloadConvertBase converter
...
-        {
-            if (chapters.Count > 1 && chapters[0].Title == "Opening Credits")
-            {
-                chapters[1].StartOffsetMs = chapters[0].StartOffsetMs;
-                chapters[1].StartOffsetSec = chapters[0].StartOffsetSec;
-                chapters[1].LengthMs += chapters[0].LengthMs;
-                chapters.RemoveAt(0);
-            }
-            if (chapters.Count > 1 && chapters[^1].Title == "End Credits")
+		private void Converter_RetrievedMetadata(object sender, AAXClean.AppleTags tags)
```
Mbucari, line 171: Unlike the AAX(C) files, Mpeg-DASH wiles delivered by widevine contain no metadata.

Source/FileLiberator/DownloadOptions.Factory.cs
```diff
@@ -0,0 +1,362 @@
...
+	/// </summary>
+	public static async Task<DownloadOptions> InitiateDownloadAsync(Api api, LibraryBook libraryBook, Configuration config)
+	{
+		var license = await ChooseContent(api, libraryBook, config);
+		var options = BuildDownloadOptions(libraryBook, config, license);
+
+		return options;
+	}
+
+	private static async Task<ContentLicense> ChooseContent(Api api, LibraryBook libraryBook, Configuration config)
```
Mbucari, line 34: This method contains the logic for choosing Widevine or ADRM

Mbucari:
@rmcrackan If you want to test this, change the github url to point to my repo in this section:

https://github.com/Mbucari/Libation/blob/9ca879cc3d26b940eb0e4b0c3e89312c19bd0c86/Source/AudibleUtilities/Widevine/Cdm.Api.cs#L100-L102

Source/AaxDecrypter/NetworkFileStream.cs
rmcrackan: Oh shit, that might have been me.

rmcrackan:
I'm just reading through the above comments for now and I'm nodding and smiling a lot

> Increase tag template options for contributor and series types

If this addresses a specific ticket, include that # in here. Thanks for remembering the docs and tests

> possibly due to blacklisting

I completely forgot about this case. Thanks for remembering. If you haven't already: we may want protections against an accidental, malicious, or bug which could cause a loop of thinking one is blacklisted, fetching a new one, thinking it's blacklisted, etc, and potentially exhausting the source server's stockpile.

rmcrackan (approved):

rmcrackan:
> If you want to test this

I'd love to but you can imagine what work is like after being gone for 2 weeks. Also my boss's boss, and *his* boss are here from the UK this week, so it's even more fun. I'm so thankful that you have such a great track record for quality and responsiveness because I'm not sure when I'd be able to really review this batch.

Source/AaxDecrypter/NetworkFileStream.cs
Mbucari: And I and reviewed it! I feel silly it took so long to realize what was happening.

rmcrackan:
Sigh -- I guess we should have seen this coming. See comments: https://old.reddit.com/r/audiobooks/comments/1ka1vbu/libation_123_download_highest_quality_audio/

CLHatch:
> Sigh -- I guess we should have seen this coming. See comments: https://old.reddit.com/r/audiobooks/comments/1ka1vbu/libation_123_download_highest_quality_audio/

Yeah. :-D  My plan was just to redownload all my books. :-D

Mbucari:
@rmcrackan Yup, that would have been smart! I took a quick look at the raw responses from a book that has spatial audio, and I found this.

```JSON
  "asset_details": [
    {
      "is_spatial": true,
      "name": "Dolby"
    }
  ],
```

mkb79:
> @rmcrackan Yup, that would have been smart! I took a quick look at the raw responses from a book that has spatial audio, and I found this.
> 
> 
> 
> ```JSON
> 
>   "asset_details": [
> 
>     {
> 
>       "is_spatial": true,
> 
>       "name": "Dolby"
> 
>     }
> 
>   ],
> 
> ```

The `name` key is new to me. Will check this on iOS.

simmias42:
Is there a way in a future release to have the Widevine audio files be a selectable quality option (or checkmark option) and not automatic if the user has a CDM license? It looks like there's no way for Libation to grab those 64k AAX files anymore unless I'm missing something, and many users might prefer those instead of the Widevine ones for compatibility reasons.

We've been troubleshooting a new Libation user (he just downloaded it yesterday for the first time) for audiobookshelf who had no idea why many of his books couldn't be scanned, and it turns out it was all of the Widevine ones.

PR #1241, opened 2025-05-05
Fixed doubled first name in templates

cherez:
v12.3.0 caused a regression with contributors with a single word name, causing the name to be doubled. This was caused by using that name as both the first and last name, so swap the first name with the (blank) last name rather than duplicate them.

### DIFF
--- a/Source/LibationFileManager/Templates/ContributorDto.cs
+++ b/Source/LibationFileManager/Templates/ContributorDto.cs
@@ -24,10 +24,12 @@ public class ContributorDto : IFormattable
 
 		//Single-word names parse as first names. Use it as last name.
 		var lastName = string.IsNullOrWhiteSpace(HumanName.Last) ? HumanName.First : HumanName.Last;
+  		//Because of the above, if the have only a first name, then we'd double the name as "FirstName FirstName", so clear the first name in that situation.
+  		var firstName = string.IsNullOrWhiteSpace(HumanName.Last) ? HumanName.Last : HumanName.First;
 
 		return format
 			.Replace("{T}", HumanName.Title)
-			.Replace("{F}", HumanName.First)
+			.Replace("{F}", firstName)
 			.Replace("{M}", HumanName.Middle)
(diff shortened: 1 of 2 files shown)


rmcrackan:
@cherez : Thanks for the submission!

What happens if it's a single-name author and someone only uses the first name? I don't think this is likely and it seems preferable to doubling the name, but we'll still have a bug, albeit a better one.

@Mbucari : I'm leaning toward accepting this solution. Do you have an opinion? (This method could also use some unit tests but that's neither here nor there.)

rmcrackan:
Ah, I see ultimately this will be tested by the TemplatesTests unit tests. This PR will break some unit tests such as this one. Please update affected tests:

https://github.com/rmcrackan/Libation/blob/64a85b6aabab178d515ad34eaef33059e502d036/Source/_Tests/LibationFileManager.Tests/TemplatesTests.cs#L287

cherez:
> What happens if it's a single-name author and someone only uses the first name? I don't think this is likely and it seems preferable to doubling the name, but we'll still have a bug, albeit a better one.

I'm personally impartial to whether it's a first or last name, but since the existing code seemed to be there to make it the last name, I assumed there was a reason (sorting?). Most cases of single word names I can think of would be just first name (Plato), or a pen name (Voltaire). There are a number of examples of title + last name (Lord Byron) that I suspect would be wrong if we defaulted to first name, but I don't think there's a universal solution short of a comprehensive name database, which isn't feasible for a ton of reasons.



> Ah, I see ultimately this will be tested by the TemplatesTests unit tests. This PR will break some unit tests such as this one. Please update affected tests:

I confess, I don't have a C# build environment set up locally, so I can't immediately test this. I fixed the test you pointed out, and eyeballing the other tests I don't think this should change any others. If I missed any, let me know and I can update them as well.

Mbucari:
@rmcrackan Makes sense to me!

rmcrackan:
Merged

@cherez thanks again!

@Mbucari yup, I'm holding off

Issue #1246, opened 2025-05-08
Audiobook Truncation Caused by Metadata Fixes (AAX/C Downloads)


seanbob28:
**Bug Description**  
The setting "Allow Libation to fix up audiobook metadata" is truncating audio on some (older?) titles.

**To Reproduce**  
1. Identify an affected Audible title.
   (Some known titles include: "Shopgirl" [B002UUMPCO] and "A Christmas Carol" [B002V1AGEG])
2. Check the "Allow Libation to fix up audiobook metadata" setting.
3. Download the audiobook using the AAX/C ("High" quality) option.

**Platform**
Windows 10 -- Libation Classic v12.4.1.1

**Additional Troubleshooting**
- I downloaded the affected title twice (once with the "fix metadata" setting enabled, and once with it disabled).
- The AAXC file for both downloads is identical (binary same).
- However, the decrypted M4B files differ in size and duration (and metadata corrections obviously).

**Files**  
I've attached MediaInfo extracts (in JSON format) of the two M4B files so that you can see the audio file differences, such as:

_Don't Fix Metadata:_
"FileSize":"103733873",
"Duration":"13053.237",

_Fix Metadata:_
"FileSize":"103584234",
"Duration":"12983.762",

[Shopgirl \[B002UUMPCO\] NoMetadataFix.json](https://github.com/user-attachments/files/20095804/Shopgirl.B002UUMPCO.NoMetadataFix.json)
[Shopgirl \[B002UUMPCO\] MetadataFix.json](https://github.com/user-attachments/files/20095803/Shopgirl.B002UUMPCO.MetadataFix.json)

**Note**
This could probably be considered a duplicate of Issue #1062, which was closed due to the Widevine implementation. However, this bug appears to be specifically related to the AAX/C downloads. Just in case, I added a similar comment/update to that issue too.

Mbucari:
Thanks for the sleuthing, @seanbob28!

I analyzed Shopgirl B002UUMPCO, and there are a few different things going on here.

Firstly, the audio files delivered by widevine are different from the AAXC files. The widevine versions are, drumroll please, 1 minute and seven seconds shorter than the AAXC files. The chapter info delivered by the audible API is timestamped for the widevine files, not for the AAXC file. This is apparent not only from the total length, but also from listening to the audio at the chapter markers. The markers are correct for the widevine files, but they are a few seconds early for the AAXC files.

[...]

Issue #1236, opened 2025-05-03
aax not saved in v12.4.0


jostyee:
**Describe the bug**  
With "Retain the aax file after successfully decrypting" enabled, aax files not saved in the destination directories any more in new version of Libation, which was working with previous versions.

**Expected behavior**  
keep original aax files along with decrypted m4b files.

**Screenshots**  

<img width="889" alt="Image" src="https://github.com/user-attachments/assets/d130a813-349d-401d-b17a-58c485fead86" />

**Platform**  
Mac 15.4.1

ikanakova:
In version 12.2 or 12.3.1, aax did not save for me.

Alexander-R-Z:
I have the same issue

jostyee:
If "Use Widevine DRM" is checked, a dash file will be stored in the directory.

rmcrackan:
Fixed in [version 12.4.1](https://github.com/rmcrackan/Libation/releases/tag/v12.4.1)

rmcrackan:
Fixed in [latest version](https://github.com/rmcrackan/Libation/releases/latest) 12.4.2

Issue #1249, opened 2025-05-08
Docs update for new ability to download higher-quality content


wtanksleyjr:
**Is your feature request related to a problem? Please describe.**  
The new download method has special requirements for how to use it that I saw in one of the original issues that it was created to solve, something about removing and re-adding my accounts. The problem is that I can't find that comment anymore.

**Describe the solution you'd like**  
I thought it would be nice to add it to the docs at least as a "notes on upgrading old installations to download high quality content" or something like that. Not asking for a startup walkthrough update, just something for old users.

**Describe alternatives you've considered**  
I mean it would be nice if it could also explain what enabling the various switches means. I'm sure a lot of people want the Dolby audio, but I just want the normal high quality that's been out of my reach since Audible changed the API. This is how to do that, right?

Mbucari:
> something about removing and re-adding my accounts

There's a message box that appears when you go into your settings, try to enable widevine content, and your accounts are not registered as android devices. Once you re-add your accounts and they are registered with audible as android devices, then that message box no longer appears.

> it would be nice if it could also explain what enabling the various switches means.

Something more than the mouseover tips?

![Image](https://github.com/user-attachments/assets/908d587d-d69d-42c0-a39a-f6e5d1fe451c)

![Image](https://github.com/user-attachments/assets/53e8ee7d-d4cd-4d6a-b969-7765153e7d71)

![Image](https://github.com/user-attachments/assets/dc112272-e995-4291-bce9-4287c0ef45d6)

wtanksleyjr:
Oh, I see - my version is 12.3.1.1, it doesn't look like that. Good job reworking the user interface, I think; now that I see that, I think I "just know" how to make high quality actually reliably work (simply click widevine and follow the instructions but don't do spatial audio).

It feels like a more difficult decision than it really should be, but that's not your fault, we open-source people simply haven't done enough work with that codec yet.

May I ask, though, whether it might be possible to use other means to "register as android" - could I just log my android in to that account, for example? Would it be possible to implement a button in the account UI to upgrade the account? It's kind of annoying and confusing to delete an entry and then type exactly the same entry back in.

I guess I COULD just post a feature request, but if you think it's possible I suppose I will do that and close this. Otherwise, if it's not possible, I'll just close this (or you can).

Mbucari:
> It feels like a more difficult decision than it really should be

[...]

Issue #1228, opened 2025-04-30
Please add an optional checkbox or audio quality option for Widevine audio


simmias42:
Is there a way in a future release to have the Widevine audio files be a selectable quality option (or checkmark option) and not be automatic if the user has a CDM license?

For books that only had 64/22 AAX files as their highest quality option in the past, it doesn't look like there's a way for Libation to grab them anymore, and instead Widevine audio will be downloaded for users with a CDM license regardless of quality settings. Many users might prefer those 64k AAC-LC files instead of the Widevine ones for compatibility reasons.

We've been troubleshooting a new Libation user (he just downloaded it yesterday for the first time) for audiobookshelf who had no idea why many of his books couldn't be scanned, and it turns out it was all of the Widevine ones. Unfortunately, the only fix for now is to tell people to roll back to older versions of Libation.

Thanks!

rmcrackan:
This is improved in [latest releast](https://github.com/rmcrackan/Libation/releases/latest), v12.4.2

PR #1250, opened 2025-05-09
Bug fixes and a change to license request logic

Mbucari:
## [Fix dash files not being saved](https://github.com/rmcrackan/Libation/commit/28ba62aead001dd0245976a0552eba8f57351f1b) (#1236)

I fixed this for aaxc files, but then broke it for dash files. It now works for all encrypted file types.

## [Pad final chapter to prevent truncation from incorrect chapter info](https://github.com/rmcrackan/Libation/commit/b11a4887d7ae919428ce7ea522440c1b444dbd7e) (#1246)

As I described in that issue, incorrect chapter markers which are shorter than the audiobook will no longer truncate the audio file.

## [Use Libation settings to decide which DRM is downloaded.](https://github.com/rmcrackan/Libation/commit/19db226f5ac778e46cf65ecc5f36820d27badff8)

Since I added the fine-grained options for downloading widevine content in 12.4, it makes sense to only use those options to dictate which content Libation tries to download. I removed the fallback to traditional ADRM. If for whatever reason widevine content becomes unavailable, users can uncheck the "Use widevine DRM" option to revert back to the ADRM-onyl downloads.

### DIFF
--- a/Source/AudibleUtilities/ApiExtended.cs
+++ b/Source/AudibleUtilities/ApiExtended.cs
@@ -11,11 +11,13 @@ using Polly;
 using Polly.Retry;
 using System.Threading;
 
+#nullable enable
 namespace AudibleUtilities
 {
 	/// <summary>USE THIS from within Libation. It wraps the call with correct JSONPath</summary>
 	public class ApiExtended
 	{
+		public static Func<Account, ILoginChoiceEager>? LoginChoiceFactory { get; set; }
 		public Api Api { get; private set; }
 
 		private const int MaxConcurrency = 10;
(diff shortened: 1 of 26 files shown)


Source/LibationUiBase/LibationContributor.cs
```diff
@@ -41,7 +41,8 @@ public static IEnumerable<LibationContributor> AdditionalContributors
 			GitHubUser("muchtall"),
 			GitHubUser("ScubyG"),
 			GitHubUser("patienttruth"),
-			GitHubUser("stickystyle")
+			GitHubUser("stickystyle"),
+			GitHubUser("cherez"),
```
rmcrackan, line 45: good memory

Mbucari:
## [Fix occasional error of audio downloads hanging.](https://github.com/rmcrackan/Libation/pull/1250/commits/10c01f4147f49b5351862b5fe153af322deea89d)

I think I discovered the source of the bug which was causing some people's prowess to hang.
First of all, the step runner would hang with an unhandled exception. I fixed that.
Secondly, sometimes Audible would return more than 1 key, and only the second key was valid for the file. We now pass all possible keys down to AaxDecrypter and let it figure out which one to use.

## [Change ApiExtended to always allow provide login option](https://github.com/rmcrackan/Libation/pull/1250/commits/f35c82d59dfb63043f3711b6617ab75bcb69245d)

Previously, only some calls to `ApiExtended.CreateAsync()` would prompt users to login if necessary. Other calls would only work if the account already had a valid identity, and they would throw exceptions otherwise.

Changed ApiExtended so that the UI registers a static ILoginChoiceEager factory delegate that ApiExtended will use in the event that a login is required.

## [Ask users to clear the accounts when enabling widevine](https://github.com/rmcrackan/Libation/pull/1250/commits/ef67ae9d6a292fe5bd69a7df588888751d9bc162)

First, changed the "Use widevine DRM" default setting to "false". This should be opt-in.

Second, when a user does opt in and they still have iphone accounts, Libation asks them for permission to clear their accounts and prompt them for a login on the next download. If the choose "yes", Libation deletes and re-adds their account. Because of the changes made above, any account api action will trigger the login prompt.

rmcrackan:
Nice. This was quite a bit in a short amount of time.

> I discovered another bug that really needs to be fixed, but I'm done for a while.

Want me to release now? Or wait for that fix?

Mbucari:
> > I discovered another bug that really needs to be fixed, but I'm done for a while.
> 
> Want me to release now? Or wait for that fix?

Sorry for the confusion. I was referring to the bug that I fixed in this PR. I was just explaining why I converted to draft and added more.

> Nice. This was quite a bit in a short amount of time.

Yeah, it's actually a pretty critical bug which is why I was determined to fix it ASAP. It was just dumb luck that I didn't realize what was going on sooner, but this is probably affecting a lot of people.

Source/AudibleUtilities/ApiExtended.cs
```diff
@@ -24,52 +26,46 @@ public class ApiExtended
...
+						account.GetIdentityTokensJsonPath());
+				return new ApiExtended(api);
+			}
+			catch
 			{
-				Username = username.ToMask(),
-				LocaleName = localeName,
-			});
+				if (LoginChoiceFactory is null)
+					throw new InvalidOperationException($"The UI module must first set {LoginChoiceFactory} before attempting to create the api");
```
Mbucari, line 51: Oops, that should be a nameof() call. Would you mind fixing that please?

rmcrackan:
Something fucky is going on with my browser. I'm trying to add a comment to DownloadOptions.Factory.cs ~line 82 : Did you want to move this large explanation elsewhere?

rmcrackan:
I notified/closed these. Lemme know if this PR resolved any other tix: #1249 , #1246 , #1236

Mbucari:
> Something fucky is going on with my browser. I'm trying to add a comment to DownloadOptions.Factory.cs ~line 82 : Did you want to move this large explanation elsewhere?

The explanation is no longer relevant. Which license gets used is now determined by libation settings, no longer by heuristics.

Mbucari:
@rmcrackan this will improve #1228.

Also, the old fix for #1228 would often download the older, non widevine audio files if they were roughly the same size as the widevine files. Now, when "Use widevine DRM" is checked, Libation always tries to download widevine content. If it fails, it logs the error and will not fall back to the old ADRM license request.

Issue #1302, opened 2025-07-17
Automatic upgrade from version 12.4.5.1 -> 12.4.6 does not work


Quintinon:
**Describe the bug**  
I'm attempting to use the automatic upgrade feature, but cannot get it to actually auto-upgrade. The upgrade progress bar plays, but when closing and reopening libation nothing has changed and I'm re-prompted with a new version is available.

**To Reproduce**  

1. Have Libation 12.4.5 windows chardonnay downloaded, and extracted (in the downloads folder)
2. Note: This is version 12.4.5.1 in the about menu in libation
3. Open Libation (windows key, type libation, clicked icon, it found it from my downloads folder)
4. Notice The new version popup
5. Click yes
6. See the Upgrading bar in the bottom footer
7. Wait for it to finish and then disappear
8. Wait another minute or two just in case something else is happening in the background (I also tried without this step)
9. Close Libation
10. Wait a couple of minutes, notice nothing happens (I was kind of expecting a UAC prompt)
11. Re-open Libation (again with windows key, type libation, click icon. Only found instance is the 12.4.5 version in downloads folder), and see I'm still on version 12.4.5.1 in the about menu*Expected behavior**  

**Screenshots**  

**Platform**  
Windows 11 x64, Chardonnay, version 12.4.5.1

**Log Files**  

[_Log202507.zip](https://github.com/user-attachments/files/21298854/_Log202507.zip)

rmcrackan:
@Mbucari I assume this is related to the release index logic in the latest PR

Mbucari:
@Quintinon The auto-updater requires a utility called `ZipExtractor.exe` which ships alongside Libation. For some reason, Libation can't find it. My guess it it got erroneously flagged by your antivirus software and deleted. I'm affraid you'll need to manually download the latest release.

Mbucari:
@rmcrackan we should probably add a notification if the upgrade process fails.

Quintinon:
I don't mind manually downloading the latest release, so I'm just providing additional information to help diagnose the issue.

I can see ZipExtractor.exe in my Libation folder and windows has no recent items in the protection history indicating actions were taken.

<img width="730" height="701" alt="Image" src="https://github.com/user-attachments/assets/d880cb6c-ee91-431b-8741-adf79de8bc7a" />

Based on the logs it looks like its trying to run ZipExtractor from the wrong location.

[...]

PR #1303, opened 2025-07-17
Fix upgrade bug when Libation's working dir isn't program files dir

Mbucari:
Add MockUpgrader for testing the Upgrade process.
Fixes issue #1302

### DIFF
--- a/Source/LibationAvalonia/Views/MainWindow.axaml.cs
+++ b/Source/LibationAvalonia/Views/MainWindow.axaml.cs
@@ -166,20 +166,23 @@ namespace LibationAvalonia.Views
 		private void Configure_Upgrade()
 		{
 			setProgressVisible(false);
-#if !DEBUG
-			async System.Threading.Tasks.Task upgradeAvailable(LibationUiBase.UpgradeEventArgs e)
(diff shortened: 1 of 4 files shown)


rmcrackan:
Why not use `#if !DEBUG` instead of `// Local function is declared but never used` ? This is obviously just a style choice and I have no strong feelings. I'm just curious why the function is wrapped differently than it's caller/reference.

Mbucari:
> Why not use `#if !DEBUG` instead of `// Local function is declared but never used`

I wanted to add using statements to the top of the file and remove full qualified type names. But if I did that and performed and automatic code cleanup in debug mode, visual studio would see that those usings were not used in the file and remove them. Certainly is a style choice, and I'm not even convinced I made the right one lol.

Mbucari:
Oh yeah, I had one other reason that I think is actually a good reason. It's so that while in debug mode I can find usages of the upgrader and upgrade dialog types.

rmcrackan:
> It's so that while in debug mode I can find usages of the upgrader and upgrade dialog types.

Actually that's even easier. If they vanish when in debug, you'll get compile errors lol. Probably not what you meant

Issue #1310, opened 2025-07-22
Could not load file or assembly 'Serilog.Sinks.ZipFile, Version=3.1.1.0, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.


simdo01:
**Describe the bug**  

Upgrading to new release. upgrade failed. Won't start - see log - all books deleted.

**To Reproduce**  

**Platform**  
 Windows 11

2025-07-22 07:20:31.061 -05:00 [INF] (at AppScaffolding.LibationScaffolding.logStartupState(LibationFileManager.Configuration)) 

 {}2025-07-22 07:20:31.168 -05:00 [WRN] (at LibationFileManager.InteropFactory.CurrentDomain_AssemblyResolve(System.Object, System.ResolveEventArgs)) Unable to load module Windows, Culture=neutral, PublicKeyToken=null, ContentType=WindowsRuntime
 {}2025-07-22 07:20:31.178 -05:00 [WRN] (at LibationFileManager.InteropFactory.CurrentDomain_AssemblyResolve(System.Object, System.ResolveEventArgs)) Unable to load module Microsoft.Windows.SDK.NET, Culture=neutral, PublicKeyToken=null
[...]

PR #1311, opened 2025-07-22
Fix serilog dynamic assembly loading issue (#1310)

Mbucari:
This issue was caused because the Serilog.Sinks.ZipFile.dll was still in the program files directory, and by default Serilog tries to load all "serilog.*" dlls dynamically. The error is because Serilog uses "Assembly.Load", and that will fail if the assembly hasn't already been loaded into the AppDomain.


rmcrackan:
Thanks for the quick fix!

PR #1316, opened 2025-07-28
Added nix flake and shell files for nixos developers

ajundi:
Added nix flake and shell files for nixos developers who want to work on development. You can use flake method => 'nix develop' or non flake method =>`nix-shell'

The shell.nix file is used for both flake and non-flake invokations. The lock file is also set at a version where the project works. Note the none-flake method will follow the version of the system and isn't guaranteed to work on older installations if they haven't been updated in a while.

I only tested it on linux-64.

Added vscode entries to launch.json and task.json that allows debugging on linux hosts.


rmcrackan:
I don't know anything about this -- is there any documentation that needs to be added? Or will the relevant people understand?

ajundi:
I can add instructions for setting up the environment for NixOS. Which file do you want the instructions to be in?

A nix user who has used nix for development before most likely will know what to do but It is generally one of two commands depending on user preference.
`nix-shell`
or if the user wants to use the pinned version which is what happens using flakes. then command is
`nix develop`
these commands are ran from the directory where the flake.nix and shell.nix files are located. Then the environment will be setup and they can then run vscode or vscodium or any other IDE they have on their system from that shell environment.  They can't run vscode from outside that shell if they want the dependencies Avalonia needs to be recognized within vscode as it inherits the shell environment.

rmcrackan:
I'd completely forgotten this existed -- https://github.com/rmcrackan/Libation/blob/master/Documentation/InstallOnLinux.md#nixos

You can either edit this. Or lemme know if it's fine as-is and I'll merge your PR.

ajundi:
The install instructions don't work for setting up a build and debug environment. 

I can add a document for setting up the development environment for build and debug for Nixos. The shell.nix contains a list of dependencies that are required for Avalonia to run on Linux as well as the Dotnet SDK and DOTNET_ROOT env variable that's needed to find the runtime on the system. There might be other dependencies needed for Linux that varies depending on the desktop environment being used. On windows it is much simpler so documentation is not needed for debug and build.

rmcrackan:
Feel free to edit the document above, or you can create your own markdown doc and link to it from that section in the linux page -- whichever you prefer. I don't feel strongly either way.

ajundi:
I added documentation and simplified the flake.nix file. I also pulled latest and rebased my commits as a new commit.

rmcrackan:
This looks very good. Thank you!

Issue #1336, opened 2025-08-15
Can't download audiobooks with widevine enabled


Graukolos:
**Describe the bug**
All audiobook downloads fail if the "Use Widevine DRM" option in the settings is on.

**To Reproduce**
1. Toggle the "Use widevine DRM" setting
2. login to audible account again
3. try to download any audiobook
4. Get the prompt that an error occured

I have logged in through the external browser method, as the default password login has never worked for me somehow.
From some of the other issues I gathered that the "device" Libation presents to audible switched from iPhone to Android to get widevine to work, maybe there is a problem with the browser login + pretending to be an android device?

I have tested both with the version that came out an hour ago (12.5.1.1) and the one before that.
The attached log is from the current version with log level set to debug.

Thank you for this awesome project :D

**Platform**
Fedora 42 amd64
Libation 12.5.1.1

**Log Files**
[Log202508 (Copy).log](https://github.com/user-attachments/files/21800530/Log202508.Copy.log)

Mbucari:
@Graukolos an you confirm that this is also happening on [12.5.0](https://github.com/rmcrackan/Libation/releases/tag/v12.5.0)?

Graukolos:
yes, just tested it

Mbucari:
@rmcrackan This is insanely weird.  It looks like something broke inside dotnet between 12.5.0 and 12.5.1, because it's a crypto engine error. It's only happening on linux which uses OpenSSL under the hood. I downloaded the 12.5.0 source package and built it with rider on Fedora 43 and it's now throwing the same error, so it's not anything we did (i.e. any packages we updated between 12.5.0. and 12.5.1).

It MUST be a difference in the build system, but I compared the workflow runs of [12.5.0](https://github.com/rmcrackan/Libation/actions/runs/16756346074/job/47439891827) and [12.5.1](https://github.com/rmcrackan/Libation/actions/runs/16995262814/job/48184347264) and they are both using dotnet 9.0.304.

I'm truly at a loss. I have no idea what changed or why the crypto operation is failing.

rmcrackan:
I bet it was one of these dependabot things #1330 , #1324

rmcrackan:
Hmm, those are only about download/checkout, not build...

Mbucari:
[...]

PR #1331, opened 2025-08-11
Add audio format info to wiki

Mbucari:
I deleted your DolbyAtmos.md file (sorry) and replaced it with AudioFileFormats.md. xHE-AAC was being conflated with spatial audio, so I fixed that and added explanations of the various audio formats and pertinent settings.

PR #1337, opened 2025-08-16
Fix linux crpto and series order parsing

Mbucari:
## [Use managed RSASSA-PSS with SHA-1](https://github.com/rmcrackan/Libation/pull/1337/commits/5ceda408da88a1cf011427e602be112688ab15f2)

As hard as this is to believe, the path of least resistance to solve #1336 was to implement the signing algorithm in managed code.

## [Parse floats with invariant culture](https://github.com/rmcrackan/Libation/pull/1337/commits/bfa0e4d338270fac713ee7bea3482789c416c298)

Fix the issue I identified [here](https://github.com/rmcrackan/Libation/pull/1331#discussion_r2279967145).

### DIFF
--- a/Source/AudibleUtilities/Widevine/Device.cs
+++ b/Source/AudibleUtilities/Widevine/Device.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using System.Numerics;
 using System.Security.Cryptography;
 
 #nullable enable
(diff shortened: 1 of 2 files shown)


Mbucari:
I don't know who needs to hear this, but absolutely nobody anywhere should use this RSASSA-PSS implementation in anything where they care about security at all. I completely skipped the random salt part of it because libation doesn't need security; it only needs to satisfy Audible server's challenge-response requirements.

rmcrackan:
> As hard as this is to believe, the path of least resistance to solve 1336 was to implement the signing algorithm in managed code.

I don't know how to feel about this. Hilarious? Horrifying? In awe?

> Fix the issue I identified

I'm glad you thought of this one. If I remember, I take care of i18n just fine. But I so rarely remember.

> I don't know who needs to hear this, but absolutely nobody anywhere should use this RSASSA-PSS implementation in anything where they care about security at all. I completely skipped the random salt part of it because libation doesn't need security; it only needs to satisfy Audible server's challenge-response requirements.

This is really good info to know. Maybe throw it into the comments for future us.

rmcrackan:
I added your security note to the top of that class

PR #1343, opened 2025-08-20
Place examples in their own line

MajorTanya:
Markdown collapses single line breaks, so this change makes it so the examples will have their own lines. This helps readers of the documentation to quickly understand what the actual templating syntax is.

| Before | After |
| ------- | ------- |
| before | after |


rmcrackan:
Good catch. Thanks much!

PR #1351, opened 2025-08-27
Create a not-has tag to enable conditionals based on missing tags.

Surye:
This is missing tests, largely because I am not sure how the testing is being done. However it should be simple to add, as the changes are minor, and are usable for my use case.


rmcrackan:
This is really lovely work. We actually have this feature but clearly we didn't make it very evident.

You can do this with `<!has `

https://github.com/rmcrackan/Libation/blob/ed5afe5d0f416707a1a9f44fe4047683ff7e69c3/Documentation/NamingTemplates.md?plain=1#L90

I'll add some examples in the documentation.

Thank you though. I really do appreciate that you took it upon yourself to understand and implement this.

rmcrackan:
I added more obvious usage for inverted tags here #1352

rmcrackan:
I also like your standalones example: #1353

PR #1352, opened 2025-08-27
Update NamingTemplates.md

rmcrackan:
Better documentation for inverted tags

Address deficient documentation which led to #1351


rmcrackan:
@Mbucari Lemme know if I should take this further

Issue #1375, opened 2025-09-27
Add "Included until" field for books that are free and have an expiration date.


delebash:
**Describe the solution you'd like**  
For free books that you can add to your library some have an expiration date. The expiration date is shown as "Included until 10/07(whatever the date is)".  It would be nice to be able to filter books with expiration dates so you can download them before they expire.

**Additional context**  
Here is how it looks in the grid view

<img width="1367" height="317" alt="Image" src="https://github.com/user-attachments/assets/03c8e790-5aba-44ab-bc94-7f504cea0e2a" />

and selected book view

<img width="815" height="262" alt="Image" src="https://github.com/user-attachments/assets/f8c75ef1-b5d1-4501-b548-f50baec5e592" />

In the json file you are looking for the end_date in the plans array

```
  "plans": [
        {
          "customer_eligible": null,
          "detail_plan_names": null,
          "end_date": "2099-12-31T00:00:00Z",
          "plan_name": "AccessViaMusic",
          "start_date": "1970-01-01T00:00:00Z"
        },
        {
          "customer_eligible": null,
          "detail_plan_names": null,
          "end_date": "2025-10-07T12:00:00Z",
          "plan_name": "US Minerva",
          "start_date": "2024-11-15T18:27:00.00027Z"
        },
        {
          "customer_eligible": null,
          "detail_plan_names": null,
          "end_date": "2025-10-27T18:31:29.948030687Z",
          "plan_name": "SpecialBenefit",
          "start_date": "2025-09-26T18:31:29.948030687Z"
        }
      ],
```

For me the correct end_date is for the plan_name = "US Minerva"
I am not sure what SpecialBenefit plan name means but if you exclude both SpecialBenefit and AccessViaMusic you should get the correct data.  I have searched my json file and this is the conclusion I have come to by looking at the plans array for many of my books with an end_date.  Maybe include the Special Benefit date as well, I am guessing some users have this special benefit which extends the end_date.

delebash:
merged https://github.com/rmcrackan/Libation/pull/1388

PR #1388, opened 2025-10-17
Added IncludedUntil Date

delebash:
Resolves issue https://github.com/rmcrackan/Libation/issues/1375
This date shows when an audible plus or free book will be removed from your library.

image

### DIFF
--- a/Source/LibationUiBase/GridView/GridEntry.cs
+++ b/Source/LibationUiBase/GridView/GridEntry.cs
@@ -130,9 +130,10 @@ namespace LibationUiBase.GridView
 
 		protected abstract string GetBookTags();
 		protected virtual DateTime GetPurchaseDate() => LibraryBook.DateAdded;
+		protected virtual DateTime? GetIncludedUntil() => LibraryBook.IncludedUntil;
 		protected virtual int GetLengthInMinutes() => Book.LengthInMinutes;
 		protected string GetPurchaseDateString() => GetPurchaseDate().ToString("d");
-		protected string GetIncludedUntilString() => LibraryBook.IncludedUntil?.ToString("d") ?? string.Empty;
+		protected string GetIncludedUntilString() => GetIncludedUntil()?.ToString("d") ?? string.Empty;
 		protected string GetBookLengthString()
 		{
 			int bookLenMins = GetLengthInMinutes();
@@ -213,7 +214,7 @@ namespace LibationUiBase.GridView
 			nameof(Liberate) => Liberate,
 			nameof(DateAdded) => DateAdded,
 			nameof(IsSpatial) => IsSpatial,
-			nameof(IncludedUntil) => IncludedUntil,
+			nameof(IncludedUntil) => GetIncludedUntil() ?? default,
 			_ => null
 		};
 		
--- a/Source/LibationUiBase/GridView/SeriesEntry.cs
+++ b/Source/LibationUiBase/GridView/SeriesEntry.cs
@@ -102,5 +102,6 @@ namespace LibationUiBase.GridView
 		protected override string GetBookTags() => null;
 		protected override int GetLengthInMinutes() => Children.Count == 0 ? 0 : Children.Sum(c => c.LibraryBook.Book.LengthInMinutes);
 		protected override DateTime GetPurchaseDate() => Children.Count == 0 ? default : Children.Min(c => c.LibraryBook.DateAdded);
+		protected override DateTime? GetIncludedUntil() => Children.Count == 0 ? default : Children.Min(c => c.LibraryBook.IncludedUntil);
 	}
 }


rmcrackan:
I really like this idea. Some details:

* I don't think this belongs with `Book` though. It's not really about the book, it's about the owner's relationship to the book. It feels like it belongs in LibraryBook.
* A value which is typically null doesn't feel like it belongs in the ctor. I'd use something like your `UpdateIncludedUntil`. Speaking of which, I don't see that used anywhere.
* I assume plan.IsAyce flag is only true if you are borrowing it, and it is false in cases where you own it outright and it also happens to be in the all-you-can-eat catalog?
* What happens if: you're borrowing a book, decide to purchase it, and do a new scan. Will it live-update this field in the grid?

@Mbucari You have as much of an intuition as anyone about our model. Do you have an opinion?

Note to myself: after this minutia is worked out, add to search indexing, search options, and exports.

Mbucari:
@rmcrackan 100% `IncludedUntil` needs to be in `LibraryBook`, not `Book`.

@delebash when you move `IncludedUntil` to `LibraryBook`, you'll need to move your import logic into `LibraryBookImporter`. When you do that you'll need to cover the edge case where Account A has an AYCL title that is expiring and Account B owns that same title outright. In that case, Only Account B's copy should be imported.

delebash:
I will move it to LibraryBook and LibraryBookImporter as well as addressing the other issues.  Thanks for the feedback.

delebash:
> * I assume plan.IsAyce flag is only true if you are borrowing it, and it is false in cases where you own it outright and it also happens to be in the all-you-can-eat catalog

I am making the above assumption. Looking at a some of my books that I bought IsAyce is false.  I don't have any books that where originally audible plus and that I bought.  I can't tell if when you purchase a book it is no longer considered Ayce.  

If anyone can test this that would be great. 

>* What happens if: you're borrowing a book, decide to purchase it, and do a new scan. Will it live-update this field in the grid?

Assuming IsAyce changes to false after you purchase a book then during the LibraryBookImporter upsertLibraryBooks then includedUntil will be set to null.  So as long as the grid reflects the new values from upsertLibraryBooks the grid should update correctly.

delebash:
>* When you do that, you'll need to cover the edge case where Account A has an AYCL title that is expiring, and Account B owns that same title outright. In that case, Only Account B's copy should be imported.

From the upsertLibraryBooks function comments Libation does not allow import of the same book for 2 different accounts.

		// currently, inserting LibraryBook will throw error if the same book is in multiple accounts for the same region.
			//
			// CURRENT SOLUTION: don't re-insert
			
Everything is based on the assumption that IsAyce changes to false when you buy the book.  This needs to be verified. I just don't have any books that are audible plus books that I want to buy to test this.  If IsAyce does not change to false than we need to find another solution.  If you all think of a better way to determine when an audible plus book is going to expire let me know.  TY.

Originally, I was just looking for certain plans to determine if a book was audible plus for me that plan name seems to be US Minerva but I am not sure if that holds true for everyone.  The books that I have purchased do not even have plans besides the "AccessViaMusic" and "SpecialBenefit"

From the AudibleApi IsAyce looks for plan names

		public bool IsAyce => PlanName.ContainsInsensitive("Minerva") ||
				PlanName.ContainsInsensitive("AYCL") ||
				PlanName.ContainsInsensitive("Free");

delebash:
I have one problem in the LibraryCommands.cs file line 214
`                book = new LibraryBook(bookImporter.Cache[importItem.DtoItem.ProductId], importItem.DtoItem.DateAdded, importItem.AccountId,DateTime.Now);`  DatTime.Now is just a placeholder for IncludeUntil until I get it working correctly.

I do not have a way to get the IncludeUntil value from when I add it in the LibraryBookImporter.cs

```
	var libraryBook = new LibraryBook(
						bookImporter.Cache[item.DtoItem.ProductId],
						item.DtoItem.DateAdded,
						item.AccountId,
						includedUntil)
```
It looks like I need to add IncludedUntil to the importItem.DtoItem but I do not see how to add it. Do the DtoItem properties come from the AudibleApi?

rmcrackan:
> I have one problem in the LibraryCommands.cs file line 214

If I'm reading this code correctly, you figured this out, right?

Mbucari:
@delebash I tried testing your changes, and get the following error:

```
SqliteException: SQLite Error 1: 'no such column: l.IncludedUntil'.
```

It looks like you haven't properly migrated the data model.  [Read about adding migrations here](https://learn.microsoft.com/en-us/ef/core/managing-schemas/migrations/?tabs=dotnet-core-cli).

rmcrackan:
> ```
> SqliteException: SQLite Error 1: 'no such column: l.IncludedUntil'.
> ```

I saw the one changed migration file and it didn't register that I didn't see the other 2 new ones that result from creating a migration.

Source/DataLayer/EfClasses/LibraryBook.cs
```diff
@@ -13,16 +13,18 @@ public class LibraryBook
 
         public bool IsDeleted { get; set; }
         public bool AbsentFromLastScan { get; set; }
-
+        
+        public DateTime? IncludedUntil { get; private set; }
         private LibraryBook() { }
-        public LibraryBook(Book book, DateTime dateAdded, string account)
+        public LibraryBook(Book book, DateTime dateAdded, string account, DateTime? includedUntil)
```
Mbucari: includedUntil should be removed from the constructor. Use a setter like `SetAccount` below.

Mbucari:
@delebash I rewrote your changes to LibraryBook.cs and LibraryBookImporter.cs. I can't directly edit your PR, so please copy the below changes into your PR.

<details>
  <summary>LibraryBook.cs</summary>
  
  ```C#
  using System;
  using Dinah.Core;
  
  namespace DataLayer
  {
      public class LibraryBook
      {
          internal int BookId { get; private set; }
          public Book Book { get; private set; }
  
          public DateTime DateAdded { get; private set; }
          public string Account { get; private set; }
  
          public bool IsDeleted { get; set; }
          public bool AbsentFromLastScan { get; set; }
          
          public DateTime? IncludedUntil { get; private set; }
          private LibraryBook() { }
          public LibraryBook(Book book, DateTime dateAdded, string account)
          {
              ArgumentValidator.EnsureNotNull(book, nameof(book));
              ArgumentValidator.EnsureNotNull(account, nameof(account));
  
              Book = book;
              DateAdded = dateAdded;
              Account = account;
          }
  
		  public void SetAccount(string account) => Account = account;
		  public void SetIncludedUntil(DateTime? includedUntil) => IncludedUntil = includedUntil;
		  public override string ToString() => $"{DateAdded:d} {Book}";
	  }
  }
  ```
  
</details>

<details>
  <summary>LibraryBookImporter.cs</summary>
  
  ```C#
  using System;
  using System.Collections.Generic;
  using System.Linq;
  using AudibleUtilities;
  using DataLayer;
  using Dinah.Core;
  using Dinah.Core.Collections.Generic;
  
  namespace DtoImporterService
  {
	  public class LibraryBookImporter : ItemsImporterBase
	  {
		  protected override IValidator Validator => new LibraryValidator();
  
		  private BookImporter bookImporter { get; }
  
		  public LibraryBookImporter(LibationContext context) : base(context)
		  {
			  bookImporter = new BookImporter(DbContext);
		  }
  
		  protected override int DoImport(IEnumerable<ImportItem> importItems)
		  {
			  bookImporter.Import(importItems);
  
			  var qtyNew = upsertLibraryBooks(importItems);
			  return qtyNew;
		  }
  
		  private int upsertLibraryBooks(IEnumerable<ImportItem> importItems)
		  {
			  // technically, we should be able to have duplicate books from separate accounts.
			  // this would violate the current pk and would be difficult to deal with elsewhere:
			  // - what to show in the grid
			  // - which to consider liberated
			  //
			  // sqlite cannot alter pk. the work around is an extensive headache
			  // - update: now possible in .net5/efcore5
			  //
			  // currently, inserting LibraryBook will throw error if the same book is in multiple accounts for the same region.
			  //
			  // CURRENT SOLUTION: don't re-insert
  
  
			  //When Books are upserted during the BookImporter run, they are linked to their LibraryBook in the DbContext
			  //instance. If a LibraryBook has a null book here, that means it's Book was not imported during by BookImporter.
			  //There should never be duplicates, but this is defensive.
			  var existingEntries = DbContext.LibraryBooks.AsEnumerable().Where(l => l.Book is not null).ToDictionarySafe(l => l.Book.AudibleProductId);
  
			  //If importItems are contains duplicates by asin, keep the Item that's "available"
			  var uniqueImportItems = ToDictionarySafe(importItems, dto => dto.DtoItem.ProductId, tieBreak);
  
			  int qtyNew = 0;			
  
			  foreach (var item in uniqueImportItems.Values)
			  {
				  if (existingEntries.TryGetValue(item.DtoItem.ProductId, out LibraryBook existing))
				  {
					  if (existing.Account != item.AccountId)
					  {
						  //Book is absent from the existing LibraryBook's account. Use the alternate account.
						  existing.SetAccount(item.AccountId);
					  }
  
					  existing.AbsentFromLastScan = isUnavailable(item);
				  }
				  else
				  {
					  existing = new LibraryBook(
						  bookImporter.Cache[item.DtoItem.ProductId],
						  item.DtoItem.DateAdded,
						  item.AccountId)
						  {
							  AbsentFromLastScan = isUnavailable(item)
						  };
  
					  try
					  {
						  DbContext.LibraryBooks.Add(existing);
						  qtyNew++;
					  }
					  catch (Exception ex)
					  {
						  Serilog.Log.Logger.Error(ex, "Error adding library book. {@DebugInfo}", new { existing.Book, existing.Account });
					  }
				  }
				  
				  existing.SetIncludedUntil(GetExpirationDate(item));
			  }
  
			  var scannedAccounts = importItems.Select(i => i.AccountId).Distinct().ToList();
  
			  //If an existing Book wasn't found in the import, the owning LibraryBook's Book will be null.
			  //Only change AbsentFromLastScan for LibraryBooks of accounts that were scanned.
			  foreach (var nullBook in DbContext.LibraryBooks.AsEnumerable().Where(lb => lb.Book is null && lb.Account.In(scannedAccounts)))
				  nullBook.AbsentFromLastScan = true;
  
			  return qtyNew;
		  }		
  
		  private static Dictionary<TKey, TSource> ToDictionarySafe<TKey, TSource>(IEnumerable<TSource> source, Func<TSource, TKey> keySelector, Func<TSource, TSource, TSource> tieBreaker)
		  {
			  var dictionary = new Dictionary<TKey, TSource>();
  
			  foreach (TSource newItem in source)
			  {
				  TKey key = keySelector(newItem);
  
				  dictionary[key]
					  = dictionary.TryGetValue(key, out TSource existingItem)
					  ? tieBreaker(existingItem, newItem)
					  : newItem;
			  }
			  return dictionary;
		  }
  
		  private static ImportItem tieBreak(ImportItem item1, ImportItem item2)
			  => isUnavailable(item1) && !isUnavailable(item2) ? item2 : item1;
  
		  private static bool isUnavailable(ImportItem item)
			  => isFutureRelease(item) || isPlusTitleUnavailable(item);
  
          private static bool isFutureRelease(ImportItem item)
			  => item.DtoItem.IssueDate is DateTimeOffset dt && dt > DateTimeOffset.UtcNow;
  
		  private static bool isPlusTitleUnavailable(ImportItem item)
			  => item.DtoItem.ContentType is null
			  || (item.DtoItem.IsAyce is true && item.DtoItem.Plans?.Any(p => p.IsAyce) is not true);
  
		  /// <summary>
		  /// Determines when your audible plus or free book will expire from your library  
		  /// plan.IsAyce from underlying AudibleApi project determines the plans to look at, first plan found is used.
		  /// </summary>
		  /// <returns>The DateTime that this title will become unavailable, otherwise null</returns>
		  private static DateTime? GetExpirationDate(ImportItem item)
			  => item.DtoItem.Plans
			  ?.Where(p => p.IsAyce)
			  .Select(p => p.EndDate)
			  .FirstOrDefault(end => end.HasValue && end.Value.Year is not (2099 or 9999))
			  ?.DateTime;
	  }
  }
  ```
  
</details>

Source/LibationAvalonia/Views/ProductsDisplay.axaml
```diff
@@ -213,6 +213,16 @@
 					</DataGridTemplateColumn.CellTemplate>
 				</controls:DataGridTemplateColumnExt>
 				
+				<controls:DataGridTemplateColumnExt Header="Included Until" MinWidth="10" Width="{CompiledBinding IncludedUntilWidth, Mode=TwoWay}" CanUserSort="True" SortMemberPath="IncludedUntil" ClipboardContentBinding="{Binding IncludedUntil}">
```
Mbucari: Change header to `Included&#xA;Until` so that it wraps.

Mbucari:
You should add `IncludedUntil` to the `DefaultColumns` list so that it isn't displayed by default.

https://github.com/rmcrackan/Libation/blob/2c882e883d49c01d7428437e1f7ea51ab5df2f07/Source/LibationFileManager/Configuration.PersistentSettings.cs#L209-L213

delebash:
@Mbucari Thanks for the changes. You are checking for existing books and skipping import.  So I think we need also add an check in LibraryBookImporter where you already check to see if the IsAyce=false then we set IncludedUntil to null. Does that seem correct?

```
  existing.AbsentFromLastScan = isUnavailable(item);
			      if (existing.AbsentFromLastScan)
			      {
				      existing.SetIncludedUntil(null);
			      }
```

Mbucari:
> You are checking for existing books and skipping import. So I think we need also add an check in LibraryBookImporter where you already check to see if the IsAyce=false then we set IncludedUntil to null. Does that seem correct?

No. Every `ImporterBase` derrived class is working with a single copy of Libations' DbContext. LibraryBookImporter tries to find an existing copy of the LibraryBook in the DbContext. If it isn't found, LibraryBookImporter creates it and adds it to the DbContext. If it _is_ found, then LibraryBookImporter can change that instances' properties, and those changes will be saved when the import is finalized and `DbContext.SaveChanges()` is called.

My change to LibraryBookImporter gets or adds the LibraryBook, then sets `IsUnavailable` on that LibraryBook instance.

Mbucari:
@delebash You're still missing the migration code.

In the developer power shell, navigate into Source\DataLayer
Then run this command:
```console
dotnet ef migrations add AddIncludedUntil
```

Three files will be created:
20251020#####_AddIncludedUntil.cs
20251020#####_AddIncludedUntil.Designer.cs
LibationContext.db

You can delete LibationContext.db, but keep the other 2 .cs files.

rmcrackan:
> Three files will be created:

It should also overwrite the manual changes in the other migration file.

Source/DataLayer/Migrations/LibationContextModelSnapshot.cs
```diff
@@ -194,6 +194,9 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.Property<DateTime>("DateAdded")
                         .HasColumnType("TEXT");
 
+                    b.Property<DateTime?>("IncludedUntil")
+                        .HasColumnType("TEXT");
+
```
Mbucari: Revert this before you run `dotnet ef migrations add AddIncludedUntil`.

delebash:
Just forgot to add them to git and commit.

delebash:
Ok let me revert and re add migration.

Mbucari:
> It should also overwrite the manual changes in the other migration file.

I just tested, and it does NOT. I added the migration without reverting the LibationContextModelSnapshot.cs changes, and I still got the SQL exception. I first had to revert LibationContextModelSnapshot, then add the migration.

delebash:
If I start from a fresh database everything works as expected.  I had trouble actually running the ef remove command, so I just deleted my database>deleted the migration file>manually removed IncudedUntil from LibationContextModelSnapshot.cs.  Then I ran the dotnet ef migrations add AddIncludedUntil.  I am just inexperienced with EF.  Let me know if this works. TY.

delebash:
When I tried to revert the migration changes, it would not allow it as it said the changes where already applied to my database.

Mbucari:
@jwillikers I just tested your changes, and it successfully migrates existing databases.I think it's ready to go!

@rmcrackan ~I think you should merge this. One thing I noticed is that Date columns in the main grid are strings and so they don't sort properly. This was already true with the "Purchase Date" column, and @delebash made the "Included Until" column in the same way. I'd like to make date columns sortable, so maybe hold off on releasing this until I do that?~

Scratch that, "Purchase Date" actually _does_ already sort correctly. This is actually a simpler fix than I though. Give me a few minutes to get it together.

rmcrackan:
> Scratch that, "Purchase Date" actually does already sort correctly. This is actually a simpler fix than I though. Give me a few minutes to get it together.

Sounds good. There was that other ticket that was active until you made some test code to consolidate db connections (then the ticket went to radio silence). Is your db code from that something you'd like to include soon? Or is it too untested. I love the idea but it obviously is incredibly far reaching.

Source/LibationUiBase/GridView/GridEntry.cs
```diff
@@ -128,6 +132,7 @@ public void UpdateLibraryBook(LibraryBook libraryBook)
 		protected virtual DateTime GetPurchaseDate() => LibraryBook.DateAdded;
```
Mbucari: Below this line, add:
```c#
protected virtual DateTime? GetIncludedUntil() => LibraryBook.IncludedUntil;
```

Then go to SeriesEntry at the end of the file: https://github.com/rmcrackan/Libation/blob/2c882e883d49c01d7428437e1f7ea51ab5df2f07/Source/LibationUiBase/GridView/SeriesEntry.cs#L104-L105


Add this line to the end of the class:

```c#
protected override DateTime? GetIncludedUntil() => Children.Count == 0 ? default : Children.Min(c => c.LibraryBook.IncludedUntil);
```

Mbucari:
> Is your db code from that something you'd like to include soon? Or is it too untested. I love the idea but it obviously is incredibly far reaching.

I'm adding some documentation, cleaning it up a bit, and running more tests, but yes I think it'll be ready to ship with the next release today.

delebash:
@Mbucari Done

Mbucari:
@rmcrackan This is ready to merge.

rmcrackan:
Thanks to you both! I love that more people want to be involved and seeing the back and forth of everything coming together :)

Issue #1383, opened 2025-10-10
<series#[N]> tag no longer parsing correctly in Custom File Naming after upgrading to 12.5.4


manalive:
## Describe the bug  
To make sure I'm not just typing it incorrectly, I'm pasting from the documentation page: <series#[3]>

In the preview and the reality, it just replaces the series number with the digit entered (in this case "3").

## To Reproduce  

1. Go to Options > Download./Decrypt > Custom File Naming
2. Click on 'Edit' for the folder path
3. Use the <series#> tag with a number formatter
4. bemoan your fate

## Expected behavior  
Parsing should fit the documentation so <series#[3]> should result in "001" if the book is the first in a series intead of "3".

## Platform  
Windows 10

rmcrackan:
@Mbucari Do you have time to look at this and add a few lines above here: https://github.com/rmcrackan/Libation/blob/2c882e883d49c01d7428437e1f7ea51ab5df2f07/Source/_Tests/LibationFileManager.Tests/TemplatesTests.cs#L477

Mbucari:
@manalive We'll fix this soon, but in the meantime you can replace `<series#[3]>` with `<series#[000.#]>` to achieve the same effect.

rmcrackan:
Fixed in [version 12.5.6](https://github.com/rmcrackan/Libation/releases/latest)

Issue #1357, opened 2025-09-01
I cannot bring the program up, it just crashes, will not load


Merren0015:
PLEASE FILL OUT THE FOLLOWING. Bug reports with limited information or lacking an attached log file may get limited or delayed help.

___

[_Log202507.zip](https://github.com/user-attachments/files/22079345/_Log202507.zip)

[_Log202408.zip](https://github.com/user-attachments/files/22079348/_Log202408.zip)
[_Log202412.zip](https://github.com/user-attachments/files/22079347/_Log202412.zip)
[_Log202412_001.zip](https://github.com/user-attachments/files/22079344/_Log202412_001.zip)
[_Log202412_002.zip](https://github.com/user-attachments/files/22079346/_Log202412_002.zip)
[_Log202412_003.zip](https://github.com/user-attachments/files/22079349/_Log202412_003.zip)

[_Log202501.zip](https://github.com/user-attachments/files/22079350/_Log202501.zip)

**Describe the bug**  
A clear and concise description of what the bug is. Crashing

**To Reproduce**  

**Expected behavior**  

**Screenshots**  

**Platform**  
[e.g. Windows 10, Windows 11, Mac, Linux (State distribution)] Windows 11 running

**Log Files**  
Attach your Libation log file here. Logs are typically in your `[user]\Libation` folder. (For example, on windows: `C:\my_username\Libation`) Also within Libation, on the first tab in Settings you can click the button 'Open log folder'. If your user folder contains the file "LibationCrash.log", attach that also.

[_Log202507.zip](https://github.com/user-attachments/files/22079361/_Log202507.zip)

[_Log202408.zip](https://github.com/user-attachments/files/22079356/_Log202408.zip)
[_Log202412.zip](https://github.com/user-attachments/files/22079360/_Log202412.zip)
[_Log202412_001.zip](https://github.com/user-attachments/files/22079362/_Log202412_001.zip)
[_Log202412_002.zip](https://github.com/user-attachments/files/22079358/_Log202412_002.zip)
[_Log202412_003.zip](https://github.com/user-attachments/files/22079357/_Log202412_003.zip)

[_Log202501.zip](https://github.com/user-attachments/files/22079359/_Log202501.zip)

rmcrackan:
Thank you for the logs. Please download the [latest version](https://github.com/rmcrackan/Libation/releases/latest) and see if that helps. If that doesn't fix it, please upload the new log that it generates.

knotty68:
[...]

PR #1390, opened 2025-10-21
Bug Fixes and attempts to solve random crashes

Mbucari:
## [Fix possible race error on startup with autoscan](https://github.com/rmcrackan/Libation/commit/39e9f675d2000f414cb1a6e4579e81bc9835b2f8)

I don't remember what prompted this (a ticket or something I noticed).

## [Fix series order formatters to match documentation](https://github.com/rmcrackan/Libation/commit/a636c4bf7a866e05cc1b870841782bc1e4a0b397) (#1383)

Change the \<series#\> formatter to behave like the examples in the documentation.

## [Enforce sequential access to DbContext.](https://github.com/rmcrackan/Libation/commit/f733268b65760326e5b2e63d09c640fdea4e2945)

This is intended to help with #1357. Based on the crash log in [this comment](https://github.com/rmcrackan/Libation/issues/1357#issuecomment-3393345812), I'm thinking that Libation may have a cross-threading bug where the database is being accessed by more than one thread simultaneously. This change enforces sequential access to the database, so only one LibationContext instance may be active at a time.

Also, I'd noticed that `UpdateBook() wasn't being used and deleted it. That change got swept up in this commit, and I was too lazy to make it its own commit. That's OK, right? :-)

### DIFF
--- a/Source/ApplicationServices/DbContexts.cs
+++ b/Source/ApplicationServices/DbContexts.cs
@@ -9,7 +9,7 @@ namespace ApplicationServices
 	{
 		/// <summary>Use for fully functional context, incl. SaveChanges(). For query-only, use the other method</summary>
 		public static LibationContext GetContext()
-			=> LibationContext.Create(SqliteStorage.ConnectionString);
+			=> InstanceQueue<LibationContext>.WaitToCreateInstance(() => LibationContext.Create(SqliteStorage.ConnectionString));
 
 		/// <summary>Use for full library querying. No lazy loading</summary>
 		public static List<LibraryBook> GetLibrary_Flat_NoTracking(bool includeParents = false)
--- a/Source/DataLayer/InstanceQueue.cs
+++ b/Source/DataLayer/InstanceQueue.cs
@@ -1,0 +1,103 @@
+﻿using System;
+using System.Diagnostics;
+using System.Threading;
+
+#nullable enable
+namespace DataLayer;
+
+/// <summary> Notifies clients that the object is being disposed. </summary>
+public interface INotifyDisposed : IDisposable
+{
+	/// <summary> Event raised when the object is disposed. </summary>
+	event EventHandler? ObjectDisposed;	
+}
+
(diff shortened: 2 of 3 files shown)


rmcrackan:
> That's OK, right?

Yup. I kept it there as a just in case -- and that was *years* ago. tbh, I'm surprised it isn't used. It *feels* like something that would be used quite a bit. Oh yeah, I also kept it there as a proof of concept of that pattern. Actually, I do like it as a reference. I'll save that pattern somewhere else.

Ok, so as far as I can tell, your db access stuff looks fine. It also looks like a fair amount of complexity added to one of the most used areas of code, so it has a high pucker factor and I'm nervous. What's your confidence level? I know, that's a dumb question -- you're as invested as I am so you wouldn't have put it in if you weren't confident. I think I just need to hear it.

Mbucari:
> What's your confidence level?

That it'll fix anybody's problems? 80%.
That it won't deadlock or break Libation for anybody? 99%.

I've sat down with pen and paper and sketched out the different cases where one instance is active and a different thread calls `WaitToCreateInstance`. No matter where the first instance is in its process, the next call to `WaitToCreateInstance` will either wait and be triggered as intended, or skip waiting because it doesn't need to. I've also run simulations that create 100 tasks, each of which waits a random amount of time before trying to create an instance of an object, and the object waits a random amount of time before disposing of itself. I then awaited WhenAll on them, and all the simulations worked as expected. Each instance is created and disposed before the next one is created. No deadlock.

Mbucari:
I cleaned up the commit and restored `UpdateBook()`.

rmcrackan:
Fyi. Out all day at state fair :) Phone only, no PC. I can’t wait to see this tomorrow

Mbucari:
> Fyi. Out all day at state fair :)

Have fun!

> I can’t wait to see this tomorrow

It's pretty slick. Multi thread synchronization is always a mind-F (to me, at least), but this came together really well and I'm quite proud of it.

rmcrackan:
> I've sat down with pen and paper and sketched out the different cases...

I certainly can't ask for more than that. Let's do this!

PR #1402, opened 2025-10-27
Add support for postgres

twsouthwick:
Supporting postgres simplifies deployments to environments such as kubernetes. Since sqlite doesn't work well on nfs shares it can be easier for databases to have a dedicated db set up that applications can connect to. Sqlite is easier for most deployments though, so this will default to that if the settings haven't been updated to support it.

This change does the following:

- Separate out SQLite from the DataLayer and adds a Postgres assembly for migrations as well
- Add a configuration setting for a postgres connection string that will be used if it is there, otherwise reverts to the original sqlite string
- Add a copydb command for the cli to bootstrap the postgres db
- A convenience script to update migrations for both dbs at the same time

### DIFF
--- a/Source/ApplicationServices/DbContexts.cs
+++ b/Source/ApplicationServices/DbContexts.cs
@@ -11,21 +11,15 @@ namespace ApplicationServices
         public static LibationContext GetContext()
             => InstanceQueue<LibationContext>.WaitToCreateInstance(() =>
             {
-                var context = !string.IsNullOrEmpty(Configuration.Instance.PostgresqlConnectionString)
+                return !string.IsNullOrEmpty(Configuration.Instance.PostgresqlConnectionString)
                     ? LibationContextFactory.CreatePostgres(Configuration.Instance.PostgresqlConnectionString)
                     : LibationContextFactory.CreateSqlite(SqliteStorage.ConnectionString);
-
-                context.Database.Migrate();
-
-                return context;
             });
 
         /// <summary>Use for full library querying. No lazy loading</summary>
         public static List<LibraryBook> GetLibrary_Flat_NoTracking(bool includeParents = false)
         {
             using var context = GetContext();
-
-            context.Database.Migrate();
 
             return context.GetLibrary_Flat_NoTracking(includeParents);
         }


rmcrackan:
Thank you so much for this PR. Initial thoughts: I like this and it looks well thought out. It's also ambitious so this may take a while to review.

My biggest concern is: `GetContext()` and `GetLibrary_Flat_NoTracking(bool includeParents = false)` call `context.Database.Migrate();` everytime. This doesn't feel right.

Lessor concerns:

* Where is LIBATION_CONNECTION_STRING coming from?
* `migrate.json` was deleted. This was the sqlite constr. Where is it now?
* Why was `_DB_NOTES.txt` removed? These are notes to myself and to future dev'ers. You're free to improve them though.
* It feels like `dotnet-tools.json` should be git-ignored since it's only for your dev env
* Libation is assumed to work on all major OSs unless in an area specified otherwise. Is `add-migrations.ps1` likely to run on mac and linux? I get that it's a convenience script so it's not crucial. I'm curious though.

@Mbucari 

I'd love your opinions in general. Specifically though, `LibationContextFactory.cs` makes changes to your recent fixes. Can you verify that it's in the spirit of those fixes. It appears to be but you'd know better. I'd also like your opinion on the `context.Database.Migrate();` calls.

Mbucari:
@rmcrackan The `InstanceQueue<T>` part looks fine to me. I have no clue about anything else in the PR.

rmcrackan:
@twsouthwick I'd like to complete your PR. There's a lot of great stuff in here. After you address my questions above, I can merge and release a new version for you w/postgres.

twsouthwick:
Hi @rmcrackan - been busy so just saw your questions:

>Where is LIBATION_CONNECTION_STRING coming from?

It's an environment variable (mostly for use in containered environments:

https://github.com/twsouthwick/Libation/blob/1b5db9b28fd43b5b4ae1ccfbbaea460be7d72b50/Source/LibationFileManager/Configuration.ConnectionStrings.cs#L12

> migrate.json was deleted. This was the sqlite constr. Where is it now?

It wasn't actually used anywhere that I could see - the sqlite constr was a const string here: https://github.com/rmcrackan/Libation/blob/f4e7cf341893887e3dfae92ad39639a2dae65866/Source/LibationFileManager/SqliteStorage.cs#L9

it appeared therefore to just be an artifact of some previous version and cleaned it up. If I missed it, happy to add it back

> Why was _DB_NOTES.txt removed? These are notes to myself and to future dev'ers. You're free to improve them though.

I can update them to reflect the current state, but TBH it wasn't correct and wasn't helpful when I tried creating migrations, so I figured it was out of date. If you'd prefer the file with directions to run the script, I'll update it

> It feels like dotnet-tools.json should be git-ignored since it's only for your dev env

it's for all devs, though. That way the ef tools (with the correct version) will be installed to the CLI as part of a restore and the add-migrations.ps1 script will work with no extra effort.

> Libation is assumed to work on all major OSs unless in an area specified otherwise. Is add-migrations.ps1 likely to run on mac and linux? I get that it's a convenience script so it's not crucial. I'm curious though.

Yes, powershell works on linux/mac: https://learn.microsoft.com/en-us/powershell/scripting/install/installing-powershell?view=powershell-7.5

twsouthwick:
> I'd also like your opinion on the context.Database.Migrate(); calls.

I could have sworn the migrate was already there, not sure why I would have added that. I've removed them

rmcrackan:
Thanks for the clarifications. I'll merge today and do a release after some local testing.

rmcrackan:
@twsouthwick

I found why context.Database.Migrate() was there. DesignTimeDbContextFactoryBase used to do this for us. I re-added the Migrate.

> I can update them to reflect the current state, but TBH it wasn't correct and wasn't helpful when I tried creating migrations, so I figured it was out of date. If you'd prefer the file with directions to run the script, I'll update it

Yes, please. I did some trivial reverts until I can investigate later, including adding back `_DB_NOTES.txt` with a disclaimer that the instructions may be out of date. I'd rather have a starting point that I can edit later rather than not have that file at all. If you have improvements, that would be great.

@ducamagnifico , @wtanksleyjr , @CLHatch -- hey docker folks, you'll probably really like this latest update: optional postgres in [version 12.6](https://github.com/rmcrackan/Libation/releases/latest). If you have suggestions on how to improve the [docker docs](https://github.com/rmcrackan/Libation/blob/master/Documentation/Docker.md) I'd love a good write up. Notable changes: [CLI has a new copydb](https://github.com/rmcrackan/Libation/blob/master/Documentation/Advanced.md) , support for env var `LIBATION_CONNECTION_STRING`

PR #1439, opened 2025-11-20
Update to .NET 10

Mbucari:
- Update all project runtime targets
- Update all dependencies
  - NOTE: Using Npgsql.EntityFrameworkCore.PostgreSQL RTM build from MyGet
- Delete unused pubxml files (they were made redundant by recent workflow changes)
- Replace Libation.sln with Libation.slnx

### DIFF
--- a/Source/LibationWinForms/Program.cs
+++ b/Source/LibationWinForms/Program.cs
@@ -21,7 +21,7 @@ namespace LibationWinForms
 		private static Form1 form1;
 
 		[STAThread]
-		static async Task Main()
+		static void Main()
 		{
 			Task<List<LibraryBook>> libraryLoadTask;
 
@@ -46,7 +46,7 @@ namespace LibationWinForms
 				LibationUiBase.Forms.MessageBoxBase.ShowAsyncImpl = ShowMessageBox;
 
 				// do this as soon as possible (post-config)
-				await RunSetupIfNeededAsync(config);
+				RunSetupIfNeededAsync(config).Wait();
 
 				// most migrations go in here
 				LibationScaffolding.RunPostConfigMigrations(config);


Mbucari:
## [Improve import performance on large batches](https://github.com/rmcrackan/Libation/pull/1439/commits/f9ac0253fb6bf3a151327ed275538df315d625ad)

When importing a large batch (my test case was about ~6600), the query was taking over 20 seconds to run. Whereas fetching the entire library was under a second.

## [Refactor DbContext access and disposal](https://github.com/rmcrackan/Libation/pull/1439/commits/a55da5f1879ed8544211ee0c907584ba90e7b945)

- Remove instance queue. This is a database, after all, and is designed to be accessed and written to concurrently
- Reduce the number of calls to DbContexts.Create()
- Ensure that no LibationContext remains open across an await boundary. Multithread context access is the most likely culprit for past issues.
- Make all Update UserDefinedItem methods asynchronous.

## [Revert hack workaround for DataGridView error](https://github.com/rmcrackan/Libation/pull/1439/commits/913019cdfdf95ce421157b2d648b33e6140034c0)

This is your note in the .NET 10 update thread.

rmcrackan:
> Replace Libation.sln with Libation.slnx

I haven't played with slnx yet. Is there a native tool for this conversion from sln?

You've been deeper in the builds lately. Would you recommend accepting or rejecting these new dependabot bumps #1441 , #1442 ?

Mbucari:
> I haven't played with slnx yet. Is there a native tool for this conversion from sln?

`dotnet sln migrate`

> Would you recommend accepting or rejecting these new dependabot bumps #1441 , #1442 ?

I just tested the macos update and it works well. Yes, go ahead and bump.

rmcrackan:
Bumped. I'd still like to not do a release while I'm out of town tho

Mbucari:
@rmcrackan We should add an editorconfig to the repo, then do a solution-wide code cleanup to get eerything on the same page and keep it that way. A lot of the earliest code uses spaces, but most all new code uses tabs. Inconsistent bracket use on single-line if/for/where blocks. Also a lot of naming violations in early code (lotta lowercase method and type names). The problem is that without a standard, it makes code review more difficult because so many lines with, for example, only white space changes are showing up as changed lines. Not having a standard also prevents using the code cleanup tool on older files because doing so would create a lot of git changes that aren't really changes. But if we define the style and do all cleanup in a single commit, we can continue using the tool in the future without affecting more lines than were actually changed.

rmcrackan:
I'm not crazy about enforced styles. I don't find it that hard to read. And to be perfectly honest and petty: it feels too much like being at work, having opinionated tools criticizing my personal code.

RE whitespace only changes, if they get excessive, I turn this on:

whitespace

Mbucari:
Fair enough! Thanks for the tip about hiding whitespace. I seem to remember having see or used it, but I haven't in a while. I need to start as a matter of course.

Issue #1417, opened 2025-11-10
Library appears to be empty after new start up


Skinopolus:
I opened the Programm today and instead of listing all the books in my library it was basically a blank window and even after the scanning indicator disappeard there were no books listed. I can still see all files on my hard drive and even open the respective folder through "locate Audiobooks" but they do not show up in the program.

Added the latest log file here:
[Log202511_002.log](https://github.com/user-attachments/files/23453185/Log202511_002.log)

rmcrackan:
That's really weird. Is there anything in the search/filter bar?

Skinopolus:
> That's really weird. Is there anything in the search/filter bar?

No Filters or anything. Just complete blank in all fields :(

<img width="1920" height="1020" alt="Image" src="https://github.com/user-attachments/assets/50473d59-6928-498b-b185-751a46b3c7d7" />

Mbucari:
@Skinopolus It may be related [to this bug](https://github.com/rmcrackan/Libation/issues/1413#issuecomment-3505298574). That user was not seeing new books. Deleting everything and resetting Libation appeared to solve their problem.

rmcrackan:
@Mbucari I feel like you're right. But this user's LibationFiles folder suggests they've been using it since 11.1 -- quite a long time. Deleting their database could get rid of a lot of book download info they don't want to lose.

Mbucari:
@Skinopolus If you'd like, you can email me your LibationContext.db file and I'll try to figure out why your books aren't displaying.

Skinopolus:
> [@Mbucari](https://github.com/Mbucari) I feel like you're right. But this user's LibationFiles folder suggests they've been using it since 11.1 -- quite a long time. Deleting their database could get rid of a lot of book download info they don't want to lose.

That's true. Especially with the plus cataloug changing the access so frequently

Mbucari:
@Skinopolus email me at mbucari1@gmail.com

Skinopolus:
> [@Skinopolus](https://github.com/Skinopolus) email me at [mbucari1@gmail.com](mailto:mbucari1@gmail.com)

I looked for the  LibationContext.db  file, but have multiple ones.

One DB-SHM and one DB-WAL as well as one database file in the general Libation folder and one Database file one in "LibationFiles" (the same folder as in which I have all the Audio Book files). Which one should I share?
also reached out via mail. Thanks for the offer!

Mbucari:
[...]

Issue #1365, opened 2025-09-11
Crashing Liabtion


Merren0015:
PLEASE FILL OUT THE FOLLOWING. Bug reports with limited information or lacking an attached log file may get limited or delayed help.

___

## Describe the bug  
Trying to load my library the program crashes and go back restart to have it crash again.  I changed the program to my primary C drive and took out my external drive, while keeping my liberated books in my external drive;  but the problem persists.  Below are my logs.  
## Platform  
[e.g. Windows 10, Windows 11, Mac, Linux (State distribution)]

## Log Files
Attach your Libation log file here. If your user folder contains the file "LibationCrash.log", attach that also.

**Default Log File Locations**
|Platform|Folder|
|-|-|
|Windows|`%userprofile%\Libation`|
|macOS|`~/Library/Application Support/Libation`|
|Linux|`~/.local/share/Libation`|

Alternative, you may open the log file folder from within Libation. Open Libation's settings, and on the first tab in Settings you can click the button 'Open log folder'. 

[_Log202408.zip](https://github.com/user-attachments/files/22265547/_Log202408.zip)
[_Log202412.zip](https://github.com/user-attachments/files/22265543/_Log202412.zip)
[_Log202412_001.zip](https://github.com/user-attachments/files/22265545/_Log202412_001.zip)
[_Log202412_002.zip](https://github.com/user-attachments/files/22265544/_Log202412_002.zip)
[_Log202412_003.zip](https://github.com/user-attachments/files/22265549/_Log202412_003.zip)

[_Log202501.zip](https://github.com/user-attachments/files/22265548/_Log202501.zip)

[_Log202507.zip](https://github.com/user-attachments/files/22265546/_Log202507.zip)

Merren0015:
![Image](https://github.com/user-attachments/assets/3f9a68de-fd26-4d72-b40c-6d26db0df3d9)

![Image](https://github.com/user-attachments/assets/293a2ad0-1eee-4db9-a1b4-2ec6dabe0b14)

rmcrackan:
Thank you for your logs and screenshots. Can you attach the `.log` files which start with `Log202509` (no underscore)?

Merren0015:
[Log202509_002.log](https://github.com/user-attachments/files/22322747/Log202509_002.log)
[Log202509.log](https://github.com/user-attachments/files/22322748/Log202509.log)
[Log202509_001.log](https://github.com/user-attachments/files/22322746/Log202509_001.log)

[...]

Issue #1338, opened 2025-08-17
Scan Isn't Finding Library


bradleyesqueda-lgtm:
Using MacBook M1. I opened the program like normal, did what I had to do, and closed it. Went back immediately after and suddenly, my library is gone and it just says scanning forever and then nothing happens. Did an uninstall and reinstall, nothing. Logged in and out multiple times, nothing. Please help!

**Describe the bug**  

**To Reproduce**  

**Expected behavior**  

**Screenshots**  

**Platform**  
[e.g. Windows 10, Windows 11, Mac, Linux (State distribution)]

**Log Files**  
Attach your Libation log file here. Logs are typically in your `[user]\Libation` folder. (For example, on windows: `C:\my_username\Libation`) Also within Libation, on the first tab in Settings you can click the button 'Open log folder'. If your user folder contains the file "LibationCrash.log", attach that also.

Mbucari:
If you go into your settings folder (`~/Library/Application Support/Libation`), I suspect there are two files named `LibationContext.db-wal` and `LibationContext.db-shm`. Delete those and restart Libation, and it should work. If it doesn't, attach your log files. Libation > Settings > Open Log Folder button

bradleyesqueda-lgtm:
That worked. Thank you so much. Apologies for not including the log files.

Mbucari:
@bradleyesqueda-lgtm thank you for letting us know. I'm sorry for the way I spoke last night.

PR #1456, opened 2025-11-25
Fix database lock from -wal and -whm files

Mbucari:
Delete LibationContext.db-shm and LibationContext.db-wal files as part of startup routine.

This addresses issues #1338, #1417, and #1365.


rmcrackan:
Merge now?

Mbucari:
Yup. This is a one-off to fix an ongoing issue. I don't think it's worth releasing now, but I'll let you make that call.

rmcrackan:
I agree. Especially after back to back releases.

PR #1458, opened 2025-11-25
Bug fixes in the downloader

Mbucari:
I noticed two errors in those logs that actually are bugs.

### DIFF
--- a/Source/ApplicationServices/LibraryCommands.cs
+++ b/Source/ApplicationServices/LibraryCommands.cs
@@ -586,7 +586,7 @@ namespace ApplicationServices
 
         // below are queries, not commands. maybe I should make a LibraryQueries. except there's already one of those...
 
-        public record LibraryStats(int booksFullyBackedUp, int booksDownloadedOnly, int booksNoProgress, int booksError, int booksUnavailable, int pdfsDownloaded, int pdfsNotDownloaded, int pdfsUnavailable)
+        public record LibraryStats(int booksFullyBackedUp, int booksDownloadedOnly, int booksNoProgress, int booksError, int booksUnavailable, int pdfsDownloaded, int pdfsNotDownloaded, int pdfsUnavailable, IEnumerable<LibraryBook> LibraryBooks)
         {
             public int PendingBooks => booksNoProgress + booksDownloadedOnly;
             public bool HasPendingBooks => PendingBooks > 0;
@@ -655,7 +655,7 @@ namespace ApplicationServices
 
 			Log.Logger.Information("PDF counts. {@DebugInfo}", new { total = pdfResults.Count, pdfsDownloaded, pdfsNotDownloaded, pdfsUnavailable });
 
-            return new(booksFullyBackedUp, booksDownloadedOnly, booksNoProgress, booksError, booksUnavailable, pdfsDownloaded, pdfsNotDownloaded, pdfsUnavailable);
+            return new(booksFullyBackedUp, booksDownloadedOnly, booksNoProgress, booksError, booksUnavailable, pdfsDownloaded, pdfsNotDownloaded, pdfsUnavailable, libraryBooks);
         }
     }
 }
--- a/Source/LibationAvalonia/ViewModels/MainVM.BackupCounts.cs
+++ b/Source/LibationAvalonia/ViewModels/MainVM.BackupCounts.cs
@@ -1,5 +1,4 @@
 ﻿using ApplicationServices;
-using Avalonia.Threading;
 using DataLayer;
 using LibationFileManager;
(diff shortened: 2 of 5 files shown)


rmcrackan:
I'm getting lost with some directory create logic. Good news: I can repro this on my box:

There is a downloaded book with a green stoplight because I wanted to open the book's directory. I click on the stoplight and this message comes up:

```
Failed to Create Downloads In Progress Directory
---------------------------
Libation was unable to create the "Downloads In Progress" folder in:
C:\Users\rmcrackan\AppData\Local\Temp\Libation

Please change the In Progress location in the settings menu.
```

After I close this dialog, the folder is correctly opened.

These folders already exist and I've been using them since forever:

* C:\Users\Dinah\AppData\Local\Temp\Libation\DecryptInProgress
* C:\Users\Dinah\AppData\Local\Temp\Libation\DownloadsInProgress

I also tried right-clicking a book to re-download. Same error.

I deleted my log then repro'd both errors to get this streamlined log:

[Log202511.log](https://github.com/user-attachments/files/23755611/Log202511.log)

Mbucari:
Interesting @rmcrackan. I wonder if this is a dotnet bug. This is how we've been doing in-progress directories for a long time. Here's the code in v12.0.0:
https://github.com/rmcrackan/Libation/blob/e3b0f80016382fe2a934aeadd94ff17b50463f78/Source/LibationFileManager/AudibleFileStorage.cs#L22-L23

The new code is only slightly different:
```c#
new DirectoryInfo(Configuration.Instance.InProgress).CreateSubdirectoryEx(subDirectory)
```
 [CreateSubdirectoryEx](https://github.com/rmcrackan/Libation/blob/2bf6f7a4f2067c4936e442997ca88cfd0da42053/Source/FileManager/FileUtility.cs#L275) is workaround to a bug in CreateSubdirectory that can't create a subdirectory on a drive root.

I wonder, can you change that above line (AudibleFileStorage.cs#36) to the below and see if you still get the error.
```C#
Directory.CreateDirectory(Path.Combine(Configuration.Instance.InProgress, subDirectory));
```

rmcrackan:
No change

Screenshot 2025-11-25 16 29 42

Mbucari:
Hmmmm. Unfortunately, I can't reproduce this error. Can you try turnig off "Just My Code" and stepping into CreateDirectory? You might see something useful.

rmcrackan:
The 'just my code' never seems to make a notable difference for me.

Found it. What a weird disconnect. My last computer used this, which is in my `\LibationFiles\Settings.json`:  
`"InProgress": "C:\\Users\\rmcrackan\\AppData\\Local\\Temp\\Libation",`

My new laptop windows user is `Dinah`. I'm using an old appsettings which has the above value in the json. But when you go to GUI Edit Settings, it shows this for 'In progress files': `C:\Users\Dinah\AppData\Local\Temp\Libation`

Somehow the GUI knows to show the curr user's folder, but it's using the value in the app settings, which points to a bad user. To repro, go into your Settings.json and change just the user part of the path. Then open the GUI settings and see what it says for 'in progress files'

rmcrackan:
Merged. Another release needed?

Mbucari:
Thank goodness! I was really worried because I had no clue what was going on. 

> Somehow the GUI knows to show the curr user's folder, but it's using the value in the app settings, which points to a bad user.

That makes sense. That directory select control takes a raw directory path string as it's input, and then tries to match a known directory to that path.

> Another release needed?

Yes. I'm not going to have much time over the next few days, but the good news is that everyone else is probably going to be busy too. What was that old quote you had about not pushing releases the week of holidays?

Mbucari:
> The 'just my code' never seems to make a notable difference for me.

When disabled, you can step into third-party library code. If source link is available, you get to step through the original source code. If not, visual studio decompiles it and you get to step through the decompiled code. And you get to read all of the in-scope variables.

rmcrackan:
> What was that old quote you had about not pushing releases the week of holidays?

just don't do it

If we're both about to be busy -- which makes total sense esp for this time of year -- then I'll set a calendar reminder to do a release next week.

rmcrackan:
> That makes sense. That directory select control takes a raw directory path string as it's input, and then tries to match a known directory to that path.

This is definitely a bug. The GUI value and backend value should never be out of step. It's just really low priority. I'll create an issue for ... whenever.

Mbucari:
> then I'll set a calendar reminder to do a release next week.

I think you should release this one now. If something acute blows up, I can definitely do something about it tomorrow. But there's not much in this release except fixes for known bugs. And I really think we should get it out.

rmcrackan:
Released

Issue #1478, opened 2025-12-03
Error and app crash on update


bdelbanco:
PLEASE FILL OUT THE FOLLOWING. Bug reports with limited information or lacking an attached log file may get limited or delayed help.

___

## Describe the bug  
I went through the usual libation update method and got a hard error which does not allow me to reopen libation

## To Reproduce  

On attempting to open libation

## Expected behavior  
A clean update and restart of Libation

## Platform  
Windows 11

## Log Files
Attach your Libation log file here. If your user folder contains the file "LibationCrash.log", attach that also.

**Default Log File Locations**
|Platform|Folder|
|-|-|
|Windows|`%userprofile%\Libation`|
|macOS|`~/Library/Application Support/Libation`|
|Linux|`~/.local/share/Libation`|

Alternative, you may open the log file folder from within Libation. Open Libation's settings, and on the first tab in Settings you can click the button 'Open log folder'. 

[Log202512.log](https://github.com/user-attachments/files/23917358/Log202512.log)

rmcrackan:
So sorry about that. Hopefully fixed in the new [Libation version 12.8.1](https://github.com/rmcrackan/Libation/releases/latest)

Issue #1477, opened 2025-12-03
Save button does not close dialog


rmcrackan:
Libation Classic

Clicking book detail dialog's "Save" button should close the dialog

Repro

* Click a book detail button. ie: a button in the "Tags and Details" column
* Dialog opens
* Make a change
* Click save
* Saving succeeds. Dialog remains open

rmcrackan:
Partly fixed in the new [Libation version 12.8.1](https://github.com/rmcrackan/Libation/releases/latest); partly not a bug, just a different expected action.

PR #1479, opened 2025-12-04
Two minor bug fixes

Mbucari:
[Fix books dialog not saving or updating properly](https://github.com/rmcrackan/Libation/commit/ea3716f48a0c43a7fefef81b173ff2b7f595bd29) (#1477)

[Don't throw exceptions from failure to delete db-wal and db-shm files](https://github.com/rmcrackan/Libation/commit/986dbd678f97c4fab54af4b7ccbeea3e71423e07) (#1478)

### DIFF
--- a/Source/AppScaffolding/LibationScaffolding.cs
+++ b/Source/AppScaffolding/LibationScaffolding.cs
@@ -105,15 +105,35 @@ namespace AppScaffolding
 
 		/// <summary>
 		/// Delete shared memory and write-ahead log SQLite database files which may prevent access to the database.
+		/// These file may or may not cause libation to hang on CreateContext,
+		/// so try our luck by swallowing any exceptions and continuing.
 		/// </summary>
 		private static void DeleteOpenSqliteFiles(Configuration config)
 		{
 			var walFile = SqliteStorage.DatabasePath + "-wal";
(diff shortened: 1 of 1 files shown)


rmcrackan:
1 question, 1 concern:

Question: I don't see anything about closing the dialog and you're removing the assignment to DialogResult. I assume this goes back to a recent PR item about dialogs and a new paradigm. What's the new way?

Concern: with swallowing the -wal/-shm exceptions, do we risk getting stuck in a tight error loop which would have previously been broken by the exception?

Mbucari:
> What's the new way?

It's no longer a dialogue which needs a result. You can keep the details window open and click on new books, and they will show up in the books dialog. I've found it useful for quickly looking at some details of multiple books. When you click the save button, it saves book status, PDF status, and tags, but it doesn't close. You close the window with the x button. You can change a book, save, click a new book, change, save, etc. 

> do we risk getting stuck in a tight error loop

No, DeleteOpenSqliteFiles is only called once during startup [from here](https://github.com/rmcrackan/Libation/blob/426d5a87b4376f57f400bce934a93df9f0bd6e61/Source/AppScaffolding/LibationScaffolding.cs#L93).

Issue #1487, opened 2025-12-06
Import-Scan Library pops up an error


sticklizard:
PLEASE FILL OUT THE FOLLOWING. Bug reports with limited information or lacking an attached log file may get limited or delayed help.

___

## Describe the bug  
Import-Scan Library pops up an error
## To Reproduce  

1. Click on Import
2. Click on Scan Library
3 See error

## Expected behavior  
I expect to see any new files (I think there's one)

## Screenshots  

<img width="602" height="482" alt="Image" src="https://github.com/user-attachments/assets/4b48e500-708b-4047-87f9-a1ead50e6dd5" />

## Platform  
Windows 11

## Log Files
Attach your Libation log file here. If your user folder contains the file "LibationCrash.log", attach that also.

[Log202512_001.log](https://github.com/user-attachments/files/23998416/Log202512_001.log)

**Default Log File Locations**
|Platform|Folder|
|-|-|
|Windows|`%userprofile%\Libation`|
|macOS|`~/Library/Application Support/Libation`|
|Linux|`~/.local/share/Libation`|

Alternative, you may open the log file folder from within Libation. Open Libation's settings, and on the first tab in Settings you can click the button 'Open log folder'.

Mbucari:
@rmcrackan There are two errors here. One is an error in the error display of an error message (ugh). I've fixed that. The next error is strange. Basically, the full database list contains two entries with the same ASIN. That shouldn't happen, and I'm not sure how it could. We take some great pains to _not_ do that.

@sticklizard Please lunach hangover and run this SQL command:

```Sql
Select AudibleProductId From Books Group By AudibleProductId Having Count(*) > 1
```

sticklizard:
Additional info:  When I reran the scan, it got the same error, then the program closed.  When I restarted it, the book was in the list and I was able to liberate it, so I don't know if the SQL result will help you, but here it is anyway:

<details>
  <summary>SQL command output</summary>
  

1984886878	
B01JKBSOVG	
B08JJCTDJT	
B08JJDLDQ1	
B08JJDPSS9	
B08JJDPZC3	
B08JJDS9HJ	
B08JJF1KWD	
B08JJF1RX7	
B08JJF2YHD	
B08JJF31NN	
B08JJF3YLX	
B08JJF4GLQ	
B08JJF8GFH	
B08JJF9QY4	
B08JJFCWJ4	
B08JJFCYJR	
B08JJFP82F	
B08JJFQ4PT	
[...]

Issue #1483, opened 2025-12-05
odd naming effect, recently noticed.


AprilShowersAreNice:
PLEASE FILL OUT THE FOLLOWING. Bug reports with limited information or lacking an attached log file may get limited or delayed help.

___

## Describe the bug  

I have recently started playing with the custom file naming, etc.  I had a number of user issues which I have sorted ...  The one I can not find is related to the name of the files saved.  I do not know when this started happening, I am just beginning to re-download my library since the option to ignore Audible Plus books was introduced (/noticed).

The issue is ... (My download today (20251205 ... tried a few times)
Lensman-02-First Lensman_E E Doc Smithaaxc.aaxc
Lensman-02-First Lensman_E E Doc Smithcue.cue
Lensman-02-First Lensman_E E Doc Smithkey.key
Lensman-02-First Lensman_E E Doc Smithm4b.m4b

The extension seems to have been added before the extension separator:
SmithAAXC.AAXC

I expected Smith.AAXC

Capitals were for emphasis only.

It did not do this with two of the files, shown below: 

Lensman-02-First Lensman_E E Doc Smith.jpg
Lensman-02-First Lensman_E E Doc Smith.metadata.json

## To Reproduce  
Just download another "book" using the program.

## Platform  
I am still on Windows 7

Mbucari:
@AprilShowersAreNice I'm not seeing this behavior with the default naming templates on Windows. Can you please post them template you're using?

AprilShowersAreNice:
Hello, Mbucari, thank yyou for getting back to me so quickly.

My Custom naming for Folders :: 
Author\<first author>\<if series-><first series>\<first series>-<series#[00]>-<title short><-if series><!if series->ZZZ_Various<-if series>

And For Files ::
<if series-><first series>-<series#[00]>-<-if series><title short>_<first author>

I did not change the chapter one.

Hopefully this is the information you are requesting.

Thank you.

AprilShowersAreNice:
Apologies I just noticed that the ... And For Files ::: did not copy properly ... 

It should have been ::

```
<if series-><first series>-<series#[00]>-<-if series><title short>_<first author>
```

Weird the display is the same from :: -serie .... to ...  ort>_

Hopefully it is just a weird display thing.


And just in case, the one for folders...

```
Author\<first author>\<if series-><first series>\<first series>-<series#[00]>-<title short><-if series><!if series->ZZZ_Various<-if series>
```

[...]

PR #1490, opened 2025-12-09
Two bugfixes

Mbucari:
I'll let you decide if this is worth releasing.

## [Fix file utility modifying file extension using replacement character](https://github.com/rmcrackan/Libation/commit/a7e9479eab3f0611a80c6c47e320f6f98824a69c)

The file extension should not be subject to renaming. #1483

## [Invoke MessageBox on UI thread](https://github.com/rmcrackan/Libation/commit/9545b3a874ab2617e1a5c84300d8bb18030adf81)

Fixes the error displaying the error message in #1487. This does not address whatever caused duplicated books to be entered into the database.


rmcrackan:
> I'll let you decide if this is worth releasing

Nah. Feels really minor. Let's merge now and release with the next batch.

PR #1504, opened 2025-12-21
Migrate from VSTest to MTP

Youssef1313:
Microsoft.Testing.Platform is the newer alternative to VSTest. This PR moves from VSTest to MTP.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,5 @@
+<Project>
+  <PropertyGroup>
+    <EnableMSTestRunner>true</EnableMSTestRunner>
+  </PropertyGroup>
+</Project>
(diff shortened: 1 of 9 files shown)


rmcrackan:
@Mbucari I feel like you've kept up with the latest improvements better than I have. After a quick google, I have no issue switching from VSTest to the newer Microsoft.Testing.Platform, but I have no personal experience with it. Do you have any objections to me merging this?

rmcrackan:
@Youssef1313 Thank you for your PR.

I had fun poking around your chess library. I did one of these years ago and it's always interesting to see the different ways people go about it. Such a well-known set of problems with so many different ways to approach it.

Mbucari:
@rmcrackan No objection here. I just approved the workflows to run and all existing tests pass, so I guess that's good enough for me.

Issue #1513, opened 2025-12-25
MacOS: No user-visible version info for Libation.app


tadwohlrapp:
I noticed that the Libation.app bundle doesn't show its version number to the user:

<img width="2064" height="1088" alt="Image" src="https://github.com/user-attachments/assets/1f648b60-d95b-4635-a268-684242a11fb1" />

This little issue seems to be caused by a missing `CFBundleShortVersionString` key in [`Info.plist`](https://github.com/rmcrackan/Libation/blob/4f44c26b57564718ad53b26f9fe611181cc74e5c/Source/LoadByOS/MacOSConfigApp/Info.plist).
https://github.com/rmcrackan/Libation/blob/4f44c26b57564718ad53b26f9fe611181cc74e5c/Source/LoadByOS/MacOSConfigApp/Info.plist#L19-L20

The existing [`CFBundleVersion`](https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundleversion) key is used internally by the system, while [`CFBundleShortVersionString`](https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundleshortversionstring) is the user-visible one.

The super simple fix:

```xml
	...
	<key>CFBundleVersion</key>
	<string>VERSION_STRING</string>
	<key>CFBundleShortVersionString</key>
	<string>VERSION_STRING</string>
	...
```

rmcrackan:
@Mbucari Can you confirm that tadwohlrapp's fix works on your mac?

Mbucari:
@rmcrackan Yup, that works.

rmcrackan:
@tadwohlrapp Thanks for the heads up. Fixed in code and will go out with the next release.

PR #1515, opened 2025-12-26
Add CFBundleShortVersionString to Info.plist

Mbucari:
Addresses #1513


rmcrackan:
Merged. I'll wait to do a release later when there's more substance.

PR #1517, opened 2025-12-27
Fix mp3 conversion of liberated AC-4 files and add metadata tags

Mbucari:
I had a chat with some Audiobookshelf people, and I think AUDIBLE_ACR, AUDIBLE_DRM_TYPE, and AUDIBLE_LOCALE tags could be useful.

### DIFF
--- a/Source/AaxDecrypter/MpegUtil.cs
+++ b/Source/AaxDecrypter/MpegUtil.cs
@@ -2,7 +2,9 @@
 using AAXClean.Codecs;
 using NAudio.Lame;
 using System;
+using System.Linq;
 
+#nullable enable
 namespace AaxDecrypter
 {
 	public static class MpegUtil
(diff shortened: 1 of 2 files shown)


rmcrackan:
I still haven't gotten around to personally using audiobookshelf yet. Have you?

Mbucari:
I installed it to poke around a little, but I just don't get the use case. I'm home every evening, so if I need a new audiobook I load it up. I then listen to it in its entirety before starting another. I don't need to access my whole library from everywhere.

rmcrackan:
Ditto. I bet that although some people use it on the road, I have a suspicion that others use it share their libraries with friends/family.

CLHatch:
> Ditto. I bet that although some people use it on the road, I have a suspicion that others use it share their libraries with friends/family.

I personally use it instead of the Audible app, mostly on my phone. I also have audiobooks from Humble Bundle and other sources in my library, not just Audible. I've also cast out to Google speakers in a bedroom though (when our blind cat was a kitten, we found that having a book read in the room she was in would calm her down while I was away at work).

ABS can also download podcasts (real podcasts with RSS feeds, not what Audible refers to as podcasts), and that has been nice to have, a list of podcasts we listen to in one place, streaming from a local server instead of the Internet.

wtanksleyjr:
Me too - that is, I use it instead of Audible so all downloading is just done using my phone.

BTW: Libation downloads AC-4, I move it to audiobookshelf (although I've replaced the ffmpeg exes with custom-built), I click the Chromecast button in the audiobookshelf client, and boom, I'm listening to AC4 Dolby Atmos sound on my Atmos speakers. The Quiddich match in the first Harry Potter book is a good test case. It's nice when a complex setup works!

rmcrackan:
I stand corrected. Those sound like great use cases.

CLHatch:
> Me too - that is, I use it instead of Audible so all downloading is just done using my phone.
> 
> BTW: Libation downloads AC-4, I move it to audiobookshelf (although I've replaced the ffmpeg exes with custom-built), I click the Chromecast button in the audiobookshelf client, and boom, I'm listening to AC4 Dolby Atmos sound on my Atmos speakers. The Quiddich match in the first Harry Potter book is a good test case. It's nice when a complex setup works!

I wasn't sure about the device compatibility of AC4, so I opted for EAC3.

CLHatch:
> I stand corrected. Those sound like great use cases.

One option the phone apps for ABS has is to download any books you want from your server so it can play them without service, too. I'll sometimes do that for the books I plan to listen to next, just so it doesn't have to stream them. 

Oh, and being a server, it tries to keep track of where you are in the book, so you can pick up from the same spot later on another device, say your PC. There are issues with that that they've been trying to track down though.

wtanksleyjr:
Oh, sorry, yes; E-AC-3 is what I meant. DUDE it's so hard to keep those things straight. E-AC-3 is compatible with almost any modern system and degrades smoothly.

PR #1535, opened 2026-01-06
Document Pacstall for program installation on Linux

niontrix:
Added information about Pacstall for AUR-like program installation on Ubuntu/Debian.


rmcrackan:
> Pacstall is the AUR Ubuntu wishes it had

lol. Thanks for doing this :)

rmcrackan:
In theory, I've now added you to the contributors list as seen in Libation > Settings > About. In practice, we need to add a scrollbar before it'll be visible: #1537 

No rest for the wicked.

niontrix:
Sure thing! :)

Issue #1552, opened 2026-01-13
Error attempting to export your library


JanaLumi:
PLEASE FILL OUT THE FOLLOWING. Bug reports with limited information or lacking an attached log file may get limited or delayed help.

___

## To Reproduce  

1. Go to Export in the menu bar and click on it
2. In the Export menu, click on Export Library
3. Attempted to save file in both in the user folder directly and documents folder, same error
4. See error:

Error attempting to export your library

Could not load file or assembly 'ClosedXML, Version=0.105.0.0, Culture=neutral, PublicKeyToken=fd1eb21b62ae805b'. The system cannot find the file specified.

   at ApplicationServices.LibraryExporter.ToXlsx(String saveFilePath, IEnumerable`1 libraryBooks) in D:\a\Libation\Libation\Source\ApplicationServices\LibraryExporter.cs:line 39
   at ApplicationServices.LibraryExporter.ToXlsx(String saveFilePath, IEnumerable`1 libraryBooks) in D:\a\Libation\Libation\Source\ApplicationServices\LibraryExporter.cs:line 39
   at LibationWinForms.Form1.exportLibraryToolStripMenuItem_Click(Object sender, EventArgs e) in D:\a\Libation\Libation\Source\LibationWinForms\Form1.Export.cs:line 29

## Expected behavior  

To have my library downloaded as an xlsx file

## Platform  
Windows 11

## Log Files
Attach your Libation log file here. If your user folder contains the file "LibationCrash.log", attach that also.

**Default Log File Locations**
C:\Users\%userprofile%\Libation
[Log202601.log](https://github.com/user-attachments/files/24577146/Log202601.log)

|Platform|Folder|
|-|-|
|Windows|`%userprofile%\Documents\Libation`|

Alternative, you may open the log file folder from within Libation. Open Libation's settings, and on the first tab in Settings you can click the button 'Open log folder'.

wtanksleyjr:
I see this too, I'm on Docker.

rmcrackan:
Sorry all. Fixed in [Libation version 13.1.3](https://github.com/rmcrackan/Libation/releases/latest)

PR #1558, opened 2026-01-13
Fix export, remove spatial audio option, and add file move progress.

Mbucari:
##  [Add file move progress](https://github.com/rmcrackan/Libation/commit/3948e25c991c5875aca1aaa6a7d3265a8cb395e7)

This is something I had hanging around for a while but just hadn't gotten around to integrating yet.

### DIFF
--- a/Source/AudibleUtilities/AudibleUtilities.csproj
+++ b/Source/AudibleUtilities/AudibleUtilities.csproj
@@ -6,7 +6,7 @@
 
   <ItemGroup>
     <PackageReference Include="AudibleApi" Version="10.1.2.1" />
-    <PackageReference Include="Google.Protobuf" Version="3.33.3" />
+    <PackageReference Include="Google.Protobuf" Version="3.33.4" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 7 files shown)


rmcrackan:
~~Before I merge this, can you tell if #1552 is a quick fix? It appears to be a new bug since the recent changes to the export backend.~~ Thanks for addressing the widevine ui so fast.

rmcrackan:
nvm, I see it in your commits

Mbucari:
You're welcome, @rmcrackan. I can't tell you how depressing this is to me. I wish I had counted the hours I devoted to adding the widevine support. On second thought, Knowing that figure would probably be even more depressing LOL. I banged my head against a wall for dozens of hours, trying all sorts of different avenues until I landed on the solution we have. I am comforted by the fact that we still have access to the highest quality stereo, and we will likely continue to have that access. A very small part of the catalog is available in Dolby Atmos. Still, losing it is depressing. Oh, and all the countless hours I spent figuring out how to add AC4 decoder support to aaxclean.codecs. That was just a few weeks old. FML.

rmcrackan:
I’m so sorry. This whole thing is a kick in the nuts.

I’d ask how other officially supported devices or itunes can listen to the new format but I bet the answer is: they can’t. and they’re such a tiny sliver of listeners that audible just doesn’t have to care

PR #1601, opened 2026-02-09
 Enable project-wide nullable reference types 

Mbucari:
There's nothing new in here, but there are a LOT of changes.


rmcrackan:
> Files changed 432

Suffice it to say I'm not exactly combing through this one

PR #1663, opened 2026-03-07
Add documentation on how to customize replacement characters on CLI

cvigano:
This MR adds missing documentation on how to customize the replacement characters in `Settings.json`, which may be required when the OS running `libationcli` does not line up with the expected filesystem for the library. That is the case when using the Docker image but the Library is on an NTFS filesystem.

### DIFF
--- a/docs/advanced/command-line-interface.md
+++ b/docs/advanced/command-line-interface.md
@@ -147,3 +147,81 @@ foreach($q in $Qualities){
   }
 }
 ```
+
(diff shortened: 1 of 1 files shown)


rmcrackan:
Thank you so much! I'm also going to play around with this and see if I can get my docs-publishing tool to create it as a collapsible box.

rmcrackan:
collapsible box added. tx again

Issue #1668, opened 2026-03-10
accessibility of login with classic libation


dgl1984:
The tab key is not allowing reliable navigation in libation classic.  Although its possible to log in, I accomplished this through, for lack of a better description, dumb luck.  After antigravity viewed the source, it told me the following, which hopefully is helpful.

# Issue Title: Accessibility: Keyboard Tabbing & Focus Issues in Classic (WinForms) Login/Accounts Flow

**Describe the bug**
As a keyboard/screen reader user, the tabbing behavior in the Classic (WinForms) interface for account setup and login is highly unreliable. Focus jumps unpredictably, gets trapped, or traverses UI elements out of logical visual order.

After reviewing the WinForms `Designer.cs` files, it appears this is caused by `TabIndex` collisions and fragmented sequential ordering in the dialogs.

**Steps to Reproduce (and underlying causes)**

**1. Accounts Dialog Tab Collision**
When opening the Accounts Dialog, pressing `Tab` does not follow a predictable path between the action buttons.
* **Root Cause details:** In `AccountsDialog.Designer.cs`, both the `importBtn` and `saveBtn` are explicitly assigned `TabIndex = 1;`. This collision confuses the Windows focus manager. 
* **Fix needed:** The `TabIndex` values should be sequential (e.g. DataGridView = 0, Import = 1, Save = 2, Cancel = 3).

**2. External Login Dialog Reading Order**
When using the "External Browser" login option, tabbing through the dialog jumps back and forth across the window visually, which makes reading the instructions with a screen reader very confusing.
* **Root Cause details:** In `LoginExternalDialog.Designer.cs`, the control TabIndices do not match the visual Top-to-Bottom flow. For instance, the `tldrLbl` has `TabIndex = 9` instead of keeping with the flow of the buttons.
* **Fix needed:** Renumber all TabIndices sequentially from `0` to `9` following the visual reading order (Locale Label -> Username Label -> Login URL Label -> Login URL TextBox -> Copy Button -> Launch Browser Button -> TLDR Label -> Instructions -> Response TextBox -> Submit Button).

**3. Embedded Web Login Focus Trap**
When using the embedded `WebLoginDialog.cs`, the `WebView2` control swallows keyboard focus, making it extremely difficult to Tab out of the browser view back into the parent WinForms controls to cancel or close the dialog. The javascript injecting `pass.focus()` upon load also forcibly yanks focus away from the user's current place. 
[...]

PR #1678, opened 2026-03-12
Add info in walkthrough for screen-reader users

rmcrackan:


rmcrackan:
Merge after getting confirmation from user in ticket #1668

Issue #1677, opened 2026-03-12
Libation crashes after fresh installation


jimbel656:
PLEASE FILL OUT THE FOLLOWING. Bug reports with limited information or lacking an attached log file may get limited or delayed help.

___

## Describe the bug  
I reinstalled libation and when starting it for the first time a fatal error occurs and libation must close.

"The configuration file './LibationFiles/Settings.json' was not found and is not optional. The expected physical path was '/usr/lib/libation/LibationFiles/Settings.json'."

The mention files exist, so i do not know what to do.
reinstalled it a few times, but it is the same everytime.

## To Reproduce  

1. Install libation
2. Open libation with command termional
3. Fatal Error occures

## Expected behavior  
LIbation starts normally

## Platform  
Linux Debian 13

## Log Files
[LibationCrash.log](https://github.com/user-attachments/files/25938888/LibationCrash.log)

**Default Log File Locations**
|Platform|Folder|
|-|-|
|Windows|`%userprofile%\Libation`|
|macOS|`~/Library/Application Support/Libation`|
|Linux|`~/.local/share/Libation`|

Alternative, you may open the log file folder from within Libation. Open Libation's settings, and on the first tab in Settings you can click the button 'Open log folder'. 

The configuration file './LibationFiles/Settings.json' was not found and is not optional. The expected physical path was '/usr/lib/libation/LibationFiles/Settings.json'.

   at Microsoft.Extensions.Configuration.FileConfigurationProvider.Load(Boolean reload)
   at Microsoft.Extensions.Configuration.ConfigurationRoot..ctor(IList`1 providers)
   at Microsoft.Extensions.Configuration.ConfigurationBuilder.Build()
   at LibationFileManager.Configuration.ConfigureLogging() in /home/runner/work/Libation/Libation/Source/LibationFileManager/Configuration.Logging.cs:line 36
   at AppScaffolding.LibationScaffolding.configureLogging(Configuration config) in /home/runner/work/Libation/Libation/Source/AppScaffolding/LibationScaffolding.cs:line 227
   at AppScaffolding.LibationScaffolding.RunPostMigrationScaffolding(Variety variety, Configuration config) in /home/runner/work/Libation/Libation/Source/AppScaffolding/LibationScaffolding.cs:line 153
[...]

PR #1679, opened 2026-03-12
Rmcrackan/1677 linux crash

rmcrackan:


rmcrackan:
Merge after getting confirmation from one of the people in ticket #1677

Issue #1704, opened 2026-03-30
Export of authorisation file to audible-cli appears to fail (and why)


Aberdyfi:
## Describe the bug  
Background: I have two audible accounts. I added the older (pre-amazon) account first. I was offered the opportunity to import my audible-cli authorisation file, which I accepted. That worked beautifully. 

Then I authorised my newer amazon account. That also worked beautifully. However, when I exported an authorisation file for use with audible-cli, audible-cli crashed. I am not an expert programmer (I know just enough to appreciate that Libation is a really great program), however inspection of the auth file revealed two possible causes.

1. The second line is `"website_cookies": {},`. audible-cli seems to prefer `"website_cookies": null,`. My guess is that this is not the cause.
2. A more substantive issue concerns `device_private_key`. audible-cli formats this as `"-----BEGIN RSA PRIVATE KEY-----\n` ... `\n-----END RSA PRIVATE KEY-----\n`. (The `\n` is important here.) In particular, they key itself is broken down by several `\n`. (Notepad++ displays the `\n` explicitly. It is not interpreted as a LF.) Each letter group consists of 64 characters except the last.

I did edit the auth file to fix these two issues, but my testing then broke down. (In a nutshell, audible-cli rejected the original auth file because "cookies did not take the expected form". Then it crashed, but that could be because I did not edit the auth file correctly. Specifically, I inexplicably broke the device_private_key into 65-character groups. I did not spot that error and so followed a different route.)

## To Reproduce  
Steps to reproduce the behavior: Export an auth file from Libation and see if it can be used in audible-cli.

## Platform  
Using Windows 10 on an oldish Thinkpad. 

## Log Files
There was no Libation crash. It is just that the import/export function is incredibly helpful.

rmcrackan:
Thanks for the details. I made a temp release for you. Can you try this and tell me if you're still having issues -- https://github.com/rmcrackan/Libation/actions/runs/23762876045#artifacts

Aberdyfi:
Sorry about the delay in getting back!

Here is what happened:
1. I can now export the authorisation file for my Pre-Amazon account successfully.
2. I still cannot export my Amazon account, but see the following.

[...]

PR #1708, opened 2026-03-30
fix bug with importing mkb79 auth with website_cookies: null vs empty

rmcrackan:


rmcrackan:
Attempting to fix #1704

PR #1713, opened 2026-04-03
Add `<first tag[...]>`

Jo-Be-Co:
In addition to #1690 the extra tag `<first tag>` was added. This will retrieve the first element from the list returned by `<tag>`.


rmcrackan:
I'll do a full review tomorrow. At first glance though this looks like good.

rmcrackan:
I did a real review and this looks perfect. Thanks again.

PR #1731, opened 2026-04-11
1729 docker resilient creation of db link

Jo-Be-Co:


rmcrackan:
I've been trying to dig through all of my out-of-town emails and I didn't notice your PR when I made a tweak for the docker issue. So I caused you a merge conflict. I apologize.

Jo-Be-Co:
My approach is more about checking for an existing link. And if one does exist, I’ll note that down and try to delete it first.
The method using `ln -sf` is a bit more direct, but it actually does the same thing.

Should I still resolve the merge conflict? Although I can’t really say at the moment whether I should create the link with or without `-f`.

PR #1735, opened 2026-04-13
1714 conditional tag cmp working on two properties

Jo-Be-Co:


Jo-Be-Co:
still need to add documentation and some more tests

Jo-Be-Co:
However, the changes have already been implemented to the extent that a review is possible – I’m open to feedback and requests for adjustments.

rmcrackan:
Overall this looks good. There are some broken tests related to `:`

Source/_Tests/LibationFileManager.Tests/TemplatesTests.cs
```diff
@@ -639,7 +639,7 @@ public void HasValue_test(string template, string expected)
 		[DataRow("<cmp author != 'Sherlock'->true<-cmp>", "true")]
 		[DataRow("<!cmp author != 'Sherlock'->false<-cmp>", "")]
 		[DataRow("<cmp tag = 'Tag1'->true<-cmp>", "true")]
-		[DataRow("<cmp tag[separator(:)slice(-2..)] = 'Tag2:Tag3'->true<-cmp>", "true")]
+		[DataRow("<cmp tag[separator( : )slice(-2..)] = 'Tag2 : Tag3'->true<-cmp>", "true")]
```
rmcrackan: For clarity, can you include this test with and without the spaces

rmcrackan:
Your latest commit looks nice.

Jo-Be-Co:
In fact, I would have liked to avoid the parsing of the property and leave it to the later evaluation.

However, situations will always come (like the : ) in which the blanket collection no longer works. Especially because, for example, everything possible can be used as a delimiter.

Therefore, the regular expression must then take into account the syntax with '...', "..." and [...]. All with somewhat different escape behaviour. And even so, we can't guarantee everything here ...

Jo-Be-Co:
> Your latest commit looks nice.

Did you intend to target the mathematical operations and symbols?

rmcrackan:
> Did you intend to target the mathematical operations and symbols?

Not sure what you mean, so I assume the answer is no.

Jo-Be-Co:
Based on your comment that my last commit looks nice, I first thought that you meant the corrections on the regular expression and then had the feeling that I finally actually added the comparisons for lists and the operators `≡≠≤≥∉∌∈∌⋂⊆⊇⊂⊃`.

Anyway, syntax is actually just a proposal. As this new `<cmp->` tag could do everything the `<is->` tag can do, the latter might be finesse-superceded, though it is that new...

rmcrackan:
Nope. I meant addressing the `:`

Jo-Be-Co:
Feature complete now – all that’s missing is the documentation. We could improve the syntax further if you don’t like my suggestion. But in my opinion, this could be merged once the documentation is ready.

rmcrackan:
This looks great as-is and I'm willing to merge now. You've put a ton of time into this though so I worked with a code analysis tool to give you a full report of possible improvements **if** you want to keep tweaking it. If you're ready to put this behind you, I'm perfectly happy with it's current version. Your call.

---

## Issues and risks (ordered by severity)

### 1. Likely performance regression for regex checks (`~` / `=~`)

Previously, `GetRegExpCheck` took the pattern string and built `Regex` once, then returned a delegate that only ran `IsMatch`.

The new `GetRegExpCheck(string exactName)` returns a delegate that does `new Regex(pattern, ...)` inside every invocation, with `pattern = v2.ToString()`.

For `<is ... [~pattern]->` and similar, `v2` is still logically fixed per compiled template, but every evaluation (every book/file processed) can now pay regex compilation again. That is much more expensive than `IsMatch` alone.

**Suggestion:** Restore "compile once per predicate" while keeping the two-argument shape... for example, capture `Regex` in a closure when `GetPredicate` builds the delegate for string ops (pattern still coming from `valStr`), and only use per-call compilation when the pattern is truly dynamic (for example `<cmp a ~ b->` where `b` is another property).

### 2. Numeric parse failure defaults to `0`

In the `num_op` branch:

```csharp
int.TryParse(valStr, out var valInt);
return (valInt, ...);
```

If `valStr` is malformed, `valInt` is `0` and comparisons proceed silently. That may produce surprising truth values instead of a clear error (or "false").

**Suggestion:** If `!int.TryParse`, either throw a template error with context (consistent with invalid regex) or treat as non-match with a warning, depending on project conventions.

### 3. `ToIntObject` throws `ArgumentOutOfRangeException`

For `#=` / `#>=` etc., non-handled types in `ToIntObject` throw. That can surface as a hard failure during naming for an unusual property type.

**Suggestion:** Return a sentinel / `null` and let the predicate return `false`, or map with explicit rules and tests (for example `long`, `double` if those appear in DTOs).

### 4. Single-operand `<is>` property capture is more restrictive

The opening matcher for the optional property changes from a broad `.+?` to something built on `[^<=~>!]` (plus non-greedy growth). That deliberately keeps `=`, `<`, `>`, `!`, `~` out of the "property" segment so binary `<cmp>` can parse operators.

**Risk:** Any property path or formatter edge case that legitimately needs those characters in the first segment of `<is ...>` could stop matching or match incorrectly. The updated tests (for example audible subtitle / spacing around `=`) suggest you already tuned behavior; it would be worth one explicit test for a contrived property or formatter token containing `=` if that is supported anywhere.

### 5. Unicode / normalization for list operators

Operators like combining sequences versus a single code point can confuse users and tests across editors and OSes. The docs table lists multiple spellings; that is honest, but copy/paste from the doc might not match what the regex accepts unless normalized consistently.

**Suggestion:** In docs, call out "prefer named (`:not_contains:`) or ASCII (`!>>`) forms for portability" (you already lean that way in tests).

### 6. Minutia

- Comment typo: "Condtionals" appears in regex comment blocks (pre-existing pattern extended).
- `private static (Object, ConditionEvaluator) GetPredicate`... prefer `object` for consistency with modern C# style.
- `GetRegExpCheck` compiling inside the hot path also affects `<cmp ... ~ ...>` when the right-hand side is not a literal (pattern changes per row), where compile-per-call might be intentional; consider a cache keyed by pattern string if compilation stays lazy.

Jo-Be-Co:
> This looks great as-is and I'm willing to merge now. You've put a ton of time into this though so I worked with a code analysis tool to give you a full report of possible improvements **if** you want to keep tweaking it. If you're ready to put this behind you, I'm perfectly happy with it's current version. Your call.

I'll check this.

> ## Issues and risks (ordered by severity)
> ### 1. Likely performance regression for regex checks (`~` / `=~`)

Here I don't have the pattern at hand at the time I construct the lambda. So the evaluation is somewhat delayed.
I can't think of situations where others as string literals will be used honestly as regular expressions. So it might be a good idea to cache the compiled expressions.

> ### 2. Numeric parse failure defaults to `0`
> In the `num_op` branch:

For the correct parsing I should be context sensitive. At the `<is>` tag the value shall be passed and parsed correctly. The `<cmp>` tag only need the operator.

> ### 3. `ToIntObject` throws `ArgumentOutOfRangeException`
> For `#=` / `#>=` etc., non-handled types in `ToIntObject` throw. That can surface as a hard failure during naming for an unusual property type.

Having all these unit tests there shouldn't be any type which would trigger the exception.
Though new tags might also introduce new types I would expect tests to fail.
While writing this it think of the need to throw `<locale>` and `<language>` into a numeric test ...
 
> ### 4. Single-operand `<is>` property capture is more restrictive
> The opening matcher for the optional property changes from a broad `.+?` to something built on `[^<=~>!]` (plus non-greedy growth). That deliberately keeps `=`, `<`, `>`, `!`, `~` out of the "property" segment so binary `<cmp>` can parse operators.

This might be from the time I wanted to introduce another syntax for the `<is>`. Here I needed to get rid off false catches.
Later I named the new part `<cmp>` so this change might not be needed anymore.

> ### 5. Unicode / normalization for list operators
> Operators like combining sequences versus a single code point can confuse users and tests across editors and OSes. The docs table lists multiple spellings; that is honest, but copy/paste from the doc might not match what the regex accepts unless normalized consistently.

Every of this new operators could be used by all ascii variants. The Unicode is optional therefore.
All the codes from the documentation are also used by the parser. This surely will help.
The :not_contains: and :not_in: symbols might also be composed. The parser catches both - here I don't see any chance for trouble.

rmcrackan:
That is all reasonable. Sounds like you're fine with everything as-is, correct? Ready for me to merge?

Jo-Be-Co:
no no. I'll do some updates. Caching for regex and the minor fixes at least.

Jo-Be-Co:
> > This looks great as-is and I'm willing to merge now. You've put a ton of time into this though so I worked with a code analysis tool to give you a full report of possible improvements **if** you want to keep tweaking it. If you're ready to put this behind you, I'm perfectly happy with it's current version. Your call.
> 
> I'll check this.
> 
> > ## Issues and risks (ordered by severity)
> > ### 1. Likely performance regression for regex checks (`~` / `=~`)
> 
> Here I don't have the pattern at hand at the time I construct the lambda. So the evaluation is somewhat delayed. I can't think of situations where others as string literals will be used honestly as regular expressions. So it might be a good idea to cache the compiled expressions.

Caching implemented ✅

> > ### 2. Numeric parse failure defaults to `0`
> > In the `num_op` branch:
> 
> For the correct parsing I should be context sensitive. At the `<is>` tag the value shall be passed and parsed correctly. The `<cmp>` tag only need the operator.

Parsing only done when `<is ...->` variant is choosen. The pattern now matches digits obligatory in that case. ✅

> > ### 3. `ToIntObject` throws `ArgumentOutOfRangeException`
> > For `#=` / `#>=` etc., non-handled types in `ToIntObject` throw. That can surface as a hard failure during naming for an unusual property type.

Added logics for dates. Unknows type will evaluate the check to false. ✅

> > ### 4. Single-operand `<is>` property capture is more restrictive
> > The opening matcher for the optional property changes from a broad `.+?` to something built on `[^<=~>!]` (plus non-greedy growth). That deliberately keeps `=`, `<`, `>`, `!`, `~` out of the "property" segment so binary `<cmp>` can parse operators.
> 
> This might be from the time I wanted to introduce another syntax for the `<is>`. Here I needed to get rid off false catches. Later I named the new part `<cmp>` so this change might not be needed anymore.

This extra security in tags will only reduce the set of tag names slightly. No problems involved. ✅

> > ### 5. Unicode / normalization for list operators
> > Operators like combining sequences versus a single code point can confuse users and tests across editors and OSes. The docs table lists multiple spellings; that is honest, but copy/paste from the doc might not match what the regex accepts unless normalized consistently.
>
> Every of this new operators could be used by all ascii variants. The Unicode is optional therefore. All the codes from the documentation are also used by the parser. This surely will help. The :not_contains: and :not_in: symbols might also be composed. The parser catches both - here I don't see any chance for trouble.

Nothing to do from my point of view. ✅

Jo-Be-Co:
> > > ### 4. Single-operand `<is>` property capture is more restrictive
> > > The opening matcher for the optional property changes from a broad `.+?` to something built on `[^<=~>!]` (plus non-greedy growth). That deliberately keeps `=`, `<`, `>`, `!`, `~` out of the "property" segment so binary `<cmp>` can parse operators.
> > 
> > 
> > This might be from the time I wanted to introduce another syntax for the `<is>`. Here I needed to get rid off false catches. Later I named the new part `<cmp>` so this change might not be needed anymore.
> 
> ~~This extra security in tags will only reduce the set of tag names slightly. No problems involved. ✅~~

I rolled it back. As I mentioned earlier, I kept it by mistake.

Source/_Tests/LibationFileManager.Tests/TemplatesTests.cs
Jo-Be-Co: sure

rmcrackan:
This is a really solid PR. Thank you again. Let me know when you're all done with edits and I'll merge.

Jo-Be-Co:
Oh, I'm done - ignite!

rmcrackan:
Merged

1) is #1714 fulfilled?
2) is #1750 ready for review?

Jo-Be-Co:
I'll check if there is anything to merge in #1750 and will also place a comment into #1714.

The latter might be closed with this PR or after the next rollout ...

PR #1745, opened 2026-04-17
Some optimizations after last update to AudibleApi

Jo-Be-Co:
Latest updates to the AudibleApi ensure correct setting of the language for each marketplace. Now all do have a region included and the lookup `Localization` => `language` => `CultureInfo` => `RegionInfo` will lead to regions usable for LocalDto.

### DIFF
--- a/Source/LibationFileManager/Templates/LocaleDto.cs
+++ b/Source/LibationFileManager/Templates/LocaleDto.cs
@@ -17,17 +17,13 @@ public partial record LocaleDto : IFormattable
 	private string DefaultFormat { get; }
 
 
-	public LocaleDto(string hint) : this(hint, "{O}")
+	public LocaleDto(string hint, string defaultFormat = "{O}") : this(Localization.Get(hint), hint, defaultFormat)
 	{
 	}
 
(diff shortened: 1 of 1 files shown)


Source/LibationFileManager/Templates/LocaleDto.cs
```diff
@@ -36,34 +32,14 @@ public LocaleDto(Locale locale, string hint, string defaultFormat)
...
-
 		try
 		{
-			return (new RegionInfo(countrcode), culture);
-		}
-		catch
-		{
-			// ignored
-		}
+			if (languageFromLocale == string.Empty) return (new RegionInfo(input), null);
```
rmcrackan: 2 lines plz

Jo-Be-Co: Fixed - though this is done quite often all over the code.

This is sometime slightly weird, when AI makes some line completions guesses and the IDE is reformatting inbetween.

rmcrackan:
Short and sweet. I like the easy ones

Jo-Be-Co:
Indeed - and this one was one of the _"do it now or you'll do it never!"_ tasks!

Issue #1754, opened 2026-04-20
Conditional output on format templates


Jo-Be-Co:
Add the possibility to do conditional formatting on properties with format templates like `{X}`.
  
> - for an author without an ASIN assigned to them (e.g. Virginia Grohl, Lacey Lamar, and likely many more random authors), it still leaves empty square brackets after the author's name. the `<has author[{ID}]-> [<author[{ID}]>]<-has>` does not work for author ID ... presumable because the logic is simply checking for the presence of the author property, which exists, and because the author ID is never a property itself, but rather just a format to represent the author property. 

A author might result in `Stephen--Hawking-[]` by formatting with `<first author[{F}-{M}-{L}-\[{ID}\]]>`. In this examples there is missing a middle name and the audible ASIN.

### Improvement
Allow additional braces around placeholders in format templates. Text between outer and inner brace may only be returned when the specified value exists and is non empty:

`{{F}-}{{M}-}{L}{-[{ID}]}` => `Stephen-Hawking`

The pattern needs to be escaped on special characters as it will be surrounded by `[...]` and possibly by `format(...)` on list properties.

Jo-Be-Co:
This is definitely implemented and rolled out.

~~I only gave up~~ I initiated this issue solely to provide a more precise account of the potential changes or at least the intended uses for them.

_This has been a bad translation ..._

rmcrackan:
Sounds good. Thanks. I'll close it.

PR #1755, opened 2026-04-20
#1754 Conditional output on format templates

Jo-Be-Co:


Jo-Be-Co:
I might add a text block to the user documentation - the coding is done ...

rmcrackan:
Looks good. I'll merge and release.

I think we're not very far from having to replace regex with an actual state-machine parser. CommonFormatters is getting gnarly.

rmcrackan:
In no time at all, you went from 'guy who made a cool PR' to 3rd biggest contributor to this project. I feel like I should get to know you a bit. What's your name, talented stranger? :)

PR #1758, opened 2026-04-22
fix naming template markdown

rmcrackan:


rmcrackan:
@Jo-Be-Co I'm going to merge so we can see the rendered version on the docs page, but treat this like an open PR -- I don't assume it's complete until you sign off. I made each commit deliberately focused for easier review.

docs/features/naming-templates.md
```diff
@@ -280,7 +280,8 @@ The format template must sometimes be enclosed in square brackets and sometimes
 
 Not all elements of a property are always present or have content. In this case, format templates would contain gaps after substitution. Groups of spaces are automatically merged. Other characters, however, remain unchanged. By doubling the curly brackets, you can specify text fragments before and after the placeholder, which are only used if the placeholder is replaced with content.
 
-`<narrator[format({{F} }{\({M}'\) '}{L})]>` → `Neil Gaiman, Christopher (Evan) Welch`
+<!-- VitePress compiles each .md as a Vue SFC: a literal `{{` inside normal backtick `code` is treated as Vue interpolation and breaks the build. Elsewhere this doc uses backticks for inline examples; this line is HTML `code` with `v-pre` so `{{` can be written literally and still render like the other snippets, without additional hacks such as obscure invisible characters (like &#200B : 0-width space). -->
+<code v-pre>&lt;narrator[format({{F} }{\({M}'\) '}{L})]&gt;</code> ? <code>Neil Gaiman, Christopher (Evan) Welch</code>
```
Jo-Be-Co, line 284: Good point about the invisible characters. Ideally, the examples should still be suitable for copying and pasting. However, you’ve inserted a question mark between the two code snippets here. I originally had a Unicode arrow here.
`&rarr;` or `&#8594;` or `&#x2192;`

rmcrackan: Good catch. My typical editor was being way too helpful/confused with all of the weirdly nested markdown and vue/vitepress syntax, so I edited the file in my standard text editor. It evidently doesn't recognize that glyph so it used `?`. I just changed it to a `->` for now because I don't readily know how to do the unicode arrow and I'm in a hurry right now.

Jo-Be-Co: I thought the html entities would help ...
Anyway it's just optics.

rmcrackan: I have no preference either way. It just wasn't convenient while I was making edits. If I think about it, I may fix it later.

PR #1768, opened 2026-04-26
Add native macOS "Check for Updates…" menu item

majouji:
## Summary
- add a native macOS `Check for Updates…` app-menu item near `About`
- route manual update checks through a shared Avalonia update-check view model
- reuse that shared state in the About dialog so button text and in-flight state stay consistent
- add targeted documentation around the shared updater/menu wiring

## Notes
- non-macOS behavior is unchanged

## Manual verification
- confirm `Libation > Check for Updates…` appears on macOS
- confirm it opens the same update flow as About
- confirm repeated clicks during an active check do not stack
- confirm `About Libation` still works

### DIFF
--- a/Source/LibationAvalonia/Dialogs/AboutDialog.axaml
+++ b/Source/LibationAvalonia/Dialogs/AboutDialog.axaml
@@ -15,7 +15,7 @@
 
 		<controls:LinkLabel Grid.Row="1" FontSize="14" VerticalAlignment="Center" Text="https://getlibation.com" Tapped="Link_getlibation"/>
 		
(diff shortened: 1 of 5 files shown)


rmcrackan:
Thanks for working on this. I get notices for new PRs, commits, and comments, but I don't know if it will tell me when it switches from draft to actual PR. So just drop a comment in here when you're ready for review.

majouji:
Sorry, won’t have time to thoroughly verify this and push it through but feel free to pick it up or take it over the finish line if you’d like 🙏🏽

rmcrackan:
No worries. tbh I don't even own a mac. All of the cool mac-specific work was done by @Mbucari . Lemme know if you feel like picking it up again later.

PR #1771, opened 2026-04-27
Re-add Category Name Column to database

Mbucari:
Category names were accidently removed during the last DB migration in d67692355f3911ce6722766c6f93a9d47bda4a03.

Re-add the Names column and modify CategoryImporter to upsert the names on update.


rmcrackan:
Thanks for catching this

PR #1790, opened 2026-05-05
Update Avalonia to v12

Mbucari:
Avalonia 12 uses compiled bindings by default. Converted all remaining reflection bindings into compiled bindings, which improves performance. Also, Avalonia 12 adds native linux accessibility, and the Avalonia WebView control is now free and open source.

Updated all other dependencies.

Fixed binding errors - These weren't every apparent to the user, but they existed and the framework worked around them. I Finally got around to fixing everything, and this should result in better performance.

Code cleanup


Mbucari:
@rmcrackan Also, I looked into the mac notarization failure, and it appears to have been a fluke. I re-ran the failed job and it succeeded.

Mbucari:
> and it appears to have been a fluke

Nevermind, it wasn't a fluke. The problem is that the "Build Bundle" step failed silently: `hdiutil: create failed - Resource busy`

I modified the mac bundler script to add a `sync` command before the `hdiutil` which should fix that.

rmcrackan:
Thank you so much -- both for the avalonia upgrade and for fixing the mac bit

Issue #1762, opened 2026-04-24
Extend filter and output options for list items like <author>


Jo-Be-Co:
> [...] Though, I really wish that the "count" syntax was for more than just `<ch count>`, thereby enabling `<author count>` and `<narrator count>` (and " count" in general) so that users can keep _short and tidy folder path names_ while still seeing which ones have multiple authors, enabling people to use the syntax of `<first author><is author count[#>=2]> + <author count[##-1]> more<-is>` to output "First Author + 11 more" for large multi-title compilations. In the absence of this being available I've gone with "et al." instead. Though, folder names in Windows can't end with a period, so it strips the  .  from the end of "et al." in folder names with 3+ authors, but leaves them in place for narrators in curly braces. 
>
> [...]
>
> - Not critical, but a _very_ nice-to-have tandem implementation, would be to ignore specific authors/narrators/composers such that people who want to ignore specific entities (e.g. "Audible", "Audible Original", or "Disney Press", or a translator) won't end up with many titles all masked in the folder structure and hidden in "Audible et al." or something silly like that. 

### Extend filter options
Add the possibility to filter all list items (`<author>`, `<narrator>`, `<series>` and `<tag>`). The validation of the filters should be implemented in the same way as the checks from `<is>` and `<cmp>`. The filters operate on a formatted version of the elements. The format is specified in the same way as the `format()` part of a list element. When constructing the command, the existing brackets must be taken into account. The property itself must still be recognised correctly and in full: `<author[... filter( ... ) ...]>`

Reduce the list of elements using ~`distinct()` or~ `unique()`. A format string can be used to define the criteria for evaluating unique values.

### Extend output options
The `count()` command is introduced to return the size of a list. The entries in the list are counted after `slice()`, `max()`, `filter()` and `unique()`. A number format can be specified with `count()`. If the list is empty, no number is returned.

Jo-Be-Co:
This should have rendered as feature request (enhancement).

@rmcrackan : If you have commented on this, possibly with a few comments about the syntax, I could implement this.

The most complex thing will probably be to make the check operations useable for this. The rest is straight forward.

rmcrackan:
I'm fine with adding new keywords like `count` inside of the formatting brackets. That area should only be for our keywords so it shouldn't cause a problem.

Jo-Be-Co:
Would you rather be in favour of `Unique()` or `Distinct()`, or both?

This is actually about one and the same regular expression.

[...]

PR #1795, opened 2026-05-07
#1762 add unique() and count() to list properties

Jo-Be-Co:


Jo-Be-Co:
OK - going to fix the OS depending tests this weekend

Jo-Be-Co:
Weird error on macOS build - could you try a rerun?

rmcrackan:
> Weird error on macOS build - could you try a rerun?

No problem. It looks like a re-run passed.

rmcrackan:
Wow -- there's a lot to love here. Just the CompareCondition refactor on its own is sexy.

The only significant minor issue that stood out to me is the interaction between separator and count. If I'm reading this right, then for `<author[separator(;) count(00)]>` , the separator branch wins and count() is ignored. (Likewise for format and count). This either needs to be accounted for or dis-allowed and documented. Either way it should get unit tests.

Speaking of tests, as always, I love seeing all of your new unit tests so we know how things look in the real world. A few others that I think should be added:

* `unique()` with no format -- the docs example `<tag[unique()]>` is shown but never tested
* `count(00)` zero-padded -- again: the example is shown in the docs but would be nice to see in unit tests
* `count()` over `<series>` and `<tag>` -- only `<author>` is tested; the implementation is shared but the tests don't prove that on the other list types. At present this isn't important because of the shared implementation. However, this will provide regression testing later in case of possible refactors
* Empty-list count returning nothing -- the `count == 0 ? [] : ...` branch
* Interaction with unique() + count() -- eg: distinct-author count

Minutia:

* typo in ReplacementCharacters.cs : `they user` -> `the user`
* ContributorDto.cs -- extracts a `DefaultFormat` const and routes `ToString()` through `TemplateStringFormatter` instead of conditionally short-circuiting. (Behavior is preserved, but it's an independent refactor.)
* The `NameFormat_formatters` test was migrated from `GetName(...)` to `GetFilename(...).PathWithoutPrefix` to match the rest of the file. That is fine, but it forced a test-data change `:` -> `_` because `:` is now stripped by the filename replacement pipeline. So the test is now also implicitly testing replacement-char behavior, which it wasn't before. None of this is wrong, but it would be cleaner as separate commits. (The branch already has 3 commits; the second/third commit messages "test fixes" and "extract comparison to separate reusable class" don't really cover these, so reviewers have to spot them.)
* Minor regex inconsistency (cosmetic). `UniqueRegex` uses `(?:[Uu]nique)\(...\)`; the non-capturing group is redundant. Other formatters use the bare `[Cc]ount`, `[Mm]ax`, `[Ss]lice`, `[Ss]eparator`.

Please don't be put off by the volume of my review. Nearly everything I found is minor. I think your core changes are marvelous: focused scope, localized code changes, follows existing patterns, good docs, unit tests. I'm frankly shocked at how little time you've been doing C# given your quality.

Jo-Be-Co:
> Wow -- there's a lot to love here. Just the CompareCondition refactor on its own is sexy.

Thanks.

> The only significant minor issue that stood out to me is the interaction between separator and count. If I'm reading this right, then for `<author[separator(;) count(00)]>` , the separator branch wins and count() is ignored. (Likewise for format and count). This either needs to be accounted for or dis-allowed and documented. Either way it should get unit tests.

`separator()` is really a method that only comes into play when there is more than one element. If `count()` has now consolidated everything into a single number, then `separator(;)` would actually be ineffective.
So, in the case of `<author[separator(;) count(00)]>`, I would provide the number.
`format()` is actually useless if a count is performed afterwards. Therefore, ideally, you should skip the formatting entirely when `count()` is present.
 
> [...]
> * Empty-list count returning nothing -- the `count == 0 ? [] : ...` branch

To demonstrate this, I have provided a test that uses `slice()` to create a set with 0 elements. More on this below.

> Minutia:
> 
> * typo in ReplacementCharacters.cs : `they user` -> `the user`
> * ContributorDto.cs -- extracts a `DefaultFormat` const and routes `ToString()` through `TemplateStringFormatter` instead of conditionally short-circuiting. (Behavior is preserved, but it's an independent refactor.)

I would put this into separate commits - would you like also a decoupled pull request?

> * The `NameFormat_formatters` test was migrated from `GetName(...)` to `GetFilename(...).PathWithoutPrefix` to match the rest of the file. That is fine, but it forced a test-data change `:` -> `_` because `:` is now stripped by the filename replacement pipeline. So the test is now also implicitly testing replacement-char behavior, which it wasn't before. None of this is wrong, but it would be cleaner as separate commits. (The branch already has 3 commits; the second/third commit messages "test fixes" and "extract comparison to separate reusable class" don't really cover these, so reviewers have to spot them.)

The test case described above (slice and count) was designed so that the number and the associated text would be completely removed. Any surrounding spaces were also to be removed. However, this functionality is only applied in the final step of `GetFilename(...).PathWithoutPrefix`. I will provide proper tests and undo this change.

> * Minor regex inconsistency (cosmetic). `UniqueRegex` uses `(?:[Uu]nique)\(...\)`; the non-capturing group is redundant. Other formatters use the bare `[Cc]ount`, `[Mm]ax`, `[Ss]lice`, `[Ss]eparator`.

Good spot! This is a left over from `unique|distinct` 😉 

> Please don't be put off by the volume of my review. Nearly everything I found is minor. I think your core changes are marvelous: focused scope, localized code changes, follows existing patterns, good docs, unit tests.

That's exactly right  👍 . I'm always pestering my colleagues with loads of 'optimizations' on their pull requests even when everything is already working perfectly fine.

> I'm frankly shocked at how little time you've been doing C# given your quality.

I always say that a good, experienced developer is, first and foremost, able to devise a solution to a problem using the right ideas and design patterns, and then put that solution into practice.
The knowledge required for this primarily concerns the libraries that can be used.
Only then does the practical knowledge of programming languages and their syntax come into play.

rmcrackan:
> `separator` ... `count`

Fair enough. I'd still show it in unit tests with a comment explaining the behavior.

> I would put this into separate commits - would you like also a decoupled pull request?

I have no preference. Your choice.

Let me know when this is ready for final review. It's a well-implemented enhancement.

Jo-Be-Co:
> Minutia:
> 
> * typo in ReplacementCharacters.cs : `they user` -> `the user`

I’m not sure if I’ve understood you correctly here. I thought I was just correcting a typo by writing ‘the user’ instead of ‘they user’. Is this one of those ‘them/they’ things, and is ‘they user’ actually the politically correct way to say it???

In that case I will skip this file completely as this was the only change on `ReplacementCharacters.cs`

BTW: I tried to layout the commits right without any fixed and interwoven other changes - but my IDE did not put all changes to the staging ...

rmcrackan:
> they user

"the user" is correct. One of us probably meant to say "they" or "the user" and the comment said "they user".

I meant to fix this quickly in master and I think I committed the typo-fix on your branch. I'm not sure what's going on now. Just ignore it. If it's still there after the merge, I'll do a quick fix later.

rmcrackan:
> I tried to layout the commits right without any fixed and interwoven other changes - but my IDE did not put all changes to the staging ...

Effort appreciated. Clean commits are nice, but it's not worth a lot of headache. The review is basically complete anyway. Once you tell me your tweaks are finished, I'll merge.

Jo-Be-Co:
Sorry for the delay – everything’s sorted now. If you’re happy with the current set of commits, feel free to merge them.

Jo-Be-Co:
Ah – now I’ve realised the misunderstanding. The typo appears twice. When I edited it the first time, I only corrected one of them.

When you mentioned the typo, I thought you were referring to my correction, even though you actually only meant the second one and had corrected that separately.

So I reset my branch to sort things out – which led to the conflict...

I'll do a last update.

PR #1801, opened 2026-05-10
fix: key LibraryBook dictionaries by (ProductId, Account) to prevent collision

### LINKED ISSUE #1800: Dictionary collision when two accounts own the same book
## Summary

In `DtoImporterService/LibraryBookImporter.cs`, both dictionaries built during `upsertLibraryBooks` are keyed by `AudibleProductId` / `ProductId` alone:

```csharp
// line 48
var existingEntries = DbContext.LibraryBooks.AsEnumerable()
    .Where(l => l.Book is not null)
    .ToDictionarySafe(l => l.Book.AudibleProductId);   // ← single key

// line 51
var uniqueImportItems = ToDictionarySafe(importItems,
    dto => dto.DtoItem.ProductId ?? string.Empty,       // ← single key
    tieBreak);
```

## Impact

When two accounts own the same book, both dictionaries can only hold **one entry per `AudibleProductId`**. The second account's `LibraryBook` (or import item) silently collides with and overwrites the first.

The existing workaround on line 60–64 (`SetAccount`) attempts to patch this by reassigning the surviving record to the incoming account — but this destroys the original account's ownership record rather than preserving both.

Concretely:
- Account A and Account B both own book X
- `existingEntries["X"]` = whichever `LibraryBook` was enumerated last
- When processing Account A's import item, the lookup finds Account B's `LibraryBook`
- `SetAccount` moves it to Account A — Account B's record is gone

## Steps to reproduce

1. Sign in to two Audible accounts in Libation
2. Ensure both accounts own at least one book in common
3. Run a library scan for both accounts
4. Observe that only one account retains the shared book after the import

## Expected behaviour

Each account's `LibraryBook` entry for a shared book is tracked and updated independently.

## Fix

Key both dictionaries by a composite `(ProductId, AccountId)` so each `(book, account)` pair gets its own slot. See the linked PR.

CryptoJones:
Closes #1800

## Summary

Both dictionaries in `upsertLibraryBooks` were keyed by `AudibleProductId` / `ProductId` alone. When two accounts own the same book, the second account's entry collides with and silently overwrites the first. The existing `SetAccount` workaround would then move the surviving record to the wrong account, destroying the original account's ownership.

**Fix:** use a composite `(ProductId, AccountId)` key in both dictionaries so each `(book, account)` pair is tracked independently.

## Changes

**`DtoImporterService/LibraryBookImporter.cs`**

| | Before | After |
|---|---|---|
| `existingEntries` key | `l.Book.AudibleProductId` | `(l.Book.AudibleProductId, l.Account)` |
| `uniqueImportItems` key | `dto.DtoItem.ProductId` | `(dto.DtoItem.ProductId, dto.AccountId)` |
| Lookup | `TryGetValue(item.DtoItem.ProductId, …)` | `TryGetValue((item.DtoItem.ProductId, item.AccountId), …)` |
| Absent-book check | `!uniqueImportItems.ContainsKey(asin)` | `!uniqueImportItems.ContainsKey((asin, lb.Account))` |

The `SetAccount` workaround block is removed — it is unreachable with the composite key since the found entry is guaranteed to belong to the correct account.

The private `ToDictionarySafe<TKey, TSource>` generic helper is unchanged; `(string, string)` is a non-nullable ValueTuple so it satisfies the `where TKey : notnull` constraint.

## Test plan

- [ ] Single-account import behaves identically to before
- [ ] Two accounts sharing at least one book: both accounts retain their own `LibraryBook` after a full library scan
- [ ] `AbsentFromLastScan` is set correctly per `(book, account)` pair, not per book globally

### DIFF
--- a/Source/DtoImporterService/DtoImporterService.csproj
+++ b/Source/DtoImporterService/DtoImporterService.csproj
@@ -14,6 +14,12 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
+      <_Parameter1>DtoImporterService.Tests</_Parameter1>
+    </AssemblyAttribute>
+  </ItemGroup>
+
+  <ItemGroup>
     <ProjectReference Include="..\AudibleUtilities\AudibleUtilities.csproj" />
     <ProjectReference Include="..\DataLayer\DataLayer.csproj" />
   </ItemGroup>
--- a/Source/DtoImporterService/LibraryBookImporter.cs
+++ b/Source/DtoImporterService/LibraryBookImporter.cs
@@ -29,17 +29,12 @@ public class LibraryBookImporter : ItemsImporterBase
 
 	private int upsertLibraryBooks(IEnumerable<ImportItem> importItems)
 	{
-		// technically, we should be able to have duplicate books from separate accounts.
-		// this would violate the current pk and would be difficult to deal with elsewhere:
-		// - what to show in the grid
-		// - which to consider liberated
-		//
-		// sqlite cannot alter pk. the work around is an extensive headache
-		// - update: now possible in .net5/efcore5
-		//
-		// currently, inserting LibraryBook will throw error if the same book is in multiple accounts for the same region.
-		//
-		// CURRENT SOLUTION: don't re-insert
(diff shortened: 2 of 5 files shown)


rmcrackan:
Closing per discussion in #1800

PR #1803, opened 2026-05-10
1762 add filter() to list properties

Jo-Be-Co:


Jo-Be-Co:
user documentation will follow

Jo-Be-Co:
Oh - to get feature complete, I should also add number literals so that `<cmp ...>` and `filter()` are more equivalent.

So if you can write `<cmp first author[{L}] #> 5->` you should also be able to write `<author[filter({L} #> 5)]->` with the same logic being applied.

rmcrackan:
To me this looks very good esp regard to feature symmetry with cmp and having focused, well-tested code. After you add the docs you mention, I'm ready to merge.

I ran the PR through an analysis tool and it also only had *very* minor suggestions:

* Maintainability -- The second argument to `GetPredicate` is not the same shape as a full conditional [...] check string; it is only the operator token that still satisfies `CheckRegex`. A one-line comment above that call would save the next reader from thinking the arguments are swapped or incomplete.
* Operator surface vs RHS -- `FilterRegex` allows the same rich op set as other features (list symbols, # numeric ops, etc.), but the RHS is always a quoted string run through `TryQuotedString`. That is fine for = / != / ~ style use, but list and numeric operators inside filter() may not line up with how `<cmp ...>` parses operands until the author’s planned number literal work (and any similar list-literal work) lands. Worth a short XML remark on `FilterRegex` or in user docs when they arrive.
* Docs -- The thread already notes user documentation will follow; linking filter() to the same mental model as bracket conditions (once numeric RHS exists) will help a lot.

Jo-Be-Co:
Just a quick answer to one of your comments
> * Maintainability -- The second argument to `GetPredicate` is not the same shape as a full conditional [...] check string; it is only the operator token that still satisfies `CheckRegex`. A one-line comment above that call would save the next reader from thinking the arguments are swapped or incomplete.

In my eyes the second argument to `GetPredicate()` is exactly the same as in the calls done by evaluating the `<cmp ...>` tags. In any case, only the matched operator is used, which comes from the respective regular expression for the entire syntax.

The syntax and the regular expression on the `<is ...>` tags differs slightly. Here, numbers are requested directly when one of the numerical operators is used. Then, however, `GetPredicateAndValue()` is used.

Would we like to have other the literals on the RHS?
Like `<narrator[filter({L} !<< author)]>`?
If so, the next challenge for the syntax on the RHS would also be to accommodate formatting!

Jo-Be-Co:
BTW: all list operations should work as intended because a single string is always handled as a single element list if needed.

But with only strings on both sides, it would be clearer to use string comparisons directly (i.e. `=` and `!=`).

rmcrackan:
I have no opinion about any of those automated suggestions. At worst, they just need a line explaining them. If you don't feel like adding it, that's fine.

This is a good PR. Once the user documentation has parity, I'm ready to merge.

Jo-Be-Co:
OK, everything's on board. I'm curious to see how the analysis tool rates it.

rmcrackan:
**EDIT:** this comment was automatically added with a software analyzing tool

---

Small copy-paste fixes (optional but nice for polish):

**1. Test method name**  
`CommonFormattersTests`: the test is still named `TryQuotedString_Various` but it exercises `TryGetLiteral`. Rename to `TryGetLiteral_Various` (or similar) so the name matches the API.

**3. Newline at end of file**  
Add a final newline at EOF (EditorConfig / POSIX style, quieter diffs on the next edit):  
- `Source/LibationFileManager/Templates/IListFormat[TList].cs`  
- `Source/_Tests/FileManager.Tests/CommonFormattersTests.cs` (if your diff still shows no newline there)

**4. Docs typo**  
`docs/features/naming-templates.md`, name list formatters table: the third example ends with ``<narrator[filter(~ 'Fry')]`` -- add the missing closing ``>`` on the tag, e.g. ``<narrator[filter(~ 'Fry')]>` `` so it matches real template syntax.

Thanks for the feature and docs.

rmcrackan:
**EDIT:** this comment was automatically added with a software analyzing tool

---

Clarification for item **4**: the doc example should end with the full closing tag, i.e. add the missing `>` after `]` so it reads `<narrator[filter(~ 'Fry')]> ` (same pattern as the other tags in that cell).

rmcrackan:
Yuck, I'm not doing that again. I tested a new feature in this analysis tool which will create the github comments for me. The 2 comments above are directly from this tool, not from me. Which feels really gross. However this comment you're reading is me. Actual me.

Your PR looks great and merge-ready. I guess since the tool found typos, we may as well fix those, then I'll merge.

Jo-Be-Co:
> Yuck, I'm not doing that again. I tested a new feature in this analysis tool which will create the github comments for me. The 2 comments above are directly from this tool, not from me. Which feels really gross. 

OK, but once you know 'who's speaking', the comments are fine.

> However this comment you're reading is me. Actual me.

🐍 👀 trust me 🤣

rmcrackan:
It's been a bad day for me vs. AI. I'm sure you noticed the recent AI PRs, some of which won't even compile. I'm assuming it's a free tier of Claude Code, but what reputable coding-specific AI generates code and doesn't even make sure it compiles??

Jo-Be-Co:
Yes, sometimes you lose and sometimes the others win 😉
Thanks for the merge.

PR #1810, opened 2026-05-14
Add <title sort>, <title short sort>, <first series sort> template tags (#1620)

### LINKED ISSUE #1620: Removal of "A", "And", and "The" from the beginning of either a book series or book title
**Is your feature request related to a problem? Please describe.**  
One of my minor frustration points is that Libation doesn't allow me to specify specific prefixes such as "A", "And", and "The" to omit from the beginning of a title. When sorting books, libraries skip over these prefixes. 

**Describe the solution you'd like**  
Allow the option to automatically remove the "A", "And", and "The".

**Describe alternatives you've considered**  
I just end up renaming the files using PowerRenamer.

CryptoJones:
## Summary

Adds three new template tags that strip a leading English article (A / An / The, case-insensitive) from the resolved value:

| Tag | Example input | Output |
|---|---|---|
| `<title sort>` | The Hobbit: There and Back Again | Hobbit: There and Back Again |
| `<title short sort>` | The Hobbit: There and Back Again | Hobbit |
| `<first series sort>` | The Lord of the Rings | Lord of the Rings |

Words that merely start with "The", "A", or "An" but aren't whole-word articles are untouched (e.g. "Theatre of War" → "Theatre of War").

## Usage examples

Folder template that files under the sort letter:
```
<title sort short> [<id>]
```
Mixed template that puts the full title in the filename but sorts by the stripped form in the directory:
```
<if series-><first series sort>\<-if series><title sort> [<id>]
```

## Changes

- **`TemplateTags.cs`** — three new `TemplateTags` static properties
- **`Templates.cs`** — `StripLeadingArticle()` private helper; registered in `filePropertyTags` (used by Folder, File, and ChapterFile templates) and the `chapterPropertyTags` inner collection
- **`TemplatesTests.cs`** — new `SortTags` test class: 14 cases covering all three tags, no-article pass-through, case-insensitivity, and availability in the chapter template

Closes #1620

### DIFF
--- a/Source/LibationFileManager/Templates/TemplateTags.cs
+++ b/Source/LibationFileManager/Templates/TemplateTags.cs
@@ -25,6 +25,8 @@ public sealed class TemplateTags : ITemplateTag
 	public static TemplateTags Id { get; } = new("id", "Audible ID");
 	public static TemplateTags Title { get; } = new("title", "Full title with subtitle");
 	public static TemplateTags TitleShort { get; } = new("title short", "Title. Stop at first colon");
+	public static TemplateTags TitleSort { get; } = new("title sort", "Full title with subtitle, leading article (A/An/The) removed");
+	public static TemplateTags TitleShortSort { get; } = new("title short sort", "Title without subtitle, leading article (A/An/The) removed");
 	public static TemplateTags AudibleTitle { get; } = new("audible title", "Audible's title (does not include subtitle)");
 	public static TemplateTags AudibleSubtitle { get; } = new("audible subtitle", "Audible's subtitle");
 	public static TemplateTags Author { get; } = new("author", "Author(s)");
@@ -33,6 +35,7 @@ public sealed class TemplateTags : ITemplateTag
 	public static TemplateTags FirstNarrator { get; } = new("first narrator", "First narrator");
 	public static TemplateTags Series { get; } = new("series", "All series to which the book belongs (if any)");
 	public static TemplateTags FirstSeries { get; } = new("first series", "First series");
+	public static TemplateTags FirstSeriesSort { get; } = new("first series sort", "First series name, leading article (A/An/The) removed");
 	public static TemplateTags SeriesNumber { get; } = new("series#", "Number order in series (alias for <first series[{#}]>");
 	public static TemplateTags Minutes { get; } = new("minutes", "Length in minutes");
 	public static TemplateTags Bitrate { get; } = new("bitrate", "Bitrate (kbps) of the last downloaded audiobook");
--- a/Source/LibationFileManager/Templates/Templates.cs
+++ b/Source/LibationFileManager/Templates/Templates.cs
@@ -269,6 +269,8 @@ public abstract class Templates
 		{ TemplateTags.Id, lb => lb.AudibleProductId, v => v },
 		{ TemplateTags.Title, lb => lb.TitleWithSubtitle },
 		{ TemplateTags.TitleShort, lb => GetTitleShort(lb.Title) },
+		{ TemplateTags.TitleSort, lb => StripLeadingArticle(lb.TitleWithSubtitle) },
+		{ TemplateTags.TitleShortSort, lb => StripLeadingArticle(GetTitleShort(lb.Title)) },
 		{ TemplateTags.AudibleTitle, lb => lb.Title },
 		{ TemplateTags.AudibleSubtitle, lb => lb.Subtitle },
 		{ TemplateTags.Author, lb => lb.Authors, NameListFormat.Formatter, NameListFormat.Finalizer },
@@ -277,6 +279,7 @@ public abstract class Templates
 		{ TemplateTags.FirstNarrator, lb => lb.FirstNarrator, CommonFormatters.FormattableFormatter },
 		{ TemplateTags.Series, lb => lb.Series, SeriesListFormat.Formatter, SeriesListFormat.Finalizer },
 		{ TemplateTags.FirstSeries, lb => lb.FirstSeries, CommonFormatters.FormattableFormatter },
+		{ TemplateTags.FirstSeriesSort, lb => StripLeadingArticle(lb.FirstSeries?.ToString()) },
 		{ TemplateTags.SeriesNumber, lb => lb.FirstSeries?.Order, CommonFormatters.FormattableFormatter },
 		{ TemplateTags.Language, lb => lb.Language, CommonFormatters.FormattableFormatter },
 		//Don't allow formatting of LanguageShort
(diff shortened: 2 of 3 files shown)


Jo-Be-Co:
I’ve considered similar cuts before.

Here are a few general comments:

* These cuts should be available for every language so that the new tags also work for non-English books.

* Would we always remove all leading components of all languages, or should we refer to the respective language of the book?

* Currently, only a few selected tags are provided with the new logic. While this doesn’t make sense for every tag, there are at least more title tags and the list tag for series.

* How about a formatting option for text output? In addition to indicating uppercase, lowercase, titlecase, and length restrictions, we could also include an additional S for a `<title[10l]>`.

* Since you’ve already named the new tags with “sort,” would that also be a possible sorting option (at least for series)?

A completely different approach would be to give the user the possibility of text replacement. For example, with a replace-tag that is set around another tag. The regular expression `^(A|An|The) ` could, for example, capture the current list.

rmcrackan:
This PR does not compile. Please test AI-generated code locally before submitting.

CryptoJones:
Sincere apologies — you're absolutely right, and I'm sorry for the wasted CI runs. The PR went up without a local build (the new test class moved to a different namespace and lost access to the unqualified `Shared` symbol). Just pushed `002ebbb6` with the fix and verified `dotnet test` locally first this time — 14/14 new test cases pass.

Also: [obligatory xkcd 2347](https://xkcd.com/2347/) — you're the Nebraska person here. Thank you for the time you sink into reviewing AI-generated noise like this.

CryptoJones:
Thanks for the careful review, @Jo-Be-Co. Genuinely good points:

- **Multi-language articles:** Agreed, hard-coding English `A/An/The` is a real limitation. Cleanest fix is probably keying off the book's `Language` field. The replace-tag idea you raise below would subsume this entirely.
- **Tag selectivity:** Fair — narrow scoping was deliberate but extending to author / series list is straightforward; happy to do that here or as a follow-up.
- **Format-option vs. dedicated tag:** I prefer your `<title[…s]>`-style modifier over a separate `<title sort>` tag — it composes better. Open to reworking the PR that way if maintainers prefer.
- **`series sort` as a sort order:** good catch, worth a separate issue.
- **Replace-tag with regex:** that's the most flexible design and probably the right long-term answer. I think it deserves its own issue rather than expanding this PR.

Will hold off on rework until there's direction from maintainers.

rmcrackan:
I'm *really* torn on this. I like this idea. It very much feels like the kind of thing I might have included myself in Libation's early days when it was just me and my English-speaking self. And let's be honest -- everything about *Libation* is anglocentric. **BUT** the books it liberates are not -- and this feature is about those books and their potentially non-English metadata. I'll think about this.

I'm not crazy about the proposed syntax but you and @Jo-Be-Co are working through that; I'll chime in after you 2 come to a consensus.

rmcrackan:
I couldn't find an off-the-shelf solution and it looks like Humanizer removed this in v3 (boo!). But a good-enough version seems straight-forward (I know: famous last words)

```c#
private static Dictionary<string, string[]> LeadingArticles { get; } = new(StringComparer.OrdinalIgnoreCase)
{
    ["en"] = new[] { "the", "a", "an" },
    ["fr"] = new[] { "le", "la", "les", "l'", "un", "une", "des" },
    ["es"] = new[] { "el", "la", "los", "las", "un", "una", "unos", "unas" },
    ["it"] = new[] { "il", "lo", "la", "i", "gli", "le", "l'", "un", "uno", "una" },
    ["de"] = new[] { "der", "die", "das", "ein", "eine", "einen", "einem", "einer" },
    ["pt"] = new[] { "o", "a", "os", "as", "um", "uma", "uns", "umas" },
    ["nl"] = new[] { "de", "het", "een" },
    ["sv"] = new[] { "en", "ett" },
};

public static string ToSorted(string title, string? languageHint = null)
{
    var trimmed = title.TrimStart();
    var lang = languageHint ?? DetectLanguage(trimmed) ?? "en";
    if (LeadingArticles.TryGetValue(lang, out var articles))
    {
        foreach (var art in articles)
        {
            var prefix = art.EndsWith("'") ? art : art + " ";
            if (trimmed.StartsWithInsensitive(prefix))
                return trimmed[prefix.Length..].TrimStart();
        }
    }
    return trimmed;
}
```

While I was playing with this, I also found some string sorting algo notes about sorting without diacritics, which relates to the other PR discussion. The Normalization Form D on line 1 is "canonical decomposition". It decomposes single-character with accent into latin letter plus a combining character. Prints the same but is now 2 characters. (Which allows us to strip the non-latin character in a later step.) ([Normalize method](https://learn.microsoft.com/en-us/dotnet/api/system.string.normalize?view=netframework-4.8.1&viewFallbackFrom=net-10.0), [NormalizationForm enum](https://learn.microsoft.com/en-us/dotnet/api/system.text.normalizationform))

example:  
* before:  
  U+00E9 (LATIN SMALL LETTER E WITH ACUTE)
* after:  
  U+0065 (plain e)  
  U+0301 (COMBINING ACUTE ACCENT)

```c#
var normalized = trimmed.Normalize(NormalizationForm.FormD);
var sb = new StringBuilder(normalized.Length);
foreach (var ch in normalized)
    if (CharUnicodeInfo.GetUnicodeCategory(ch) != UnicodeCategory.NonSpacingMark)
        sb.Append(ch);
return sb.ToString().ToLowerInvariant();
```

Jo-Be-Co:
Two good approaches that point in the right direction.

Whereby I would not be sure whether I like the special treatment of `'`` or if I would add the spaces to the other terms. In the end, it will come down to a developer maintained solution anyway.

I would probably have stored one regular expression per language here again. 🤓

The topic of Unicode accent removal is exactly how to customise a large part without the need for explicit mapping. But we shouldn't delve into this here.

Jo-Be-Co:
So what solutions do I see here:

* The existing implementation should be extended to other languages. In most cases, the language of the book should be the correct choice.

* When implementing with new tags, you should immediately become feature complete. A `<series sort>` will then be a bit more complex.

* The series tags already have a format option `{N}` for the output of the value. Here you could also add another letter that implements the shortening. This would then also be useable directly for sorting `<series[format({X}) sort(X)]>`. Unfortunately, this does not yet exist with the simple text fields. But would be a consideration (analogous to `<tag[{S:3}]>`). `<title[5U]>` `<title[{S:5U}]>` `<title[{X:5U}]>` ...

* On the other hand, only one place where the extension is installed is to intervene in the text formatting. But it works everywhere. But will not be needed everywhere. '<title[X]>`, `<series[format({N:X})]>`

* A `<replace>` tag or as an option for the text outputs is certainly very powerful and universally useable. But this feels like driving a truck shopping.

Personally, I would prefer one of the variants with `X`, whereby the `X` should of course be a wisely chosen letter here.

Jo-Be-Co:
> ```c#
> 
> var normalized = trimmed.Normalize(NormalizationForm.FormD);
> 
> var sb = new StringBuilder(normalized.Length);
> 
> foreach (var ch in normalized)
> 
>     if (CharUnicodeInfo.GetUnicodeCategory(ch) != UnicodeCategory.NonSpacingMark)
> 
>         sb.Append(ch);
> 
> return sb.ToString().ToLowerInvariant();
> 
> ```

My compact variant:
```c#
var cleaned = Regex.Replace(input.Normalize(NormalizationForm.FormD), @"\p{Mn}+", "");
```

I think that it is quite likely that these replacements are rather rare. Then the character-by-character processing by a StringBuilder would only generate a string copy. A regexp without a match might be more efficent here.

rmcrackan:
Thanks for the patience on this, and thanks again @Jo-Be-Co for the design pushback - that discussion is exactly what this needed.

I've thought it through and we shouldn't merge this as-is. The feature itself is good. Sorting past leading articles is a real library-organization need. What's blocking it is the shape of the solution:

1. **English-only won't cut it.** Libation's UI is anglocentric; book metadata is not. Hard-coding `A` / `An` / `The` would leave non-English titles out, which is the wrong tradeoff for a naming feature. Article stripping should key off the book's language (with a sensible fallback).

2. **New tags are the wrong API.** Three dedicated `* sort` tags don't compose, and `<first series sort>` already loses the series format tokens (`{N}`, `{#}`, etc.). I'd rather see this as a **format option** on existing text/series output - the `X`-style modifier @Jo-Be-Co suggested - so it works on `<title>`, `<title short>`, series `{N}`, and friends without tag sprawl.

3. **A general `<replace>` tag** is interesting but too broad for this issue. Separate discussion if someone wants to pursue it.

If you're still interested in landing #1620, a follow-up should:

- add a format modifier (letter TBD - pick something that won't collide with existing `u`/`l`/`t`/length syntax)
- strip leading articles using the book's language
- cover the main title and series name paths (not only three hand-picked tags)
- keep existing templates unchanged

Issue #1872, opened 2026-06-14
One title will not download from Audible


abrodel-OK:
PLEASE FILL OUT THE FOLLOWING. Bug reports with limited information or lacking an attached log file may get limited or delayed help.

___

## Describe the bug  
One book added to my audible library from the plus area of audible will not dowload.  I get the message Error will try later.   I have tried removing the book from libation and my audible library waiting a few days and then adding the book again and scanning and liberating, same thing happens.   the book is "Star Force: Enlightenment: Star Force Universe"+
## To Reproduce  
just try again 

## Expected behavior  
book downloads
## Screenshots  

<img width="394" height="169" alt="Image" src="https://github.com/user-attachments/assets/2e9fb087-e021-433f-ba73-583011aa3f99" />

## Platform  
Windows 10

## Log Files

[Log202606.log](https://github.com/user-attachments/files/28935471/Log202606.log)

rmcrackan:
@abrodel-OK Thank you much for your log. I was sure this was going to be another easy fix or misunderstanding. Boy was I wrong. You uncovered a beast. Thanks!

@Mbucari , @mkb79 -- Who's up for a challenge? B08XDZCH78 is in the US Plus catalog. It downloads fine on the official iPhone app. It shows up in my library scan and I cannot get Libation to download it (can't get license).

> POST https://api.audible.com/1.0/content/B08XDZCH78/licenserequest

response has `acr:null` :  
```json
{
  "error_code": "000307",
  "message": "Unable to retrieve asset details from Sable(AssetInfos), for marketplaceId:AF2M0KC94RCEA, asin:B08XDZCH78, acr:null, skuLite:BK_ACX0_239890, version:LATEST, aaaClientId:urn:cdo:AudibleApiExternalRouterService:Prod:Default"
}
```

I've been hacking at it all morning and have gotten nowhere. I asked cursor AI to create a tech brain dump of everything I tried this morning:

# EVERYTHING BELOW THIS LINE IS FROM CURSOR AI

---

### Baseline (stock AudibleApi behavior)

- Modern `licenserequest` body with codec negotiation -> `000307`
- No AAX fallback success (no codecs in library metadata) -> `FailedRetry`

### Attempt 1: Legacy `licenserequest` body on `000307`

On modern-body `000307`, retry with pre-AAXC body (audible-cli / iOS shape):

```json
{
  "supported_drm_types": ["Mpeg", "Adrm"],
  "quality": "High",
[...]

Issue #1580, opened 2026-01-30
Error downloading, possibly due to two colons in title


DeepSky8:
## Describe the bug  
I have added a Audible Plus book to my library. The book consistently errors instead of downloading. 

When I tried searching for the book title in my library, the search function threw an error (image below). This makes me think that Libation may have an issue with the second colon.

The title is "Star Force: Enlightenment: Star Force Universe"

## To Reproduce  

1. Add "Star Force: Enlightenment: Star Force Universe" to Audible library.
2. Scan/download library
3. Receive download error
4. Search title in library 
5. Receive search error

## Expected behavior  
Book downloads, or provides guidance on managing the error

## Screenshots  

<img width="1886" height="133" alt="Image" src="https://github.com/user-attachments/assets/c5eccd06-fbf6-4e24-bfec-39f32d096ab5" />

<img width="619" height="414" alt="Image" src="https://github.com/user-attachments/assets/045d1195-4c85-4b5a-9c41-50fd7b887086" />

## Platform  
Windows 11

## Log Files

[Log202601_003.log](https://github.com/user-attachments/files/24965826/Log202601_003.log)

Mbucari:
@DeepSky8 Two things are happening.
1. Your book is failing to download because you're not being granted a license from Audible. It's an Audible Plus title, and this book is only available to be downloaded if you're using Widevine DRM. @rmcrackan This is another canary in the coal mine. A standard Plus title that can't be downloaded as AAXC.
2. You're searchinf for a raw string that has colons in it, but colons are query control characters. Place your search in quotes: `"Star Force: Enlightenment: Star Force Universe"`

DeepSky8:
Many thanks!

rmcrackan:
@Mbucari You're joking, right? They're starting to move ordinary Plus titles to widevine L1? Looks like we'll be removing AYCL from Libation soon :(

Mbucari:
> [@Mbucari](https://github.com/Mbucari) You're joking, right? They're starting to move ordinary Plus titles to widevine L1? Looks like we'll be removing AYCL from Libation soon :(

It's L3, so Libation can still get it for now if "Use Widevine DRM" is enabled. But Hard times are coming, I expect.

rmcrackan:
Yeah :( It pained me to have to write this:

https://old.reddit.com/r/audiobooks/comments/1qu1qvy/libation_users_how_to_you_listen_to_your/o37dyw9/

mkb79:
[...]

PR #1874, opened 2026-06-16
#1872 : Suggest Widevine when ADRM licenserequest fails with Sable acr:null

rmcrackan:
#1872 : Suggest Widevine when ADRM licenserequest fails with Sable acr:null


rmcrackan:
See also: https://github.com/rmcrackan/Libation/issues/1580#issuecomment-4709540200

PR #1907, opened 2026-08-02
Feature/improve environment in docker and documentation

oxivanisher:
* Setup libation-master.key location when not set via variable.
* Support user specified cmd by not entering the liberation loop and running the cmd while still setting up the environment in the container.
* Use entrypoint instead of cmd to allow the user to supply their own cmd in the container.
* Create a symlink to LibationCli to not require the full path in the container.
* Add missing LOG_LEVEL variable to the env variables table.
* Document on how to access or redirect log files from within docker.
* Also expand documentation and example for the interactive shell.
* Ensure .sh files are in unix style (.gitattributes).
* Offer my help for docker issues.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.sh text eol=lf
--- a/Docker/liberate.sh
+++ b/Docker/liberate.sh
@@ -38,6 +38,25 @@ init_config_file() {
   fi
 }
 
+init_master_key() {
+  FILE=libation-master.key
+  FULLPATH=${LIBATION_CONFIG_DIR}/${FILE}
+  if [ -n "${LIBATION_MASTER_KEY_FILE}" ]; then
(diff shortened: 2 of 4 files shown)


rmcrackan:
Thanks for this - really helpful.

The `init_master_key` change is the important fix: Docker uses `/config-internal` as LibationFiles, so staging `libation-master.key` from `/config` (like the JSON files) is what makes the documented portable-key workflow actually work. The ENTRYPOINT / custom-command / docs pieces look good too.

I'll merge this and get a Docker image rebuild out so users pick it up. I appreciate the help on Docker issues.

rmcrackan:
Crap, I missed [adding this](https://github.com/rmcrackan/Libation/pull/1908/changes) when I deployed your docker fix. It will be included in all future releases.

PR #1918, opened 2026-08-08
Add CLI abs upload for Audiobookshelf backfill

caiowilson:
## What

Adds a `upload` verb to LibationCli:

```console
libationcli upload                 # every liberated book
libationcli upload B017V4IM1G      # a specific title
libationcli upload --id B017V4IM1G
```

It uploads books to Audiobookshelf using the audio files already on disk. Nothing is re-downloaded from Audible. No local file is moved or deleted.

Also fixes a file-lookup defect in `UploadToAudiobookshelf` that this verb exposes, and narrows which books qualify for upload.

## Why

`UploadToAudiobookshelf` already exists and works, but it has exactly one entry point: `LiberateOptions.CreateBackupBook` chains it onto `DownloadDecryptBook.Completed`. Upload only ever happens as a side effect of a fresh download.

That leaves no path to the server for:

- books liberated before Audiobookshelf was configured
- books liberated while the integration was disabled
- books whose upload failed once and was never retried

The only existing workaround is `liberate --force`, which re-downloads the entire library and re-consumes Audible bandwidth to move files that are already on disk.

### The verb

`UploadOptions : ProcessableOptionsBase` (`Source/LibationCli/Options/UploadOptions.cs`). `Setup.LoadVerbs()` discovers verbs by reflection, so it self-registers.

The base class already provides everything the verb needs: positional ASIN and `--id` targeting, `--override`, `--libationFiles`, the console progress bar, and stderr error printing. `UploadToAudiobookshelf.Validate` already selects the right candidate set. The verb is mostly plumbing an existing strategy into an existing runner.

Before iterating, it checks that Audiobookshelf is enabled and that server URL, token, library, and folder are all set. Without that check `Validate` rejects every book and the run looks like a silent no-op. All five settings carry `[Description]`, so they can be supplied per-run with `--override` for headless or Docker use.

### File-lookup defect

`Validate` and `GetFilesToUpload` consulted independent sources of truth:

| Method | Source | Meaning |
|---|---|---|
| `Validate` | `Book.AudioExists` (`EntityExtensions.cs:23`) | a DB status flag |
| `GetFilesToUpload` | `FilePathCache.GetFiles` | a JSON cache of paths |

A book liberated long ago has `BookStatus == Liberated` and passes `Validate`. If the path cache has no entry for it — cache lost, files moved, or liberated by a version predating the cache — `GetFilesToUpload` returned an empty list, and `ProcessAsync` returned an empty `StatusHandler`, which means success. The book was silently skipped and the run reported done.

The existing chained call site never hits this, because it runs seconds after the download that populated the cache. Bulk backfill hits it constantly.

Lookup now goes through `AudibleFileStorage.Audio.GetPaths`, which unions the path cache with a live scan of the Books directory. That method's regex is built per-`FileType` from that type's extensions, so the disk scan cannot pull in PDFs or cover art.

### Candidate filter

`Validate` now requires `LiberatedStatus.Liberated`. It previously used `Book.AudioExists`, which also accepts `LiberatedStatus.Error`. An errored liberation may have left partial files, and uploading those is worse than skipping them. Those books were only ever reachable here by accident.

### Outcome reporting

New event on `UploadToAudiobookshelf`:

```csharp
public enum UploadOutcome { Uploaded, AlreadyExists, NoFilesFound, Failed }
public event EventHandler<UploadOutcomeEventArgs>? OutcomeDetermined;
```

Raised once per processed book. The verb subscribes, writes failures to stderr, and prints an end-of-run summary:

```
Audiobookshelf upload summary:
  uploaded:          12
  already on server:  3
  no files found:     1
  failed:             2
  skipped:            0
```

Exit code stays 0 in all cases, matching every other processable verb.

## Design notes

**`ProcessAsync` still always returns a successful `StatusHandler`.** This was the one non-obvious call. It would be natural for a command whose only job is uploading to return real errors, but `UploadToAudiobookshelf` is not CLI-only — `ProcessBookViewModel.cs:272` adds it to the GUI process queue. There, a non-success `StatusHandler` falls through to `ProcessBookResult.None`, which invokes `GetFailureActionAsync` (the per-book Abort/Retry/Ignore dialog), breaks the step loop, and marks the book `Failed`. A failed Audiobookshelf upload would have failed a GUI liberation, contradicting the documented behavior in `docs/features/audiobookshelf.md`. Failures therefore travel on `OutcomeDetermined`, which is why that event carries a message. `LiberateOptions` and the GUI queue do not subscribe and are unaffected. A regression test pins this.

**`skipped` exists because the event cannot see every book.** On a targeted run, `ProcessSingleAsync` returns `"Validation failed"` without ever calling `ProcessAsync`, so no outcome is raised. Without a skipped count, `libationcli upload B0NOTLIBERATED` would print a summary of all zeros. The verb counts candidates through `RunAsync`'s per-book callback and derives the difference.

**No database migration, deliberately.** Libation records nothing about what it has already uploaded. Deduplication already runs server-side: `UploadBookAsync` calls `BookExistsAsync` before every upload and returns `AlreadyExists` on a match. Backfill is correct without local state. The cost is speed on repeat runs — roughly two library-search calls per candidate, sequential. That seemed the right trade for a command intended to run once. A `LastUploadedToAudiobookshelf` field on `UserDefinedItem` would make repeat runs fast and would also give the existing auto-upload a memory it currently lacks, but it did not seem worth a migration across both SQLite and Postgres for this. Happy to add it if you would rather have it.

**No GUI change.** The verb is CLI-only.

## Testing

New `Source/_Tests/FileLiberator.Tests/UploadToAudiobookshelfTests.cs`, 20 tests covering `Validate`, the pure upload-file-list composition, the disk-scan lookup, and the soft-fail invariant.

The defect fix was verified by reproducing it first: `GetAudioFilesOnDisk` was implemented with the old cache-only lookup and the test failed with `Expected:<1>. Actual:<0>` against a file sitting on disk, then the lookup was switched.

`FileLiberator` gains a `_InternalsVisible.cs`, matching the existing pattern in `LibationSearchEngine` and `AudibleUtilities`.

Locally: 1012 tests across all 7 test projects pass. `FileLiberator`, `LibationUiBase`, `LibationAvalonia`, and `LibationCli` build. WinForms was not built locally (macOS); CI covers it.

## Notes

- **No `LibationCli.Tests` project exists**, so the verb itself has no unit tests. It is verified by build and by `--help` output. Say the word if you would like a test project added.
- **End-to-end upload was not exercised against a live Audiobookshelf server.** The offline paths are tested; the actual `UploadBookAsync` round trip is unchanged by this PR but the new entry point into it has not been run against real hardware.
- Docs updated: `docs/features/audiobookshelf.md` gains a backfill section, and `docs/advanced/command-line-interface.md` no longer claims upload happens only during `liberate`.
- The token double-decrypt in `UploadBookAsync` (decrypts, then passes the decrypted value to `BookExistsAsync`, which decrypts again) was checked and is harmless — `DecryptToken` returns its input unchanged without the `enc:` prefix. Left alone. Mentioning it only so it is not re-investigated.

### DIFF
--- a/Source/FileLiberator/UploadToAudiobookshelf.cs
+++ b/Source/FileLiberator/UploadToAudiobookshelf.cs
@@ -15,6 +15,27 @@ public class UploadToAudiobookshelf : Processable, IProcessable<UploadToAudioboo
 {
 	public override string Name => "Upload to Audiobookshelf";
 
+	public enum UploadOutcome { Uploaded, AlreadyExists, NoFilesFound, Failed }
+
+	public sealed class UploadOutcomeEventArgs(UploadOutcome outcome, string message) : EventArgs
+	{
+		public UploadOutcome Outcome { get; } = outcome;
+		public string Message { get; } = message;
+	}
+
+	/// <summary>
+	/// Raised exactly once per processed book, classifying what happened and why.
+	/// <para/>
+	/// Upload failures are reported here rather than through the returned <see cref="StatusHandler"/>.
+	/// A non-success StatusHandler marks the book as a bad book: the GUI process queue breaks its step
+	/// loop and raises the Abort/Retry/Ignore dialog. Uploading is a courtesy step layered onto
+	/// liberation and must never fail the book, so <see cref="ProcessAsync"/> always reports success.
+	/// </summary>
+	public event EventHandler<UploadOutcomeEventArgs>? OutcomeDetermined;
+
+	private void OnOutcomeDetermined(UploadOutcome outcome, string message)
+		=> OutcomeDetermined?.Invoke(this, new UploadOutcomeEventArgs(outcome, message));
+
 	public override bool Validate(LibraryBook libraryBook)
 	{
 		if (!Configuration.AudiobookshelfEnabled)
@@ -26,7 +47,9 @@ public class UploadToAudiobookshelf : Processable, IProcessable<UploadToAudioboo
 			|| string.IsNullOrWhiteSpace(Configuration.AudiobookshelfFolderId))
 			return false;
 
-		return libraryBook.Book.AudioExists;
+		// Deliberately narrower than Book.AudioExists, which also accepts LiberatedStatus.Error.
+		// An errored liberation may have left partial files; uploading those is worse than skipping.
+		return libraryBook.Book.UserDefinedItem.BookStatus == LiberatedStatus.Liberated;
 	}
 
 	public override async Task<StatusHandler> ProcessAsync(LibraryBook libraryBook)
@@ -37,7 +60,10 @@ public class UploadToAudiobookshelf : Processable, IProcessable<UploadToAudioboo
 			var files = GetFilesToUpload(libraryBook);
 			if (files.Count == 0)
 			{
-				OnStatusUpdate("No audio files found to upload");
+				const string message = "No audio files found on disk to upload";
+				OnStatusUpdate(message);
+				Serilog.Log.Logger.Error("No audio files found on disk to upload for {Book}", libraryBook.LogFriendly());
+				OnOutcomeDetermined(UploadOutcome.NoFilesFound, message);
 				return new StatusHandler();
 			}
 
@@ -59,20 +85,26 @@ public class UploadToAudiobookshelf : Processable, IProcessable<UploadToAudioboo
 
 			if (result == AudiobookshelfApiService.UploadResult.Success)
 			{
-				OnStatusUpdate("Upload to Audiobookshelf completed successfully");
+				const string message = "Upload to Audiobookshelf completed successfully";
+				OnStatusUpdate(message);
+				OnOutcomeDetermined(UploadOutcome.Uploaded, message);
 				return new StatusHandler();
 			}
 			else if (result == AudiobookshelfApiService.UploadResult.AlreadyExists)
 			{
-				OnStatusUpdate("Book already exists on Audiobookshelf; skipping upload");
+				const string message = "Book already exists on Audiobookshelf; skipping upload";
+				OnStatusUpdate(message);
 				Serilog.Log.Logger.Information("Book already exists on Audiobookshelf, skipping upload: {Book}", libraryBook.LogFriendly());
+				OnOutcomeDetermined(UploadOutcome.AlreadyExists, message);
 				return new StatusHandler();
 			}
 			else
 			{
-				OnStatusUpdate("Upload to Audiobookshelf failed; book remains liberated on disk");
+				const string message = "Upload to Audiobookshelf failed; book remains liberated on disk";
+				OnStatusUpdate(message);
 				Serilog.Log.Logger.Error("Audiobookshelf upload failed for {Book}, but continuing as soft-failure", libraryBook.LogFriendly());
 				// Soft-fail: log the error but do not mark the book as failed
+				OnOutcomeDetermined(UploadOutcome.Failed, message + ". See log for details.");
 				return new StatusHandler();
 			}
 		}
@@ -81,6 +113,7 @@ public class UploadToAudiobookshelf : Processable, IProcessable<UploadToAudioboo
 			Serilog.Log.Logger.Error(ex, "Error uploading {Book} to Audiobookshelf; continuing as soft-failure", libraryBook.LogFriendly());
 			OnStatusUpdate($"Audiobookshelf upload error: {ex.Message}");
 			// Soft-fail: log the error but do not mark the book as failed
+			OnOutcomeDetermined(UploadOutcome.Failed, $"Audiobookshelf upload error: {ex.Message}");
 			return new StatusHandler();
 		}
 		finally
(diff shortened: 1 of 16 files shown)


rmcrackan:
Thanks for this - clear write-up, and the backfill gap is real. The `GetPaths` fix and keeping upload soft-fail via `OutcomeDetermined` (so the GUI liberate path stays clean) look right.

Please treat the items below as actionable follow-ups on this PR (or a tight follow-up PR if you prefer). Another AI agent reading this may be able to implement them directly.

### Requested change: nest under an `abs` command group

Please do **not** ship a top-level `upload` verb. Audiobookshelf should get its own CLI namespace so we can add more ABS commands later without cluttering the root verb list.

Desired UX:

```console
libationcli abs upload
libationcli abs upload B017V4IM1G
libationcli abs upload --id B017V4IM1G
```

Today LibationCli only has **flat** CommandLineParser verbs (see `[Verb("liberate", ...)]`, `[Verb("import-account", ...)]`, etc. in `Source/LibationCli/Options/`). There is no nested subcommand infrastructure yet, so this will need a small CLI design addition, not just renaming the attribute.

If nested verbs are awkward with the current CommandLineParser setup, an acceptable interim that matches existing kebab-case style would be:

```csharp
[Verb("abs-upload", HelpText = "Upload already-liberated books to Audiobookshelf. Default: all liberated titles.\n"
	+ "Optional: specify product id(s) via --id or positional ASIN(s).\n"
	+ "Books are never re-downloaded and local files are never deleted.")]
public class AbsUploadOptions : ProcessableOptionsBase
```

```console
libationcli abs-upload
libationcli abs-upload B017V4IM1G
libationcli abs-upload --id B017V4IM1G
```

Prefer true nesting (`abs upload`) if you can get it working cleanly with help text (`libationcli help abs`, `libationcli abs upload --help`). If you go interim `abs-upload`, note that in the PR so we can migrate to `abs ...` later. Update docs (`docs/advanced/command-line-interface.md`, `docs/features/audiobookshelf.md`) to match whichever form you ship.

Keep using `ProcessableOptionsBase` so positional ASINs and `--id` stay consistent with `liberate` / `convert`.

### Other follow-ups

1. **`skipped` count vs ASIN not in library**  
   `candidates++` only runs when the book exists, so `... upload B0NOTINLIBRARY` can print a stderr skip but show `skipped: 0`. Count "ASIN not found" as skipped, or document that `skipped` means only "found but Validate failed."

2. **HelpText accuracy**  
   Do not claim "whose audio is on disk" unless Validate actually checks disk. Today Validate is `LiberatedStatus.Liberated`; missing files become `NoFilesFound`. Align help text with that.

3. **m4b + mp3 dual upload**  
   `AudibleFileStorage.Audio.GetPaths` can return both after a convert. Prefer one audio payload (e.g. keep `.m4b` if present, else `.mp3`) before calling ABS.

4. **Nits (nice to have)**  
   - Sort multipart audio paths before upload.  
   - Log `NoFilesFound` at Warning, not Error.  
   - Optional: non-zero exit when ABS config / Books dir is missing (per-book soft failures can still exit 0).

### What already looks good (please keep)

- `GetPaths` instead of cache-only lookup  
- `Validate` narrowed to `Liberated` (not `Error`)  
- `ProcessAsync` always returns success `StatusHandler`; failures via `OutcomeDetermined`  
- Early ABS config check before the run  
- No DB migration for upload state in v1  
- Unit tests for Validate / file-list / disk-scan / soft-fail  

Happy to re-review once the verb naming/`abs` nesting and the skipped/help/dual-file items are addressed.

rmcrackan:
Oh yeah, you're also welcome to do the honors of adding yourself to this :)

https://github.com/rmcrackan/Libation/blob/791425670b763265d0961f21c2f0fcab064a1eab/Source/LibationUiBase/LibationContributor.cs#L30-L56

Thanks for helping to improve Libation!

rmcrackan:
~~Thanks for the follow-up - this looks good to merge.~~ Edit: failed test

Note `abs-upload` is a fine interim. In the future I may replace this with `abs upload`.

rmcrackan:
Thanks - CI is red on Windows, but it's a single test, not a broad suite failure. The other Windows jobs were cancelled after fail-fast; Linux/macOS are green.

`GetAudioFilesOnDisk_finds_audio_that_is_absent_from_the_file_path_cache` is asserting an unprefixed `Path.Combine` path against what `AudibleFileStorage` returns on Windows. On Windows, `LongPath` yields `\\?\C:\...`, so `AreEquivalent` sees 0 matches even when the file was found.

Please normalize before comparing (e.g. `Path.GetFullPath` on both sides, or compare via `PathWithoutPrefix`). No production change needed for this one unless you want paths stripped for callers.

caiowilson:
> Thanks - CI is red on Windows, but it's a single test, not a broad suite failure. The other Windows jobs were cancelled after fail-fast; Linux/macOS are green.
> 
> `GetAudioFilesOnDisk_finds_audio_that_is_absent_from_the_file_path_cache` is asserting an unprefixed `Path.Combine` path against what `AudibleFileStorage` returns on Windows. On Windows, `LongPath` yields `\\?\C:\...`, so `AreEquivalent` sees 0 matches even when the file was found.
> 
> Please normalize before comparing (e.g. `Path.GetFullPath` on both sides, or compare via `PathWithoutPrefix`). No production change needed for this one unless you want paths stripped for callers.

will do. I can't test it "for real" since I have no access to any Windows machine, but the path normalization should do it. 
about the kebab case I personally don't like it, but since it was your example, I took it to heart and followed.
I'll make a reminder to myself to write the follow-ups, but I can't say how soon. (if the gods of v1 launches let me - my day job is about to deliver- , today haha)

rmcrackan:
The kebab case isn't my favorite either but if you don't split it yourself, I'll prob refactor soon anyway to split it; so it's a temp thing.

caiowilson:
I pushed the final follow-ups to this PR, including the Windows-only test correction. I also opened [draft #1933](https://github.com/rmcrackan/Libation/pull/1933), which is the no-legacy-alias alternative described below.

### What this PR now does

- Adds the discoverable `abs` command group and its `abs upload` subcommand.
- Preserves the existing upload implementation and its behavior: positional ASINs, repeated `--id`, no re-downloads, and no local-file deletion.
- Supports `abs`, `help abs`, and `abs --help` for group help; `help abs upload`, `abs upload --help`, and `abs upload -h` for the upload options; and rejects inverse grammar such as `abs help upload`.
- Keeps the router deliberately one level deep (group + subcommand), rather than introducing a general hierarchy that the CLI does not yet need.
- Adds a focused, solution-registered CLI test project and fixes the Windows path-prefix assertion in the Audiobookshelf disk-scan test.

### How it is implemented

`CliCommandRouter` recognizes only the group/subcommand portion and forwards every remaining token unchanged to the existing option parser. Group help is rendered before parsing; nested help is rewritten to the existing option-help path. The uploader and its option-processing behavior are not duplicated or changed.

### Why #1918 still retains `abs-upload`

The legacy verb is currently a deliberate silent compatibility alias, not an accidental leftover. It lets existing scripts continue to work while providing the preferred nested spelling. I kept it because your earlier feedback said the interim form was acceptable and that you might later split/refactor it; global help intentionally shows both surfaces while that decision is pending.

### Unreleased-feature alternative

Because this has not shipped, I agree there is no user-facing compatibility obligation yet. Draft #1933 removes the public kebab-cased implementation now: only `abs upload` is accepted and documented; both root `abs-upload` and root `upload` return parse error. Internally it routes `abs upload` to a private parser binding/type that is excluded from root verb discovery, so it does not create a second public command.

Verification for that draft: focused CLI tests 22/22; full solution 1,014 passed / 0 failed / 21 platform skips; the three nested help forms exit 0 in the built CLI and both root aliases exit 2.

How would you like to manage this before release? My preference is to fold #1933 into this PR now and ship only `abs upload`. Alternatively, we can merge the current interim alias and let you remove/refactor it later, or you can take that removal yourself. I will follow whichever ownership/deprecation path you prefer.

rmcrackan:
Thanks - this looks solid, and CI being green on Windows is great.

Please land this as **one PR with no kebab alias**. Fold #1933 into #1918 so the only public upload command is nested:

```console
libationcli abs upload
libationcli abs upload B017V4IM1G
libationcli abs upload --id B017V4IM1G
```

Do not ship `abs-upload` or root `upload` as public verbs. This feature has not released, so there is no compatibility need for a silent alias. After the change:

- Global help should list the `abs` group, not `abs-upload`
- `abs-upload` / `upload` should fail parse (exit 2)
- Docs should mention only `abs upload`
- Remove or update tests that assert the legacy alias / dual help listing

Keep the existing upload behavior (positional ASINs, repeated `--id`, no re-download, no local deletes, soft-fail summary) and the router/help forms you already have (`abs`, `help abs`, `abs --help`, `help abs upload`, `abs upload --help`, `abs upload -h`). Continue rejecting inverse grammar like `abs help upload`.

### Small follow-ups in the same PR

1. **Case-insensitive group/subcommand matching** in `CliCommandRouter` (use `StringComparison.OrdinalIgnoreCase`). CommandLineParser verbs are effectively case-insensitive; `ABS upload` should route the same as `abs upload`.
2. **PR title**: rename to something like `Add CLI abs upload for Audiobookshelf backfill` so it matches the final command.
3. After dropping the alias, scrub any remaining `abs-upload` mentions in docs/tests/help expectations.

Once that single no-kebab PR is updated and CI is green, this should be ready to merge. (Just close #1933.)

rmcrackan:
Thank you so much for your contribution and for your patience. I know I had a lot of feedback.

caiowilson:
oh, I thank you for such a incredible project.
clear and considerate feedback is always welcome!
hope I can help more in the future.

rmcrackan:
This is now live in [Libation version 13.7.8](https://github.com/rmcrackan/Libation/releases/latest)

PR #1933, opened 2026-08-13
Draft: remove legacy abs-upload CLI alias

caiowilson:
## What

This draft presents the unreleased Audiobookshelf CLI with `abs upload` as the only public upload syntax.

## Why

The feature has not shipped, so retaining the interim kebab-cased alias would create compatibility work before there are external users who need it.

## How

- Route `abs upload` to a private parser binding and its option type.
- Exclude that option type from root verb discovery, so neither `upload` nor `abs-upload` is a public command.
- Preserve positional ASINs, repeated `--id`, and all supported nested help forms.
- Remove legacy-alias documentation and add parser/help regression coverage.

## Verification

- `PATH="/tmp/libation-dotnet-sdk:$PATH" dotnet test --project Source/_Tests/LibationCli.Tests/LibationCli.Tests.csproj --no-restore` — 22 passed
- `cd Source && PATH="/tmp/libation-dotnet-sdk:$PATH" dotnet test --no-restore` — 1,014 passed, 0 failed, 21 platform skips
- Real CLI checks: nested help forms exit 0; direct `abs-upload` and `upload` exit 2.

## Relationship to #1918

This intentionally overlaps #1918 and is a decision vehicle, not a second feature implementation. It can either replace the compatibility-alias version before merge or be closed if the maintainer prefers the interim alias.

### DIFF
--- a/Source/FileLiberator/UploadToAudiobookshelf.cs
+++ b/Source/FileLiberator/UploadToAudiobookshelf.cs
@@ -15,6 +15,27 @@ public class UploadToAudiobookshelf : Processable, IProcessable<UploadToAudioboo
 {
 	public override string Name => "Upload to Audiobookshelf";
 
+	public enum UploadOutcome { Uploaded, AlreadyExists, NoFilesFound, Failed }
+
+	public sealed class UploadOutcomeEventArgs(UploadOutcome outcome, string message) : EventArgs
+	{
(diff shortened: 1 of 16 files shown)


caiowilson:
Superseded by #1918, which now contains the no-alias implementation and case-insensitive nested command routing.

PR #1963, opened 2026-08-17
docs(faq): use ## for question about audible standard

tippfehlr:
I noticed that the headings switch from h2 to h3 at the end of the faq


rmcrackan:
You're right that "Does Libation work with books from Audible Standard?" should have 2 hashes. The others are subheadings for Spatial audio/Dolby Atmos though, so 3 is correct for them.

rmcrackan (approved):

rmcrackan:
Perfect, thanks again.

Issue #1984, opened 2026-08-23
libation crash


sofiep03:
[Log202608.log](https://github.com/user-attachments/files/31355693/Log202608.log)

PLEASE FILL OUT THE FOLLOWING. Bug reports with limited information or lacking an attached log file may get limited or delayed help.

___

## Describe the bug  
It just says "Libation encountered a fatal error"

## Screenshots  

<img width="601" height="480" alt="Image" src="https://github.com/user-attachments/assets/885630d6-f29f-4259-9fe2-e21ac50cab74" />

## Platform  
Mac

## Log Files

[Log202608.log](https://github.com/user-attachments/files/31355734/Log202608.log)

**Default Log File Locations**
|Platform|Folder|
|-|-|
|Windows|`%userprofile%\Libation`|
|macOS|`~/Library/Application Support/Libation`|
|Linux|`~/.local/share/Libation`|

**macOS:** If that folder does not exist because Libation never starts, try launching from Terminal and say whether it works: `/Applications/Libation.app/Contents/MacOS/Libation` (see [Troubleshooting for macOS](https://github.com/rmcrackan/Libation/blob/main/docs/advanced/troubleshoot.md#macos)).

Alternative, you may open the log file folder from within Libation. Open Libation's settings, and on the first tab in Settings you can click the button 'Open log folder'.

rmcrackan:
Thanks for the log.

This is not a Libation bug in the usual sense. Two books saved successfully, then macOS started returning `Input/output error` for `/Volumes/NO NAME/Audible audiobooks/` (your Books folder and in-progress folder). After that, Libation cannot list that folder and shows the fatal error on launch.

That usually means the USB drive dropped, is full, or needs repair. Please try:

1. Eject and remount the drive, then run First Aid in Disk Utility. Give the volume a real name so it is not `NO NAME`.
2. In Settings, keep **in-progress / temporary** on your Mac (for example `~/Documents/Libation`), and only point **Books** at the external drive.
3. If Libation will not start while the drive is unplugged, open `~/Library/Application Support/Libation/Settings.json` and set `Books` (and the in-progress path) back to a folder on the Mac.

Once the volume is healthy and remounted, Libation should start again. The two finished books are already in that folder.

rmcrackan:
I just made some changes which should prevent your bug in the future. That version will get released later today.

PR #1986, opened 2026-08-24
Add parallel recursive directory scan for performance improvement

ptyork:
Introducing myself with this tiny (but impactful) change before hitting you with a massive PR for a LocalLibraryManager subproject that I accidentally got myself tangled up in. While working on that, I discovered that parallelizing the process of walking directory trees resulted in a massive (10x or more) performance bump.

Although unrelated to my other effort, I encountered the issue here, as well, because I was diagnosing the cause of slow startup on CLI commands. I have my library (> 700 books/1100 files) stored on an SMB share, and it was taking 3 to 4 seconds just to get to the point that you called ProcessAsync on a CLI Option. This PR reduces that to around 1 second.

I'm sure the change is imperceptible for small libraries stored on local SSD's. But I'm confident that it shouldn't hurt anything either. Builds and tests pass (except for one pre-existing fail) on both Windows and Linux (WSL).

Separately, have you considered temporarily bumping the SQLitePCLRaw version to get rid of those scary warnings:

```
warning NU1903: Package 'SQLitePCLRaw.lib.e_sqlite3' 2.1.11 has a known 
high severity vulnerability, https://github.com/advisories/GHSA-2m69-gcr7-jv3q
```

Just need to add something like this to DataLayer.csproj.

```
  <ItemGroup>
    <!--
      Explicitly overrides the EF Core transitive dependency to drop the
      vulnerable 2.x package. Remove when Microsoft.EntityFrameworkCore.Sqlite
      ships with the updated binary, supposedly with v11.
    -->
    <PackageReference Include="SQLitePCLRaw.bundle_e_sqlite3" Version="3.0.5" />
  </ItemGroup>
```

### DIFF
--- a/Source/FileManager/BackgroundFileSystem.cs
+++ b/Source/FileManager/BackgroundFileSystem.cs
@@ -191,14 +191,32 @@ public class BackgroundFileSystem : IDisposable
 
 	private IEnumerable<LongPath> SafestEnumerateFiles(string path)
 	{
-		try
+		var shouldRecurse = SearchOption == SearchOption.AllDirectories;
+		var results = new ConcurrentBag<LongPath>();
+
+		// Internal function to (optionally) recursively scan each subdirectory.
+		// Parallel.ForEach speeds results by an order of magnitude for large
+		// libraries, especially over high-latency connections.
+
+		void walk(string dir)
 		{
-			return FileUtility.SaferEnumerateFiles(path, SearchPattern, SearchOption);
+			try
+			{
+				var files = FileUtility.SaferEnumerateFiles(dir,
+															SearchPattern,
+															SearchOption.TopDirectoryOnly);
+				foreach (var file in files)
+					results.Add(file);
+				if (shouldRecurse)
+					Parallel.ForEach(Directory.EnumerateDirectories(dir), walk);
+			}
+			catch { }
 		}
-		catch
-		{
-			return [];
-		}
+
+		// Actually execute the directory walk and return the results
+
+		walk(path);
+		return results;
 	}
 
 	private void AddUniqueFiles(IEnumerable<LongPath> newFiles)


rmcrackan:
Nice find on the parallelism, and thanks for the clear writeup. One thing needs fixing before this can go in.

**Hidden and system directories are now scanned.** `FileUtility.SaferEnumerateFiles` builds `new EnumerationOptions { ... }` without setting `AttributesToSkip`, so it keeps .NET's default of `Hidden | System`, and with `RecurseSubdirectories = true` that filter applied to directories too. `Directory.EnumerateDirectories(dir)` uses `EnumerationOptions.Compatible`, where `AttributesToSkip = 0`, so the manual recursion now descends into them.

I confirmed this against the real `BackgroundFileSystem`: a book under `.Trash-1000/files/` is absent from the cache on master and present on this branch. Same for macOS `.Trashes`, Synology `@eaDir` and `#recycle`, `.dropbox.cache`, and a book folder in Windows' `$RECYCLE.BIN` (Windows renames the deleted top-level entry but not the files inside, so the ASIN survives). Once such a file is cached, `GetFilePathsCustom` reports the book as on disk, `GetFilePath` persists that path into `FilePathCache`, `BulkSetDownloadStatus` flips the book to `Liberated`, and "open folder" and `DownloadPdf` point into the trash. It also costs you performance where you want it most: `@eaDir` has one subdirectory per file, and `System Volume Information` is large and unreadable.

Passing the same options fixes it, and I verified the result set then matches master exactly. Probably cleanest as a sibling to `SaferEnumerateFiles` so the two options objects can't drift again:

    public static IEnumerable<LongPath> SaferEnumerateDirectories(LongPath path)
        => Directory.EnumerateDirectories(path.Path, "*", new EnumerationOptions
        {
            IgnoreInaccessible = true,
            ReturnSpecialDirectories = false,
            MatchType = MatchType.Simple
            // AttributesToSkip left at its default of Hidden | System
        }).Select(p => (LongPath)p);

Worth knowing that this exclusion was never consistent: `AddPath`, on the watcher path, applies no attribute filter, so a hidden file created while Libation runs is cached today on both versions. So it's worth deciding what the intended behavior is and applying it in both places.

**Minor, take or leave:**

`ConcurrentBag` doesn't specify enumeration order, so `GetFilePathCustom`'s `FirstOrDefault()` can pick a different file than it used to. Mostly harmless, since within one book folder every match is the same book and `FilePathCache` freezes the first pick anyway, but an `.Order()` on the way out is free.

`catch { }` is silent. #1981 just added `Serilog.Log.Logger.Debug` to the two neighboring catch sites in this class, and a debug line naming the directory would keep that consistent. The per-directory catch is otherwise a real improvement, since the old single `try` only covered the call that built the lazy enumerable and not the enumeration itself.

Nested `Parallel.ForEach` isn't bounded by `MaxDegreeOfParallelism` (each nested call gets its own budget), so the walk takes every pool thread it can get, blocked on I/O, for its duration. Not dangerous in practice since the pool's thread injection rate is the real limit, but one bounded `Parallel.ForEach` per tree level was faster in every measurement I took (and it puts concurrency under a single MaxDegreeOfParallelism you can tune, instead of leaving it to whatever the thread pool hands out).

rmcrackan:
Oh yeah, feel free to do the honors of adding yourself to: Source/LibationUiBase/LibationContributor.cs :) 

Also, heads up that there's some overlap with #1988, which I've just put up for the crash in #1984. It touches the same file, so I wanted to flag it early rather than surprise you at merge time. No conflict, but here's what changes underneath you: `FileUtility.SaferEnumerateFiles` no longer throws when a directory stops being readable partway through. It keeps what it read, logs a warning naming the directory, and optionally reports the reason:

```csharp
SaferEnumerateFiles(LongPath path, string searchPattern = "*",
    SearchOption searchOption = SearchOption.TopDirectoryOnly,
    Action<Exception>? onIncomplete = null)
```

The new parameter is optional, so your call site compiles unchanged. The old behaviour was worse than it looked: the method returned a lazy sequence, so an I/O error surfaced wherever the sequence was finally walked, past whatever try/catch the caller had wrapped around it. Your version sidesteps that anyway by materialising into the bag inside the try.

**Two things your `catch { }` will now be hiding.** `Directory.EnumerateDirectories(dir)` is still a raw call, so a subdirectory that cannot be listed vanishes silently with no log line — the one gap left in the walk once the rest reports itself. And the bare catch also swallows non-I/O exceptions, which `SaferEnumerateFiles` deliberately rethrows so a genuine bug can't hide behind a short file list. Narrowing to `catch (Exception ex) when (ex is IOException or UnauthorizedAccessException)`, with a log line, would keep both visible.

**Relevant to your case specifically.** An SMB share is exactly the kind of volume that stops answering mid-walk, and #1984 was a USB drive doing the same thing. Your change makes the walk fast; mine makes it survivable. Worth knowing that with a parallel walk a dropping share will produce a warning per directory rather than one, so the log may get chatty in that situation.

Issue #1764, opened 2026-04-25
Use properties to create additional directory levels


Jo-Be-Co:
### Current status
After the replacement, a check is carried out on all properties to ensure that, for example, a title or the author’s name does not result in an invalid path. `<title>` => `Star Wars Classic Stories: The Rise of Skywalker`

However, the folder template is also used to specify subdirectories. This is done by directly entering a path separator: `<first author>\<title>`

If you want to introduce a directory level based on the presence of a specific property, you can do so using `<has>`, `<is>` or `<cmp>`: `<has first series->{<first series>}\<-has>`

~~Über die Formatierung einer Eigenschaft funktioniert dies nicht, da hier wieder das replacement angewandt wird: `<first series[\{{N}\}\\]>`~~

This does not work via the formatting of a property, because here again the replacement is applied: `<first series[\{{N}\}\\]>`

~~Auf einer Eigenschaft bezogen, die eine Liste von Werten liefern kann, ist die Alternative mittels `<has>` u.ä. recht aufwendig: `<series[\{{N}\}\\]>` vs. `<is series[#>=1]->{<series[slice(1)]>}\<-is><is series[#>=2]->{<series[slice(2)]>}\<-is><is series[#>=3]->{<series[slice(3)]>}\<-is>...`~~ 

Based on a property that can provide a list of values, the alternative using `<has>` and the like is quite complex: `<series[\{{N}\}\\]>` vs. `<is series[#>=1]->{<series[slice(1)]>}\<-is><is series[#>=2]->{<series[slice(2)]>}\<-is><is series[#>=3]->{<series[slice(3)]>}\<-is>... `

### Suggested enhancement
Only apply the strict replacement to the values that come directly from the audiobook properties. Format templates should be retained as they are. Here, the user bears a little more responsibility, just as they do with the folder template itself.

spacertader:
I'm curious about this. Some of this a little over my head with tags.
I had copilt read through site docs. Came up with ,in the example in libation its not coming out right what I'm aiming for but this This is my vision.

Library
│
├── Patricia A. McKillip - Od Magic
├── A.A. Livingston, Dan Livingston - The Worst Warlock
│
├── Aiden Phoenix - Outsider
│   ├── 01 - Kovana
│   ├── 02 - Collisa
│   └── 03 - Horde
│
├── Virgil Knightley, Annabelle Hawthorne - Master Class
│   ├── 01 - Master Class
│   ├── 02 - Master Class 2
[...]

PR #2006, opened 2026-08-26
Honor directory separators written in tag format strings

rmcrackan:
Addresses [#1764](https://github.com/rmcrackan/Libation/issues/1764).

## Problem

A directory separator typed into a folder template asks for a directory level. One that arrives inside a book property is just a character that cannot go in a file name. `FolderTemplate.GetTemplatePartsStrings` told the two apart using the only provenance that survives evaluation, `TemplatePart.TemplateTag is null`:

```csharp
.Select(tp => tp.TemplateTag is null
    //FolderTemplate literals can have directory separator characters
    ? replacements.ReplacePathChars(tp.Value.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar))
    : replacements.ReplaceFilenameChars(tp.Value)
)
```

A format string is consumed inside the tag, so `<first series[\{{N}\}\\]>` returned the resolved series name and the user's `\` fused into a single `TemplatePart.Value` carrying a non-null `TemplateTag`. The whole thing took the `ReplaceFilenameChars` branch and the separator was replaced along with the metadata. On the Windows default preset the backslash maps to an empty string, so it disappeared without a trace. Meanwhile the same character in `<has first series->{<first series>}\<-has>` is parsed as a literal part and survives, which is the asymmetry reported in the issue.

## Approach

`TemplatePart.Value` cannot become a list of runs: `SeriesDto`, `ContributorDto`, `LocaleDto`, `CultureInfoDto` and `StringDto` all reach `CommonFormatters.TemplateStringFormatter` through `IFormattable.ToString(format, provider)`, whose return type is locked to `string`. The run boundaries therefore travel in-band as a private use character.

- Separators in text taken verbatim from a format string are marked (`MarkSeparators`). `ReplaceWithGaps` already separates the `{tag}` matches from the surrounding text, so the split points existed.
- Separators in a resolved property value are left alone, and any marks the value carries are dropped (`RemoveSeparatorMarks`) so metadata cannot forge a directory level.
- The marks are resolved once the values have been through `ReplacementCharacters`, at the existing sanitize step: to a directory separator for the folder template, to the configured slash replacement for file names, and back to the character the user typed for metadata templates.

Marks pass through `ReplaceFilenameChars` untouched, since `\uE000` is absent from `Path.GetInvalidFileNameChars()` on both platforms and matches no entry in `Replacements`.

`separator()` on a list tag gets the same treatment, so a book can be filed under each of its series at once.

The folder template now also accepts `\` as a separator on Linux. The existing `Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar)` is a no-op there because both characters are `/`, which meant `<a>\<b>` produced one directory on Windows and a literal backslash in a file name on Linux. Without this, the same template would behave differently depending on whether the separator came from template text or from a format string.

## What is not covered

.NET number and date format strings are processed by .NET itself, so a separator escaped inside one (for example `<series#[00\\]>`) is still replaced. This is noted in the docs.

## Verification

Ten cases in `TemplatesTests` and five in `CommonFormattersTests`. Five of the new cases fail without the marking; `remove_slashes` and `Test_trim_to_max_path` are untouched and still pass, which is what pins the metadata behaviour. Full suite from `Source/`: 1638 tests, 0 failures. `Source/Libation.slnx` builds with 0 errors, and `LibationWinForms` and `HangoverWinForms` were compile-checked (they have no Linux runtime, so their behaviour is unverified here, though neither was modified).

Every example added to the docs was run against the template engine rather than written from reading the code.

The Edit Folder Template dialog previews this live. A directory level from plain template text, then the same level from inside a format string, then one level per series via `separator()`:

[folder_template_levels_from_format_strings.mp4](https://cursor.com/agents/bc-6ebc29a2-d19d-42f2-8eb6-35aa7a43124a/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Ffolder_template_levels_from_format_strings.mp4)

The reporter's own template, `<first series[\{{N}\}\\]><title short>`, producing a `{Sherlock Holmes}` directory holding an `A Study in Scarlet` directory:

[Edit Folder Template preview showing a {Sherlock Holmes} directory containing an A Study in Scarlet directory](https://cursor.com/agents/bc-6ebc29a2-d19d-42f2-8eb6-35aa7a43124a/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Ffolder_template_level_from_format_string.png)

`<series[format({N}) separator(\\)]>\<title short>` nesting one directory per series:

[Edit Folder Template preview showing Sherlock Holmes, Book Collection and A Study in Scarlet as nested directories](https://cursor.com/agents/bc-6ebc29a2-d19d-42f2-8eb6-35aa7a43124a/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Ffolder_template_one_level_per_series.png)

## Note on the other thread in the issue

The `01-06 - A Study in Scarlet` sub-thread is unrelated to separators and is not a bug. Audible reports a collection's order as the string `1-6`; `SeriesOrder.Parse` splits it into `[1, "-", 6]` and `SeriesOrder.ToString` formats each numeric part independently, so `[00]` yields `01-06`. `SeriesOrder_formatters` already pins this. Selecting only the first number would need a new format option and would be better tracked separately.

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#team-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-6ebc29a2-d19d-42f2-8eb6-35aa7a43124a?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-6ebc29a2-d19d-42f2-8eb6-35aa7a43124a&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/Source/FileManager/NamingTemplate/CommonFormatters.cs
+++ b/Source/FileManager/NamingTemplate/CommonFormatters.cs
@@ -70,13 +70,17 @@ public static partial class CommonFormatters
 		// if this function is called from toString implementation of the IFormattable interface, we only get a IFormatProvider
 		var culture = GetCultureInfo(provider);
 		var oldUiCulture = Thread.CurrentThread.CurrentUICulture;
-		var result = CollapseSpacesAndTrimRegex().Replace(TagFormatRegex().ReplaceWithGaps(templateString, GetValueForMatchingTag, Unescape), string.Empty);
+		var result = CollapseSpacesAndTrimRegex().Replace(TagFormatRegex().ReplaceWithGaps(templateString, GetValueForMatchingTag, UnescapeGap), string.Empty);
 		Thread.CurrentThread.CurrentUICulture = oldUiCulture;
 		return result;
+
+		// Everything between the tags belongs to the template, so its separators ask for a directory level.
+		static string UnescapeGap(string gap) => MarkSeparators(Unescape(gap));
 
 		string GetValueForMatchingTag(Match m)
 		{
 			var tag = m.Groups["tag"].Value;
+			// An unrecognized tag is passed through verbatim, escapes and all, so it is left unmarked.
 			if (!replacements.TryGetValue(tag, out var getter)) return m.Value;
 
 			var lang = m.ResolveValue("lang");
@@ -91,7 +95,9 @@ public static partial class CommonFormatters
 				IFormattable formattable => formattable.ToString(format, cultureToUse),
 				_ => _StringFormatter(value?.ToString(), format, cultureToUse),
 			};
-			return formatted.IsNullOrEmpty() ? string.Empty : m.UnescapeValue("pre") + formatted + m.UnescapeValue("post");
+			return formatted.IsNullOrEmpty()
+				? string.Empty
+				: MarkSeparators(m.UnescapeValue("pre")) + RemoveSeparatorMarks(formatted) + MarkSeparators(m.UnescapeValue("post"));
 		}
 	}
 
@@ -243,6 +249,46 @@ public static partial class CommonFormatters
 	{
 		return Unescape(valueSpan, ['\'', '"']);
 	}
+
+	#region Separator marks
+
+	/*
+	 * A directory separator means different things depending on where it came from. Written into a
+	 * template it asks for a new directory level; arriving inside a book property it is just a
+	 * character that has no business in a file name. Formatters fuse both into one string, and
+	 * IFormattable.ToString cannot hand back anything richer, so the two are told apart by marking
+	 * the separators that came from the template with a private use character. Templates resolves
+	 * the marks once the property values have been through ReplacementCharacters.
+	 */
+
+	public const char TemplateBackSlash = '\uE000';
+	public const char TemplateForwardSlash = '\uE001';
+
+	/// <summary>Mark the directory separators in text taken verbatim from a template's format string.</summary>
+	public static string MarkSeparators(string? literal)
+		=> string.IsNullOrEmpty(literal)
+			? string.Empty
+			: literal.Replace('\\', TemplateBackSlash).Replace('/', TemplateForwardSlash);
+
+	/// <summary>Drop separator marks from a property value so that metadata cannot forge a directory level.</summary>
+	public static string RemoveSeparatorMarks(string? value)
+		=> string.IsNullOrEmpty(value)
+			? string.Empty
+			: value.IndexOfAny(separatorMarks) < 0
+				? value
+				: value.Replace($"{TemplateBackSlash}", "").Replace($"{TemplateForwardSlash}", "");
+
+	/// <summary>Replace separator marks with the text that each one should become.</summary>
+	public static string ResolveSeparators(string? text, string backSlash, string forwardSlash)
+		=> string.IsNullOrEmpty(text)
+			? string.Empty
+			: text.IndexOfAny(separatorMarks) < 0
+				? text
+				: text.Replace($"{TemplateBackSlash}", backSlash).Replace($"{TemplateForwardSlash}", forwardSlash);
+
+	private static readonly char[] separatorMarks = [TemplateBackSlash, TemplateForwardSlash];
+
+	#endregion
 
 	public static string Unescape(ReadOnlySpan<char> valueSpan, ReadOnlySpan<char> quoteChars, bool unquoteBackslash = true, bool unescapeDoubleQuotesInsideQuotes = true)
 	{
--- a/Source/LibationFileManager/Templates/Templates.cs
+++ b/Source/LibationFileManager/Templates/Templates.cs
@@ -109,7 +109,8 @@ public abstract class Templates
 	{
 		ArgumentValidator.EnsureNotNull(libraryBookDto, nameof(libraryBookDto));
 		ArgumentValidator.EnsureNotNull(multiChapProps, nameof(multiChapProps));
-		return string.Concat(NamingTemplate.Evaluate(culture, libraryBookDto, multiChapProps, new CombinedDto(libraryBookDto, multiChapProps)).Select(p => p.Value));
+		var parts = NamingTemplate.Evaluate(culture, libraryBookDto, multiChapProps, new CombinedDto(libraryBookDto, multiChapProps));
+		return RestoreSeparators(string.Concat(parts.Select(p => p.Value)));
 	}
 
 	public LongPath GetFilename(LibraryBookDto libraryBookDto, string baseDir, string fileExtension, ReplacementCharacters? replacements = null, bool returnFirstExisting = false)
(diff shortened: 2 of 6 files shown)


docs/features/naming-templates.md
```diff
@@ -310,6 +310,28 @@ Not all elements of a property are always present or have content. In this case,
 <!-- VitePress compiles each .md as a Vue SFC: a literal `{{` inside normal backtick `code` is treated as Vue interpolation and breaks the build. Elsewhere this doc uses backticks for inline examples; this line is HTML `code` with `v-pre` so `{{` can be written literally and still render like the other snippets, without additional hacks such as obscure invisible characters (like &#200B : 0-width space). -->
 <code v-pre>&lt;narrator[format({{F} }{({M}') '}{L})]&gt;</code> -> <code>Stephen Fry, Arthur (Conan) Doyle</code>
 
+### Directory Separators
+
+In the **Folder Template**, a directory separator starts another directory level. It works in the template text and inside a format template alike, so either of these gives you an author directory holding a title directory:
+
+`<first author>\<title short>`
```
Jo-Be-Co, line 317: Nice extension of the documentation 👍

The new part here uses a mixture of backticks and `<code>` tags as well as `<code v-pre>`.

docs/features/naming-templates.md
```diff
@@ -310,6 +310,28 @@ Not all elements of a property are always present or have content. In this case,
...
+
+`<first author[{F} {L}\\]><title short>`
+
+Both `\` and `/` are accepted, on every platform. Inside a format template remember that `\` escapes the next character, so a separator has to be written `\\` there. That also lets a level appear only when the property has a value, which the surrounding template text cannot do on its own:
+
+<code v-pre>&lt;first series[\{{N}\}\\]&gt;&lt;title short&gt;</code> -> <code>{Sherlock Holmes}/A Study in Scarlet</code>
```
Jo-Be-Co, line 323: In this example `\\` from the format resolves intro `/` in the path. Shall this be an example for dependences on the OS?

Source/FileManager/NamingTemplate/CommonFormatters.cs
```diff
@@ -91,7 +95,9 @@ string GetValueForMatchingTag(Match m)
 				IFormattable formattable => formattable.ToString(format, cultureToUse),
 				_ => _StringFormatter(value?.ToString(), format, cultureToUse),
 			};
-			return formatted.IsNullOrEmpty() ? string.Empty : m.UnescapeValue("pre") + formatted + m.UnescapeValue("post");
+			return formatted.IsNullOrEmpty()
+				? string.Empty
+				: MarkSeparators(m.UnescapeValue("pre")) + RemoveSeparatorMarks(formatted) + MarkSeparators(m.UnescapeValue("post"));
```
Jo-Be-Co, line 100: Not sure about this `RemoveSeparatorMarks`.

At first I would say `value` needs to be untouched and `format` should marked. I think we will need some examples here - maybe we have them already!

Jo-Be-Co: A quick example would be to put a separator inside a number format:
`<first series[\{{N}\}{#:\\00.## - }]>`
instead of
`<first series[\{{N}\}\\{#:00.## - }]>`

Source/LibationFileManager/Templates/Templates.cs
```diff
@@ -141,7 +142,24 @@ public LongPath GetFilename(LibraryBookDto libraryBookDto, MultiConvertFilePrope
...
+		return parts.Select(p => CommonFormatters.ResolveSeparators(replacements.ReplaceFilenameChars(p.Value), backSlash, forwardSlash));
+	}
+
+	/// <summary>The directory separator that the template asked for, once it is time to build a path.</summary>
+	private static string ResolveToDirectorySeparator(string value)
+		=> CommonFormatters.ResolveSeparators(value, DirectorySeparator, DirectorySeparator);
+
+	/// <summary>Restore the characters the user typed, for templates whose output is metadata rather than a path.</summary>
+	private static string RestoreSeparators(string value)
+		=> CommonFormatters.ResolveSeparators(value, "\\", "/");
```
Jo-Be-Co, line 160: In one case you use local constants for the separators and here you don't ...

Source/FileManager/NamingTemplate/CommonFormatters.cs
```diff
@@ -244,6 +250,46 @@ public static string Unescape(string valueSpan)
...
+	/*
+	 * A directory separator means different things depending on where it came from. Written into a
+	 * template it asks for a new directory level; arriving inside a book property it is just a
+	 * character that has no business in a file name. Formatters fuse both into one string, and
+	 * IFormattable.ToString cannot hand back anything richer, so the two are told apart by marking
+	 * the separators that came from the template with a private use character. Templates resolves
+	 * the marks once the property values have been through ReplacementCharacters.
+	 */
+
+	public const char TemplateBackSlash = '\uE000';
```
Jo-Be-Co, line 264: How about masking the characters to their position inside `\uE000` to `\uF8FF`:
`\u002F` → `\uE02F`
`\u005C` → `\uE05C`

To hide a character you might add `\uE000` or use `c | 0xE000` or even `c ^ 0xE000`.

With XOR this will be bidirectional thus:
```c#
private static char[] TransformInternal(
    ReadOnlySpan<char> src,
    Func<char, bool> check)
{
    char[] arr = new char[src.Length];

    for (int i = 0; i < src.Length; i++)
    {
        char c = src[i];
        arr[i] = check(c) ? (char)(c ^ 0xE000) : c;
    }

    return arr;
}
```
will serve as the transformer for:
```c#
public static string HideChars(string input, ReadOnlySpan<char> toHide) =>
    string.IsNullOrEmpty(input)
        ? string.Empty
        : new string(TransformInternal(
            input,
            c => toHide.Contains(c)
        ));
```
and
```c#
public static string UnhideAllChars(string input) =>
    string.IsNullOrEmpty(input)
        ? string.Empty
        : new string(TransformInternal(
            input,
            c => c >= '\uE000' && c <= '\uF8FF'
        ));
```

Source/FileManager/NamingTemplate/CommonFormatters.cs
```diff
@@ -244,6 +250,46 @@ public static string Unescape(string valueSpan)
...
+	public static string MarkSeparators(string? literal)
+		=> string.IsNullOrEmpty(literal)
+			? string.Empty
+			: literal.Replace('\\', TemplateBackSlash).Replace('/', TemplateForwardSlash);
+
+	/// <summary>Drop separator marks from a property value so that metadata cannot forge a directory level.</summary>
+	public static string RemoveSeparatorMarks(string? value)
+		=> string.IsNullOrEmpty(value)
+			? string.Empty
+			: value.IndexOfAny(separatorMarks) < 0
```
Jo-Be-Co, line 277: Three nearly identical methods implemented differently.

Some do use `IndexOfAny` as a performance optimizer and one doesn't.

Some use the chars or constants directly others use `$"{varname}"`.

Source/FileManager/NamingTemplate/CommonFormatters.cs
```diff
@@ -244,6 +250,46 @@ public static string Unescape(string valueSpan)
...
+				? value
+				: value.Replace($"{TemplateBackSlash}", "").Replace($"{TemplateForwardSlash}", "");
+
+	/// <summary>Replace separator marks with the text that each one should become.</summary>
+	public static string ResolveSeparators(string? text, string backSlash, string forwardSlash)
```
Jo-Be-Co, line 282: Aren't these parameters always the same? Just the plain `'\\'` and `'/'`?

Jo-Be-Co, line 282: I would say this could be done with the generic ´unhide()` shown above.

Source/FileManager/NamingTemplate/CommonFormatters.cs
```diff
@@ -244,6 +250,46 @@ public static string Unescape(string valueSpan)
...
+	public const char TemplateForwardSlash = '\uE001';
+
+	/// <summary>Mark the directory separators in text taken verbatim from a template's format string.</summary>
+	public static string MarkSeparators(string? literal)
+		=> string.IsNullOrEmpty(literal)
+			? string.Empty
+			: literal.Replace('\\', TemplateBackSlash).Replace('/', TemplateForwardSlash);
+
+	/// <summary>Drop separator marks from a property value so that metadata cannot forge a directory level.</summary>
+	public static string RemoveSeparatorMarks(string? value)
```
Jo-Be-Co, line 274: Whenever there will be a marked (or hidden) separator it will be there due to formatting or joining properties.
I don't think there are cases where they should be deleted.

Jo-Be-Co (commented):
You might add a test with a conditional tag where a list tag (like author) might get joined before it will be checked.
In that case the joined string might contain `/` or `\\` and `\uExxx` as well.

rmcrackan:
I don't think I'm actually going to pursue this; it was just a test. If you'd like to though, you're welcome to any of the code.

Issue #82, opened 2021-08-03
Still having book download issues


sybergh0st:
I have never gotten your program to work for me, in any version.  That said, I know that at least a portion, however big or small, is on my end, as others seem to be able to get it to work.  I was able to, in version 5.3.6 get my US, and UK libraries to sync up, and it recognizes all of the titles, and their covers.  Once selected, every title gives me errors, and nothing downloads.  Yesterday, it had me upgrade to yesterday's latest version.  Same scenario.  Today, it had me upgrade to 5.4.9.  It gave me errors on the first couple of titles, and has been downloading some, ever since.  I am attaching yesterday's log, to this message, and will follow with today's when the downloading completes.
[_Log202108.log](https://github.com/rmcrackan/Libation/files/6925892/_Log202108.log)
[Log202108.log](https://github.com/rmcrackan/Libation/files/6925893/Log202108.log)

Mbucari:
Audible changed their API and broke Libation 5.4.8 and earlier. Download v5.4.9 and delete everything in %temp%\Libation\DownloadsInProgress.

sybergh0st:
I have done as you suggested.  Just an update to before I did many did succeed, but not all.  After I did the fresh install of 5.4.9, and cleared out the temp folder, I still have titles that give me the same error.  Maybe Libation cannot do these types of titles, as they are slightly different.  An example of one of the stubborn type is The Third Level, by Jack Finney.

Mbucari:
I see you uploaded a log in which The Third Level failed. That log was made with Libation version 5.3.6.2. Can you please try again with the latest Libation release and submit that log instead?

Mbucari:
> as they are slightly different.
Can you please elaborate?

sybergh0st:
Elaborate: The title is a free "book" to anyone with an audible account.  I don't think it's one of the Pus books, but it functions similar. 

[_Log202108.log](https://github.com/rmcrackan/Libation/files/6926195/_Log202108.log)
[Log202108.log](https://github.com/rmcrackan/Libation/files/6926196/Log202108.log)

Mbucari:
The Third Level (asin B09B58W92Z) is a podcast episode. Episodes aren't currently supported by Libation, but this particular book is not being correctly identified as an episode so it's not being filtered out properly.

rmcrackan:
No reason we couldn’t identify episodes eventually. We’d just need to figure out the edge cases. Not much demand/incentive at the moment. Is there something you would like?

[...]

PR #2008, opened 2026-08-26
Let one account scan more than one Audible marketplace

### LINKED ISSUE #1983: Purchases from a temporarily-changed shipping address live in a separate, unscanned marketplace library — no warning, no way to discover them
**Summary**

An `Account` in Libation is bound to exactly one `Locale` (`Account.cs`), and accounts are looked up/deduped by the pair `(accountId, locale)` (`AccountsSettings.GetAccount`). This means a single Amazon/Audible login can hold **separate, independent libraries per marketplace** (e.g. `us`, `ca`, `uk`), but Libation only ever scans the marketplace(s) you've explicitly added an `Account` entry for. There's no indication in the UI that other marketplaces might hold additional purchases.

**How this happens in practice**

Some titles are only purchasable from a specific Audible storefront (region-restricted content). A common workaround is to temporarily change your Amazon shipping address to that country, buy the title, then change the address back. The purchase is then permanently recorded under that marketplace — e.g. an account normally used under `ca` ends up with a handful of titles that only exist in the `us` library.

If you only ever added your account to Libation once (under your "home" marketplace), those titles are invisible: no error, no warning, they just never appear in your scanned library. I only discovered this by independently querying the Audible API across all locales with the same auth token — the same credentials successfully authenticate against every marketplace's API host, so this is fully detectable in code without a second interactive login.

**Concretely, for my own account:**
- `ca` marketplace: 253 titles (all correctly scanned by Libation)
- `us` marketplace: 50 titles, entirely invisible to Libation, including several titles I genuinely believed I didn't own

**Suggested fixes** (any of these would help, roughly in order of effort)

1. After a successful login, silently probe the other known marketplaces (`us`, `uk`, `ca`, `au`, `de`, `fr`, `it`, `in`, `jp`, `es`, `br`) with the same auth token/locale swap, and if any return a non-empty library, surface a prompt: "We found N additional titles in your `us` marketplace under this same login — add it as a second account?"
2. At minimum, add a "Check other marketplaces" diagnostic/button on the Accounts screen that does the same probe on demand.
3. Document this limitation clearly (e.g. in the Add Account dialog or wiki) so users who've ever changed their shipping address to buy region-locked content know to check.

**Environment**
- Libation (latest, from `master` as of 2026-08)
- Confirmed via direct testing against the Audible API (not a Libation bug report from guesswork) — happy to share the verification script if useful.

rmcrackan:
Closes #1983.

A title bought while an Amazon address was briefly set to another country stays in that country's Audible library for good. A scan of the account's own marketplace never sees it - no error, no warning, the titles are simply absent. The reporter had 50 such titles in `us` under an account they use as `ca`, and only found them by querying the API themselves.

The workaround was to add the same login a second time under the other marketplace, which requires knowing it is there and logging in again. This makes it discoverable instead.

[marketplaces_demo.mp4](https://cursor.com/agents/bc-ea951c43-e78b-4856-8f04-cd08e52b6d3e/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fmarketplaces_demo.mp4)

## What it does

Each account row gets a **Marketplaces** button. It opens a dialog that asks every marketplace how many titles this login holds there, and lets the user tick the ones to scan. From then on, one scan of that account covers all of them.

Audible honors one device registration across every marketplace, so there is no second login and no second set of tokens - only the store host changes.

## What it costs when you do not use it

Nothing measurable. The probe is one request per marketplace, in sequence, and runs only when the button is clicked - never on a scan, a login, or a schedule. An account with one marketplace scans exactly as it did, and its settings file is byte-for-byte what it was: the new property is omitted entirely when empty.

For scale, an ordinary scan pages the library 50 items at a time over 10 concurrent connections. A one-off trickle of eleven single-item GETs is not in the same league.

## Stored credentials: nothing moves

This was the constraint the design was built around, so it is worth being explicit.

- Encrypted tokens are bound to their marketplace by AES-GCM associated data built from `Identity.LocaleName`, and the identity is located in the settings file by that same name. The extra marketplaces are **locale names alone**, held on `Account` and never on `Identity`, so both keep resolving to the registration that already exists. **No re-encryption, no migration, no re-login.**
- **Upgrading:** an old file has no such property, so it loads as a single-marketplace account. No file is rewritten.
- **Downgrading:** an older Libation ignores the property on load but does not emit it on save, so a downgrade-and-save **silently drops the extra marketplaces**. Those books stay in the library but cannot find an account to download from until the marketplace is re-added. Omitting the property when empty confines this to people who actually used the feature. Worth a release-note line.
- **Existing duplicate rows** (the current recommended workaround) keep working untouched. The probe recognizes a marketplace another row already holds and reports it as configured rather than offering it. Consolidating two rows is deliberately not attempted - it would mean choosing which registration to discard.

## mkb79 / audible-cli import and export

The format has one `locale_code` beside one device registration, and audible-cli keeps one file per marketplace. So:

- **Export is unchanged.** It still names the marketplace the account is registered with. The file stays valid for audible-cli, which switches marketplaces from those same tokens anyway. Extra marketplaces are Libation-side bookkeeping with no slot in the format, so an export-then-reimport loses that list and nothing else - a test says so.
- **Import** now checks *every* claim on the marketplace rather than just registrations, so a file whose marketplace an existing account already reads as an extra is caught instead of silently creating a conflicting row. The refusal names the account responsible.

## UI, docs, and the tour

- **Accounts dialog** (both frontends): a Marketplaces column, enabled on the same condition as Export - an account that has logged in at least once, since the check uses its credentials. The button shows a count when there is more than one.
- **Scan picker** (both frontends): now lists an account's marketplaces, because one checkbox there can scan several.
- **`list-accounts`**: gains an **Also scans** column, since printing one locale for an account reading three misreports what a scan does. It appears only when some account has extras; `--bare` always emits the field last, so a script reading the first five keeps working. No CLI probe verb.
- **The tour** mentions marketplaces at the end rather than at the accounts step - that step runs before any login, so the button is necessarily disabled while the user is looking at it.
- **Docs**: a new getting-started section, an FAQ entry filed under the symptom ("some of my books are missing"), plus troubleshooting and CLI updates.

Unchanged on purpose: the database schema, `Book.Locale`, `LibraryBook.Account`, download resolution, naming templates, and search. A book already records which marketplace it came from, which is what lets a download find its way back to the right store, and what answers "why do I suddenly have these books".

## Verification

- `dotnet test` from `Source/`: **1664 tests, 0 failures** (23 skipped, the usual opt-in OS-secret-store ones). New coverage for the settings round trip, old-file/new-file compatibility, the cross-row uniqueness rule, the identity and its encryption staying put, the mkb79 round trip, and the probe summary wording.
- **Exercised in the running app on Linux** (the video): the button gating, the dialog, adding and removing a marketplace, the count updating, the settings file afterwards, the scan picker text, and `list-accounts`. The probe reached all eleven marketplace hosts for real.
- One bug this caught: after every marketplace failed to answer, the summary said "No other marketplace holds titles for this account" - asserting a negative that had never been established, which is precisely the silence this feature exists to break. Unreachable marketplaces are now reported as unchecked, with a test.
- **WinForms is compile-checked only.** It builds here via `EnableWindowsTargeting` but cannot run on Linux, so the new dialog's layout still needs a look on Windows.
- The accounts screenshots in `docs/getting-started.md` predate the new column and need retaking on Windows.

## Not verified, and worth a real account before release

Library reads across marketplaces are proven. Two things are not, and neither can be from here:

1. **Licensing and downloading** a title from a non-registered marketplace. The CDS host now follows the store locale, which is the intent, but only a real region-locked title can confirm it end to end.
2. **Token refresh longevity** under this usage.

The reporter on #1983 offered their verification script and has an account with exactly this split.

## Depends on

[AudibleApi #82](https://github.com/rmcrackan/AudibleApi/pull/82), which adds the store-locale override. This branch references `AudibleApi 11.0.5.1`, so that has to publish first.

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#team-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-ea951c43-e78b-4856-8f04-cd08e52b6d3e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ea951c43-e78b-4856-8f04-cd08e52b6d3e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/docs/frequently-asked-questions.md
+++ b/docs/frequently-asked-questions.md
@@ -85,6 +85,16 @@ Like many countries, amazon gives South Africa it's own amazon site. [Unlike man
 
 (Not exactly a *frequently* asked question but it's come up more than once)
 
+## Some Of My Books Are Missing From The Scan
+
+If you have ever changed your Amazon address to another country to buy a title that was not sold in your own, that title is in **that country's** Audible library and has stayed there ever since. Audible keeps a separate library per marketplace, so a scan of your usual one will never see it - there is no error and no warning, the titles are simply not there.
+
+Open **Settings -> Accounts**, click **Marketplaces** on the account's row, and click **Check other marketplaces**. Libation asks every marketplace how many titles it holds using the credentials you already have - no second login - and you tick the ones to include. After that, one scan of that account covers all of them.
+
+The button is only available once the account has signed in, so run a scan first if it looks greyed out. See [Getting Started - Titles bought from another country](/docs/getting-started#titles-bought-from-another-country).
+
+If the titles are missing for some other reason, check whether they are marked [absent from your last scan](/docs/advanced/troubleshoot) instead.
+
 ## My Audible Account Is From Before Amazon - How Do I Add It?
 
 Choose a **pre-amazon** locale (US, UK, or Germany) in the region dropdown, and enter your old Audible **username** in the Audible email/login field - not an email address. Details: [Getting Started - Create Accounts](/docs/getting-started#create-accounts).
--- a/Source/ApplicationServices/LibraryCommands.cs
+++ b/Source/ApplicationServices/LibraryCommands.cs
@@ -328,11 +328,22 @@ public static class LibraryCommands
 		{
 			try
 			{
-				// get APIs in serial b/c of logins. do NOT move inside of parallel (Task.WhenAll)
-				var apiExtended = await ApiExtended.CreateAsync(account, allowInteractiveLogin);
+				// An account is scanned as a unit: all of its marketplaces or none of them. A partial scan would
+				// leave the marketplaces that did not run looking empty, and AbsentFromLastScan - which is keyed
+				// by account id, not by marketplace - would mark their titles absent.
+				var accountTasks = new List<Task<List<ImportItem>>>();
 
-				// add scanAccountAsync as a TASK: do not await
-				tasks.Add(scanAccountAsync(apiExtended, account, libraryOptions, archiver));
+				// ScanLocales leads with the account's own marketplace, so any login happens once, up front
+				foreach (var locale in account.ScanLocales)
+				{
+					// get APIs in serial b/c of logins. do NOT move inside of parallel (Task.WhenAll)
+					var apiExtended = await ApiExtended.CreateAsync(account, allowInteractiveLogin, storeLocale: locale);
+
+					// add scanAccountAsync as a TASK: do not await
+					accountTasks.Add(scanAccountAsync(apiExtended, account, locale, libraryOptions, archiver));
+				}
+
+				tasks.AddRange(accountTasks);
 			}
 			catch (Exception ex) when (!allowInteractiveLogin && AuthenticationExceptionHelper.IsAuthenticationFailure(ex))
 			{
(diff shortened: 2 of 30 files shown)


docs/frequently-asked-questions.md
```diff
@@ -85,6 +85,16 @@ Like many countries, amazon gives South Africa it's own amazon site. [Unlike man
 
 (Not exactly a *frequently* asked question but it's come up more than once)
 
+## Some Of My Books Are Missing From The Scan
+
+If you have ever changed your Amazon address to another country to buy a title that was not sold in your own, that title is in **that country's** Audible library and has stayed there ever since. Audible keeps a separate library per marketplace, so a scan of your usual one will never see it - there is no error and no warning, the titles are simply not there.
+
+Open **Settings -> Accounts**, click **Marketplaces** on the account's row, and click **Check other marketplaces**. Libation asks every marketplace how many titles it holds using the credentials you already have - no second login - and you tick the ones to include. After that, one scan of that account covers all of them.
```
Jo-Be-Co, line 92: You should mention that after ticking the marketplaces and scanning the libraries the save button should be pressed.

rmcrackan: Good point. Instead of docs, the next version will just warn them in-app: #2019

PR #2033, opened 2026-09-06
feat(abs): add console progress bar during Audiobookshelf upload

ydakilux:
### Summary
Adds real-time upload progress and dynamic ETA in the console during Audiobookshelf uploads (`libationcli abs upload` and `libationcli liberate`).

### Changes
- Added `ProgressStream` wrapper in `ApplicationServices` to report bytes transferred without buffering files in memory.
- Added `progress` parameter to `AudiobookshelfApiService.UploadBookAsync`.
- Wired `AverageSpeed` moving speed and ETA calculation to `OnStreamingProgressChanged` and `OnStreamingTimeRemaining` in `UploadToAudiobookshelf`.
- Updated `ConsoleProgressBar` to use carriage return `\r` and space padding so the progress bar continuously overwrites in-place on a single line.
- Added unit tests for `ProgressStream` and `ConsoleProgressBar`.

### DIFF
--- a/Source/ApplicationServices/AudiobookshelfApiService.cs
+++ b/Source/ApplicationServices/AudiobookshelfApiService.cs
@@ -483,7 +483,8 @@ public static class AudiobookshelfApiService
 		string title,
 		string? author,
 		string? series,
-		IEnumerable<string> filePaths)
+		IEnumerable<string> filePaths,
+		IProgress<(long bytesSent, long totalBytes)>? progress = null)
 	{
 		apiToken = AudiobookshelfTokenStorage.DecryptToken(apiToken) ?? "";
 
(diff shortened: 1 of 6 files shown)


rmcrackan:
Just to let you know, I'm not going to be able to review your PRs for a while. My computer situation is currently very crippled. New computer will arrive late this week or beginning of next week.

rmcrackan:
Thanks for your patience on this one — new computer is finally up and running, so I'm working through the backlog. Thanks for adding this and including tests.

One concern before merging: in the CLI, `Progress<T>` posts callbacks to the thread pool, so they can overlap or arrive out of order. Both `lastUpdate` and `AverageSpeed` have unsynchronized state, and `AverageSpeed` modifies and enumerates a `LinkedList`. Overlapping callbacks could throw an exception outside the upload's try/catch and crash the process. Queued updates could also arrive after `OnCompleted` clears the progress bar.

Could you replace `Progress<T>` with a synchronous `IProgress<T>` implementation? Since this upload path reports sequentially, that would preserve ordering, keep exceptions inside the existing try/catch, and prevent callbacks from outliving completion. Please add tests for those behaviors; the current stream and formatting tests wouldn't catch this.

Two smaller things in `AudiobookshelfApiService.cs`:

* Please log failures when reading file lengths so an incorrect progress total isn't silent.
* Optional cleanup: `ProgressStream` already disposes its inner stream, so adding both streams to the cleanup list is redundant.

Thanks again — upload progress will be useful.

PR #2034, opened 2026-09-06
fix(abs): enhance error diagnostics for upload failures

ydakilux:
### Summary
Replaces opaque .NET `Error while copying content to a stream` exception messages with root cause analysis and actionable troubleshooting suggestions.

### Changes
- Unwraps `ex.GetBaseException()` to surface the underlying socket/I/O error (connection reset, broken pipe, etc.).
- Detects dropped connections and suggests checking reverse proxy `client_max_body_size` (Nginx default is 1MB), proxy timeouts, server disk space, and server logs.
- Displays the path to the local Libation `Log.log` file.
- Added unit tests verifying diagnostic messaging in `UploadToAudiobookshelfTests`.

### DIFF
--- a/Source/FileLiberator/UploadToAudiobookshelf.cs
+++ b/Source/FileLiberator/UploadToAudiobookshelf.cs
@@ -104,22 +104,63 @@ public class UploadToAudiobookshelf : Processable, IProcessable<UploadToAudioboo
 				OnStatusUpdate(message);
 				Serilog.Log.Logger.Error("Audiobookshelf upload failed for {Book}, but continuing as soft-failure", libraryBook.LogFriendly());
 				// Soft-fail: log the error but do not mark the book as failed
-				OnOutcomeDetermined(UploadOutcome.Failed, message + ". See log for details.");
+				OnOutcomeDetermined(UploadOutcome.Failed, message + $". See Libation log ({Path.Combine(Configuration.Instance.LibationFiles.Location, "Log.log")}) and Audiobookshelf server logs for details.");
 				return new StatusHandler();
 			}
 		}
 		catch (Exception ex)
 		{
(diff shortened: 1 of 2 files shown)


rmcrackan:
Thanks for this, and for including tests. “Error while copying content to a stream” doesn’t give users much to work with, so surfacing the underlying exception is a useful improvement.

A few things before I merge:

* **Don’t infer a lost connection from the stream-copy message.** We build the multipart content from `File.OpenRead`, so a local read failure would get the same connection-loss diagnosis and proxy troubleshooting advice. Keep the wording neutral unless the underlying exception supports something more specific.

* **Prefer exception types/error codes over English message matching.** Inspect the exception chain for `SocketException.SocketErrorCode` and the `HttpRequestError` property on `HttpRequestException`/`HttpIOException` where available. Those provide classification without depending on runtime wording or language. An `IOException` alone doesn’t establish a network failure, and unknown errors should retain a neutral fallback.

* **Check timeout/cancellation first.** Currently, a cancellation with an inner “connection reset” message gets classified as a connection failure.

* **Keep the status message to one short line.** Show what failed, the underlying error, and where to find details. The proxy/timeout/disk-space checklist belongs in troubleshooting documentation. Those are possible causes, but this exception doesn’t tell us which applies.

* **Remove the formatter’s dependency on `Configuration.Instance`.** The new tests call it without setting up mock configuration. Pass any needed logging location in, or just say “see Libation logs.” Hardcoding `LibationFiles/Log.log` isn’t reliable because logging supports custom paths and rolling filenames.

Please add tests for socket and HTTP error classifications, a stream-copy exception wrapping a local read failure, and cancellation wrapping a connection error. These should verify the classification without depending on English runtime messages or real app configuration.

Happy to merge once that’s sorted.

PR #2035, opened 2026-09-06
feat(abs): add --check-asin flag for Audiobookshelf duplicate checking

ydakilux:
### Summary
Adds an opt-in `--check-asin` flag to `libationcli abs upload` to detect existing books by Audible Product ID (ASIN) in Audiobookshelf metadata and audio file paths.

### Context
When a book's title in Audiobookshelf includes localized qualifiers (e.g. `(French Edition)`), the exact title matching check can fail and trigger a duplicate upload even though the file is already on the server with matching ASIN `[<ASIN>]`.

### Changes
- Added `--check-asin` flag to `AbsUploadOptions`.
- Added ASIN query and candidate matching (metadata `asin`, folder path, and file names) to `AudiobookshelfApiService.BookExistsAsync`.
- Falls back to standard title/author matching if no ASIN match is found.

### DIFF
--- a/Source/ApplicationServices/AudiobookshelfApiService.cs
+++ b/Source/ApplicationServices/AudiobookshelfApiService.cs
@@ -297,12 +297,63 @@ public static class AudiobookshelfApiService
 		}
 	}
 
-	public static async Task<bool> BookExistsAsync(string serverUrl, string apiToken, string libraryId, string title, string? author = null)
+	public static async Task<bool> BookExistsAsync(
+		string serverUrl,
+		string apiToken,
+		string libraryId,
+		string title,
+		string? author = null,
+		string? asin = null)
 	{
 		apiToken = AudiobookshelfTokenStorage.DecryptToken(apiToken) ?? "";
 		using var client = CreateClient(serverUrl);
 		client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", apiToken);
 		client.Timeout = TimeSpan.FromSeconds(30);
+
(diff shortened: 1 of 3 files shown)


mario-fehr:
@ydakilux how does this work when you use the gui not the cli?

ydakilux:
> @ydakilux how does this work when you use the gui not the cli?

Currently implementation is only when using the CLI.

rmcrackan:
Thanks for this! Matching by ASIN makes sense, especially when the titles don't quite agree.

A couple things before merge:

1. The filename/path check only runs on books returned by the ASIN search. Looking at [Audiobookshelf's search implementation](https://github.com/advplyr/audiobookshelf/blob/master/server/utils/queries/libraryItemsBookFilters.js#L1039), that query searches title, subtitle, ASIN, and ISBN. If the ASIN is *only* in the filename or folder, the book never reaches this check. With the localized title mismatch you described, we'd still upload a duplicate. Please either implement a lookup that covers that case or narrow this to metadata ASIN matching and update the description/help text.

2. `libraryFiles` belongs directly under the library item, alongside `media`. The lookup at `item["media"]?["libraryFiles"]` won't find it. See [the ABS response structure](https://github.com/advplyr/audiobookshelf/blob/master/server/models/LibraryItem.js#L984).

Please add regression tests for the different-title/same-ASIN case and the filename-only case, plus the existing title/author fallback. It would also help to know which ABS version you've tested against.

Also: is there a reason to make this opt-in and CLI-only? If there's a downside to checking ASIN whenever we have one, I'd like to understand it. Otherwise this seems useful for GUI uploads too.

